Issue #2, opened 2018-09-24
Add a `Cardano.Prelude`


ruhatch:
In the legacy codebase we were using the `Universum` prelude. For a large project like this it makes sense to have a `Cardano.Prelude`. While importing code from the old codebase this can simply re-export `Universum`, but we can then easily replace this when we want.

dnadales:
I'd be in favor of the smallest functional prelude we can find.

ruhatch:
Closing and opening a new issue for replacing `Universum`

PR #3, opened 2018-09-24
[#2] Add Cardano.Prelude

ruhatch:


redxaxder (approved):
Looks good to me

prelude/stack.yaml
```diff
@@ -0,0 +1,4 @@
+resolver: lts-12.10
```
dnadales: Shouldn't we have the `stack.yaml` at the top level? I guess `prelude` will be used by the other packages in this repo, won't it?

nc6: Indeed, there's no point having a stack.yaml file below the top level, it would just be ignored.

dnadales: @ruhatch pointed out that we might want to factor our some packages to different repositories. However, we can do that when that time comes.

PR #5, opened 2018-09-26
Preliminary version of the ledger specs rules. [WIP]

dnadales:
I think at this stage it is important to get feedback from the more experienced people as early as possible. Hence my pull request for this first iteration of the ledger rules.

Since I guess most of the requested reviewers are familiar with @dcoutts notation and operational formalism, I haven't put much explanations about the rules yet. I'd like to add that once the rules are more or less stable. However let me know if things are unclear.

You'd notice that the ledger rules (specially for the basic UTxO model) deviate from the one @dcoutts wrote. My intention was to try and keep as many details abstract as possible while capturing the relevant properties. But maybe I did a very naive formalization that fails to capture certain relevant aspects of the ledger.

Also I wanted to see how we could model orthogonal concepts such as balances and witnesses. This is why, for instance, the rules in Figure 4 have an existential quantification on a transition from `u` to `u'` via `tx` for some transition relation. So the rule of Figure 4 is actually modeling a *family* of rules, one for each possible transition relation in the premises.

@nc6: as an additional request, could you check that I didn't do anything silly in the Nix configuration. I'm a novice user of it.


specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -0,0 +1,353 @@
...
+\label{sec:introduction}
+
+This specification models the \textit{conditions} required for the extension of
+a ledger, which is modeled here as a list of transactions. The following
+aspects are part of such conditions:
+
+\begin{description}
+\item[Balances] relationship between the total value of input and outputs
+  in a new transaction, and the unspent outputs.
+\item[Witnesses] cryptographic entities needed to validate the expenditure
```
nc6: I would suggest `spending` in place of `expenditure` here, only because `expenditure` sounds like we're validating its shopping bill!

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -0,0 +1,353 @@
...
+  of a transaction input.
+\item[Update validation] voting mechanism which captures the identification of
+  the voters, and the participants that can post update proposals.\todo{Do we model
+  this and is this the right place for it?}
```
nc6: I think this is the right place for it, but I wonder whether it shouldn't be isolated as part of this paper, since I imagine it's largely orthogonal to other concepts?

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -0,0 +1,353 @@
...
+\end{description}
+
+Aspects that will not be modeled since they are not part of the legacy-free
+implementation:
+\begin{description}
+\item[Stake] outputs do not have a transfer of stake associated to them.
+\item[Delegation] delegation certificates are posted to the blockchain, and
+  hence they are not part of the ledger.
```
nc6: OK, I guess this is requiring a slightly more specific `chain` spec in order to avoid this at the ledger level. We should maybe discuss the best place for this?

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -0,0 +1,353 @@
...
+
+\section{State transitions for UTxO}
+\label{sec:state-trans-utxo-1}
+
+The states of the UTxO transition system, along with their associated functions
+and types are defined in Figure~\ref{fig:state-trans-utxo-defs}. In particular,
+we leave the definition of $\UTxO$ and $\Coin$ abstract, and place constraints
+on the choice of these types.
+
+Regarding $\UTxO$, we require that $(\UTxO, \cup, \cap)$ is a group, where
```
nc6: Somewhat confusing, since a group has only one operation. So either we should have `$(\UTxO, \cup, \emptyset)$` being a group, or say that this is a Rng.

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -0,0 +1,353 @@
...
+$\cup$ is the addition operation, $\cap$ is the multiplication operation, and
+$\emptyset$ is the additive identity. Given $u \in \UTxO$ we will write
+$\overline{u}$ for the inverse element of $u$. Given $u, v \in \UTxO$, we write
+$u \setminus v$ as a shorthand for $\overline{\overline{u} \cup v}$.
+
+Additionally, we require that the elements in $\UTxO$ form a preorder w.r.t. a
+relation $\subseteq$.
+
+Note that the notation chosen suggests that $\UTxO$ is a set, however this does not
+necessarily has to be the case.
```
nc6: s/has/have/

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -0,0 +1,353 @@
...
+
+As for $\Coin$ the only requirement in these rules is that the elements of this
+set form a total order w.r.t a relation $\leq$.
```
nc6: As discussed, is a total order necessary? I would assume we cannot have this with chimeric ledgers.

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -0,0 +1,353 @@
...
+  \caption{Definitions associated to the UTxO transition system with witnesses}
+  \label{fig:state-trans-utxo-witnesses-defs}
+\end{figure}
+
+\begin{figure}[h]
+  \begin{equation}
+    \label{eq:utxo-witness-inductive}
+    \inference[UTxO-wit-inductive]
+    {wits = \witnesses{tx} \wedge \inout{i}{u}{o}
+      & u \trans{}{tx} u'\\
```
nc6, line 321: Should specify this is a transaction under `UTXO` inference

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -0,0 +1,353 @@
...
+
+\todo{Can we prove properties of the transition system of this section? For
+instance I'd like to formalize ``double spending'' and prove that these rules
+prevent it. Do we want it?}
+
+\section{Witnesses}
+\label{sec:witnesses}
+
+\begin{figure}[h]
+  \emph{Abstract types}
```
nc6, line 275: Include transactions under witnesses. (`UTXOW`)

nc6 (approved):
Makes sense :-) Few comments, but I think the approach is going in a sensible direction.

specs/ledger/latex/simple-ledger-spec.tex
JaredCorduan: yea, we only need a partial ordering.  Coin should also be a monoid.

specs/ledger/latex/simple-ledger-spec.tex
JaredCorduan: the ledger state will be effected by certificates, so I would be inclined to consider the certs a part of the ledger instead of the blockchain protocol.

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -0,0 +1,353 @@
...
+
+\section{Introduction}
+\label{sec:introduction}
+
+This specification models the \textit{conditions} required for the extension of
+a ledger, which is modeled here as a list of transactions. The following
+aspects are part of such conditions:
+
+\begin{description}
+\item[Balances] relationship between the total value of input and outputs
```
JaredCorduan: would this condition be clearer if we called is something like "preservation of value"?

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -0,0 +1,353 @@
...
+  \caption{Definitions associated to the UTxO transition system}
+  \label{fig:state-trans-utxo-defs}
+\end{figure*}
+
+The transition rules for unspent outputs are presented in Figure~\ref{fig:state-trans-utxo}.
+
+\begin{figure}[h]
+
+  \centering
+  \begin{equation}\label{eq:utxo-base}
```
JaredCorduan, line 229: shouldn't the base step take an arbitrary collection of UTxO (above the line), and then set the utxo to this collection?  ie the genesis UTxO?

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -0,0 +1,353 @@
...
+  %
+  \emph{Abstract Types}
+  \begin{align*}
+    & \type{UTxO} & \text{unspent transaction outputs}
+    \nextdef
+    & \type{Tx} & \text{transactions}
+  \end{align*}
+  \emph{Abstract Functions}
+  \begin{align*}
+    & \txins{} \in \Tx \mapsto \UTxO & \text{transaction inputs (to be spent)}
```
JaredCorduan, line 202: are you trying to define `UTxO` in a way such that you do not need to define `TxIn` and `TxOut`? (and so just need `txins` and `txouts`)

dnadales: Yes, that'd be my idea. Does this make sense? Do you see any potential problems?

specs/ledger/latex/simple-ledger-spec.tex
dnadales: Definitely! Thanks!

specs/ledger/latex/simple-ledger-spec.tex
dnadales: Hehe, thanks!

specs/ledger/latex/simple-ledger-spec.tex
dnadales: What do you mean by "isolated a part of this paper"? Put it in an appendix?

specs/ledger/latex/simple-ledger-spec.tex
dnadales: Alright. But could you clarify how the delegation certificates affect the ledger state? Do you mean that they are part of the ledger, or that they affect other aspects (like utxo).

specs/ledger/latex/simple-ledger-spec.tex
dnadales: Yes, I dreamed that I could generalize the `UTxO` and this was the first thing I wrote when I woke up. Not very wise :/

After discussing with Marcin and Jared, it seems it suffices to say `UTxO` should be a Monus.

dnadales: @JaredCorduan @mdimjasevic do you have any other suggestions?

specs/ledger/latex/simple-ledger-spec.tex
dnadales: After talking to Bruno, we don't even need a partial order. I'll try to change the condition to relax this restriction.

specs/ledger/latex/simple-ledger-spec.tex
dnadales: This makes sense. But couldn't you always construct a genesis `UTxO` out of individual transactions? I wonder what the implications of my decision are.

On the other hand, if we model the base rule by allowing an arbitrary `UTxO` element, shouldn't we add some other constraints? I cannot think of any at the moment, so I guess the answer might be no.

specs/ledger/latex/simple-ledger-spec.tex
dnadales: What do you mean by this?

specs/ledger/latex/simple-ledger-spec.tex
dnadales: Well that's the thing. It doesn't have to! (I think)  My idea was to build upon other transitions.

dnadales: Transition types I meant.

specs/ledger/latex/simple-ledger-spec.tex
JaredCorduan: the delegation certs will only effect delegation, but we will need to add these to the ledger state when we add delegation: registered stake keys, registered pools, delegations, pool retirements.  and probably the epoch.  all this stuff would be great to be separated from the blockchain protocol (so we don't have to care about concurrency, rollbacks, etc), so for this reason I think it's a part of the ledger.  yea?

specs/ledger/latex/simple-ledger-spec.tex
JaredCorduan: I've been thinking off and on about this all day.  I can't quite convince myself of a good algebraic model for the UTxO, or if it makes sense to think about them without the inputs and outputs. but could just be lack of creativity on my part...

specs/ledger/latex/simple-ledger-spec.tex
JaredCorduan: oh interesting!  how do you state the preservation of value property then?

specs/ledger/latex/simple-ledger-spec.tex
JaredCorduan: I guess it goes back to me still thinking about how to make sense of the UTxO without inputs and outputs.  I don't know how you justify the initial UTxO without saying it okay to do so by fiat exactly in the base step.  Or by using Bruno's monetary policy scripts.

specs/ledger/latex/simple-ledger-spec.tex
JaredCorduan: I may just need to think some more about it...

specs/ledger/latex/simple-ledger-spec.tex
dnadales: Alright, so let's put it here then :)

specs/ledger/latex/simple-ledger-spec.tex
dnadales: As far as I can see, the only thing you need from an `UTxO` is adding and subtracting things. Or do we need more? I'm changing the constraint to a Monus, and see where we get.

coot: @dnadales  I looked a bit into monuses; the two canonical examples `(N, +, -)`
(let me call the doted minus simply as `-`), and `(P(X), \cup, \)` are quite
different; and we probably need something more like `P(X)`. In the latter the
following equality holds `(a + b) - c = (a - c) + (b - c)` which is not true in
`(N, +, -)`; but it holds in `(N, max, -)` (the same `-`, but addition is
different). It might be a useful equation.

specs/ledger/latex/simple-ledger-spec.tex
dnadales: By not using a less-than-equal relation :upside_down_face: Bruno suggested to require `Value` to be a monoid, and compare inputs and outputs with equality. An implication of this is that fees have to be part of the outputs, and forged-value has to be part of the inputs. I just adapted this, please take a look and see if it makes sense.

specs/ledger/latex/simple-ledger-spec.tex
dnadales: I made a clarification about this in the doc. See if it makes sense now.

specs/ledger/latex/simple-ledger-spec.tex
dnadales: Yes, a monus would work with `P (X)`, but I'm not sure about the implications of choosing `N` as `UTxO` (which I think it isn't likely).

specs/ledger/latex/simple-ledger-spec.tex
dnadales: On a second thought, having no constraint on the genesis `UTxO` is a weaker constrain, and we can strengthen it later. So I'm changing this according to your suggestion.

JaredCorduan (approved):

specs/ledger/latex/simple-ledger-spec.tex
Zahnentferner: `Value` should be not only a monoid, but a partially ordered group. (For the example validity condition that we discussed, it was sufficient for it to be a monoid... But, more generally, it should be a partially ordered group.)

In our discussion, I just used "preservation of value" as an example of validity condition to show that, by using a clever data structure (i.e. a data structure whose instances form a monoid) for the generalization of `Value` to the multi-currency case, the statement of the validity condition does not need to change at all.  (We discussed this because you were wondering if we would be needing to check conditions separately for each currency. And I wanted to show that we don't.)

It just happened that "preservation of value" uses equality instead of `leq` (<=). But it wasn't my point to suggest that a partial order is never needed. For "preservation of value", you don't need it, but for other validity conditions you may need it.  For example, for withdrawals, the condition that the balance in your account should be greater than the value being withdrawn needs a partial order.

Now, on the topic of fees, this was discussed (mainly by @dcoutts and myself) back in January 2018, and we quickly agreed on Duncan's initial suggestion of having explicit fees and taking the fees into account explicitly in the "preservation of value" condition. Some advantages are: this allows the use of equality instead of `leq`; this prevents transactions that accidentally end up paying excessively high fees because of wrong input selection.

I still believe that it is better to have explicit fees and a proper "preservation of value" property with equality.

dnadales: Note that `Value` is constrained now to be a commutative monoid with monus, which includes a partial order relation.

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -0,0 +1,370 @@
...
+    \nextdef
+    & \txouts{} \in \Tx \mapsto \UTxO & \text{transaction outputs}
+    \nextdef
+    & \values{} \in \UTxO \mapsto \seqof{\Value} & \text{values in an UTxO}
+  \end{align*}
+  \emph{Constraints}
+  \begin{align*}
+    & (\Value, +) & \text{monoid}\\
+    & (\UTxO, \cup, \emptyset, \cap, \subseteq) & \text{commutative monoid with monus}
```
redxaxder, line 211: \setminus instead of \cap here

dnadales: Nice catch. I'll fix this in the other branch I'm working on.

Thanks!

PR #6, opened 2018-09-27
Implementation of abstract ledger spec using Backpack.

nc6:
How does this work?

We parametrise over a signature `UTxO`, which currenly encompasses the types
`Hash`, `Coin` and `Addr`. We can specify explicitly our constraints on these
types. We then define an abstract ledger in `Ledger.Abstract` which depends on
these types, but knows nothing of their concrete representation.

In `simple/UTxO` we instantiate these parameters as in Jared's original spec.

There's a bit of a wart at the moment in that I don't think I can have the
`HasHash` class being part of the signature. So this is defined in
`Ledger.Abstract` and we then need to define the `simple/Ledger` module to
contain these.

You can build this (in ghcid) using `make gchid`, or alternatively with `cabal new-build`


specs/ledger/hs/src/sig/Ledger/Abstract.hs
```diff
@@ -0,0 +1,123 @@
...
+-- |Domain restriction
+--
+--     * __TODO__ - better symbol?
+(◁) :: Set TxIn -> UTxO -> UTxO
+ins ◁ (UTxO utxo) =
+  UTxO $ Map.filterWithKey (\k _ -> k `Set.member` ins) utxo
+
+-- |Domain exclusion
+--
+(/◁) :: Set TxIn -> UTxO -> UTxO
```
JaredCorduan: how about ⋪ ?

nc6: That would be nicer, but there's no easy entry sequence for that in my emacs `TeX` input mode, and I couldn't be bothered to look up the unicode code point!

specs/ledger/hs/src/sig/Ledger/Abstract.hs
```diff
@@ -0,0 +1,123 @@
...
+
+import qualified Data.ByteString.Char8 as BS
+import           Data.Map              (Map)
+import qualified Data.Map              as Map
+import           Data.Set              (Set)
+import qualified Data.Set              as Set
+import           UTxO
+
+-- | Hash part of the ledger paylod
+class HasHash a where
```
JaredCorduan: I didn't think to use a typeclass here, this is nice!

JaredCorduan (approved):
looks great!

specs/ledger/hs/src/sig/Ledger/Abstract.hs
```diff
@@ -0,0 +1,123 @@
...
+  hash :: a -> Hash
+
+-- |A unique ID of a transaction, which is computable from the transaction.
+newtype TxId = TxId { getTxId :: Hash }
+  deriving (Show, Eq, Ord)
+
+-- |The input of a UTxO.
+--
+--     * __TODO__ - is it okay to use list indices instead of implementing the Ix Type?
+data TxIn = TxIn TxId Int deriving (Show, Eq, Ord)
```
erikd: Would it be better to use `Word` there instead of `Int`? I don't think that should ever be negative.

specs/ledger/hs/LICENSE
```diff
@@ -0,0 +1,20 @@
+Copyright (c) 2018 Nicholas Clarke
```
erikd: Should that be IOHK or something?

specs/ledger/hs/src/sig/Ledger/Abstract.hs
```diff
@@ -0,0 +1,123 @@
...
+
+-- |Verify a digital signature
+verify :: Eq a => VKey -> a -> Sig a -> Bool
+verify (VKey vk) vd (Sig sd sk) = vk == sk && vd == sd
+
+-- |Domain restriction
+--
+--     * __TODO__ - better symbol?
+(◁) :: Set TxIn -> UTxO -> UTxO
+ins ◁ (UTxO utxo) =
```
erikd: Does this need to be an operator rather than a named function?

specs/ledger/hs/src/sig/UTxO.hsig
```diff
@@ -0,0 +1,31 @@
+{- |
```
erikd: `hsig` is a backpack thing?

specs/ledger/hs/src/simple/utxo/UTxO.hs
```diff
@@ -0,0 +1,17 @@
...
+{-# LANGUAGE DerivingVia #-}
+module UTxO where
+
+import           Crypto.Hash (Digest, SHA256)
+import qualified Data.ByteArray        as BA
+import Data.Monoid (Sum(..))
+
+type Hash = Digest SHA256
+
+newtype Coin = Coin Int
```
erikd: Does it make sense for a `Coin` value to be negative? If this does not make sense, we should use `Word` there.

specs/ledger/hs/src/simple/utxo/UTxO.hs
```diff
@@ -0,0 +1,17 @@
...
+  deriving (Show, Eq, Ord)
+  deriving (Semigroup, Monoid) via (Sum Int)
```
erikd: :tada:

specs/ledger/hs/LICENSE
nc6: Yes

specs/ledger/hs/src/sig/UTxO.hsig
nc6: Yep

specs/ledger/hs/src/simple/utxo/UTxO.hs
nc6: Now updated to be `Natural`

specs/ledger/hs/src/sig/Ledger/Abstract.hs
JaredCorduan: what does that mean in Haskell?  (sorry if that's a very basic question, but I couldn't figure it out with google...)

specs/ledger/hs/src/simple/utxo/UTxO.hs
erikd: I would have thought `Word` was the more obvious choice to replace `Int`.

specs/ledger/hs/src/sig/Ledger/Abstract.hs
erikd: This function uses a unicode symbol (`◁`) for its name.

If i am talking to someone in a google hangout, how do i pronounce the `◁`?

I think giving this function a proper name instead of just using a symbol would help readability, understandability and communicatability.

dnadales:
I'm a bit late to the review but I'll add some comments nevertheless :)

specs/ledger/hs/Makefile
```diff
@@ -0,0 +1,13 @@
+help: ## Print documentation
```
dnadales, line 1: Could we have some nix files so that we don't need to rely on an installation of `ghcid` and `ghc`?

dnadales: And once we have this, we should include instructions in the README.

specs/ledger/hs/src/sig/utxo/UTxO.hsig
```diff
@@ -0,0 +1,35 @@
...
+---------------------------------------------------------------------------------
+
+-- | An encoded hash of part of the system.
+data Hash
+
+instance Eq Hash
+instance Ord Hash
+instance Show Hash
+
+-- | A unit of value held by a UTxO.
```
dnadales, line 22: After a discussion with Bruno, I wonder whether we should have `Value` instead of `Coin`. By having `Value` we can support multi currencies in the future.

specs/ledger/hs/src/sig/utxo/Ledger/Abstract.hs
```diff
@@ -0,0 +1,120 @@
...
+import qualified Data.ByteString.Char8 as BS
+import           Data.Map              (Map)
+import qualified Data.Map              as Map
+import           Data.Set              (Set)
+import qualified Data.Set              as Set
+import           UTxO
+import           Numeric.Natural       (Natural)
+
+-- | Hash part of the ledger paylod
+class HasHash a where
```
dnadales, line 16: My brain is primed by `backpack` it seems. Why isn't `hash` a part on a signature? (for instance `UTxO`)

dnadales: Argh! Damn green button. I didn't want to start a review. Sorry :/

specs/ledger/hs/src/sig/utxo/Ledger/Abstract.hs
```diff
@@ -0,0 +1,120 @@
...
+  hash :: a -> Hash
+
+-- |A unique ID of a transaction, which is computable from the transaction.
```
dnadales, line 19: Should we global replace `--|` by `-- |`?

specs/ledger/hs/src/sig/utxo/Ledger/Abstract.hs
```diff
@@ -0,0 +1,120 @@
...
+txins = inputs
+
+-- |Compute the UTxO outputs of a transaction.
+txouts :: HasHash Tx => Tx -> UTxO
+txouts tx = UTxO $
+  Map.fromList [(TxIn transId idx, out) | (out, idx) <- zip (outputs tx) [0..]]
+  where
+    transId = txid tx
+
+-- |Representation of the owner of key pair.
```
dnadales, line 54: This has to do with witnesses (authentication?), so I would expect things like `Owner`, `SKey`, etc to be part of another module.

specs/ledger/hs/src/sig/extension/Ledger/Transition.hs
```diff
@@ -0,0 +1,67 @@
...
+--
+-- - The signal is _invalid_ for this given state, and no transition is
+--   effected.
+--
+-- - The signal is valid for the state, and some function is applied to the
+--   state to generate a new state.
+--
+-- The question of validity is determined by a set of preconditions which apply
+-- over the state and signal.
+data Transition = Transition
```
dnadales, line 31: Silly question 2: If we're using `backpack` anyway, why not having a `Transition` signature?

specs/ledger/hs/src/sig/extension/Ledger/Transition.hs
```diff
@@ -0,0 +1,67 @@
...
+  { _tPredicates :: [Predicate]
+  , _tUncheckedApply :: State -> Signal -> State
```
dnadales, line 33: Is a `Signal` here what some people call `Event`?

specs/ledger/hs/src/sig/extension/Ledger/Transition.hs
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+validateAll
+  :: [Transition] -- ^ Transition rules for the system
+  -> State -- ^ State to validate
+  -> Bool -- TODO This should be something more like `Validity`
+validateAll _ st | st `elem` baseStates = True
+validateAll transitions st =
+  let
+    (oldSt, signal) = peelState st
```
dnadales, line 65: It seems that `peelState` tries to reverse the state (to a prior state). A prior, I'd doubt whether this is possible in the general case. 

Also, if I understand what `peelState` does, I'm not sure it is the right name. But `noitisnart` is an awful name ;)

specs/ledger/hs/src/sig/utxo/UTxO.hsig
JaredCorduan: yea, that's a more future proof term, I like it

specs/ledger/hs/src/simple/ledger/Ledger.hs
```diff
@@ -0,0 +1,60 @@
...
+
+instance BA.ByteArrayAccess Tx where
+  length        = BA.length . BS.pack . show
+  withByteArray = BA.withByteArray . BS.pack  . show
+
+---------------------------------------------------------------------------------
+-- UTxO transitions
+---------------------------------------------------------------------------------
+getUtxo :: [TxWits] -> UTxO
+getUtxo = foldr (\(TxWits tx _) ls -> txins tx /◁ ls `union` txouts tx ) (UTxO Map.empty)
```
dnadales, line 31: This feels weird. We use an unicode symbol for domain exclusion, but `union` for `⋃`.  Shouldn't we be consistent?

specs/ledger/hs/src/simple/ledger/Ledger.hs
```diff
@@ -0,0 +1,60 @@
...
+      NoSignal    -> st
+      TxSignal tx -> tx : st
+  }
+  where
+    predValidInputs :: Predicate
+    predValidInputs = Predicate
+      { _pName = "predValidInputs"
+      , _pValidate = \st sig -> case sig of
+          NoSignal -> Passed
+          TxSignal (TxWits tx _) -> if txins tx `Set.isSubsetOf` unspentInputs (getUtxo st)
```
dnadales, line 47: This is dependent on the choice of `txins`, `Tx`, etc. It'd be nice if we could abstract this away. Just as a reminder to ourselves.

dnadales (commented):
Actually I didn't want to start a review.

Interesting way of modeling the transitions :)

specs/ledger/hs/src/sig/Ledger/Abstract.hs
dnadales: I guess you'd pronounce it as "domain restriction". But I see your point. 

I'm also thinking a similar objection could be made about `+` for instance. So I guess this wouldn't be a problem if all the people reading the code would know what `◁` is, but this might be a wild assumption.

I like the fact that the symbol is the same as in the formal specs though ...

specs/ledger/hs/src/sig/UTxO.hsig
dnadales: I'm a bit worried about `backpack` being experimental, not having `stack` or `haddock` support. Just to share some thoughts ...

PR #7, opened 2018-09-28
Add shell.nix file to support building ledger HS spec.

nc6:


dnadales (approved):
Thanks a lot for this!

JaredCorduan (commented):
thanks @nc6 !

PR #8, opened 2018-09-28
renaming a few things

JaredCorduan:
`Coin` is now `Value`
`union` is now `∪`
`/◁` is now `⋪` (note that this new symbol doesn't seem to display well on github)

I added a `ghci` command to the makefile, and removed `ghcid-test` (which didn't work).

### DIFF
--- a/specs/ledger/hs/Makefile
+++ b/specs/ledger/hs/Makefile
@@ -1,13 +1,10 @@
 help: ## Print documentation
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
+ghci: ## Run ghci
+	cabal new-repl ledger-simple
 ghcid: ## Run ghcid
 	ghcid \
 					--command "cabal new-repl ledger-simple"
 
-ghcid-test: ## Have ghcid run the test suite
-	ghcid \
-          --command "stack ghci utxo:lib utxo:test:utxo-test --ghci-options=-fobject-code" \
-          --test "main"
-
-.PHONY: ghcid ghcid-test help
+.PHONY: ghci ghcid help
--- a/specs/ledger/hs/src/sig/utxo/Ledger/Abstract.hs
+++ b/specs/ledger/hs/src/sig/utxo/Ledger/Abstract.hs
@@ -103,16 +103,16 @@ ins ◁ (UTxO utxo) =
 
 -- |Domain exclusion
 --
-(/◁) :: Set TxIn -> UTxO -> UTxO
-ins /◁ (UTxO utxo) =
+(⋪) :: Set TxIn -> UTxO -> UTxO
+ins ⋪ (UTxO utxo) =
   UTxO $ Map.filterWithKey (\k _ -> k `Set.notMember` ins) utxo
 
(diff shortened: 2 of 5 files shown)


specs/ledger/hs/Makefile
```diff
@@ -1,13 +1,10 @@
...
+ghci: ## Run ghci
+	cabal new-repl ledger-simple
 ghcid: ## Run ghcid
 	ghcid \
 					--command "cabal new-repl ledger-simple"
 
-ghcid-test: ## Have ghcid run the test suite
-	ghcid \
-          --command "stack ghci utxo:lib utxo:test:utxo-test --ghci-options=-fobject-code" \
-          --test "main"
```
parsonsmatt, line 11: Why didn't this work? Seems like fixing it would be worthwhile. I'm familiar with `ghcid` and am happy to help :)

specs/ledger/hs/src/simple/utxo/UTxO.hs
```diff
@@ -9,7 +9,7 @@ import           Numeric.Natural       (Natural)
 
 type Hash = Digest SHA256
 
-newtype Coin = Coin Natural
+newtype Value = Value Natural
```
parsonsmatt, line 12: `Value` is a bit of an overloaded term. Why not `Lovelace` -- that gives us a precise meaning for what it is in domain terms.

parsonsmatt (commented):
Let's bikeshed the name a bit :smile:

specs/ledger/hs/Makefile
JaredCorduan: Thanks!  I actually got the script from your excellent blog post!  This used to work just great, but some of our code is in flux at the moment, and there aren't actually any tests in this repo. :)

specs/ledger/hs/src/sig/utxo/Ledger/Abstract.hs
```diff
@@ -103,18 +103,18 @@ ins ◁ (UTxO utxo) =
...
+(∪) :: UTxO -> UTxO -> UTxO
+(UTxO a) ∪ (UTxO b) = UTxO $ Map.union a b
 
 -- |Determine the total balance contained in the UTxO.
-balance :: UTxO -> Coin
-balance (UTxO utxo) = foldr addCoins mempty utxo
-  where addCoins (TxOut _ a) b = a <> b
+balance :: UTxO -> Value
+balance (UTxO utxo) = foldr addValues mempty utxo
+  where addValues (TxOut _ a) b = a <> b
```
parsonsmatt: This implementation will leak space. Prefer `foldl'` for this sort of thing.

specs/ledger/hs/src/simple/utxo/UTxO.hs
JaredCorduan: ah, great question!  Lovelace won't work, we are thinking ahead for when this backpack'ed type will be able to be used as a multi-currency.

JaredCorduan: also, the paper where multi-currency is introduced uses the term `Value`

specs/ledger/hs/Makefile
parsonsmatt: Fair :)

JaredCorduan: We definitely want tests, and I'll add it back as soon as there's something to run.

nc6 (approved):
Fixed the foldr thing, LGTM

PR #9, opened 2018-10-02
Witnessed ledger and ledger composition

nc6:
This adds a "simple" ledger (with just transactions in) and a "witnessed" ledger. An attempt is made to build one upon the other, though it's not so nice as I'd hoped, due to different models of what's in the ledger state.

### DIFF
--- a/specs/ledger/hs/src/sig/utxo/Ledger/Abstract.hs
+++ b/specs/ledger/hs/src/sig/utxo/Ledger/Abstract.hs
@@ -5,13 +5,13 @@ Core implementation of the ledger logic.
 module Ledger.Abstract where
 
 import qualified Data.ByteString.Char8 as BS
+import           Data.Foldable         (foldl')
 import           Data.Map              (Map)
 import qualified Data.Map              as Map
 import           Data.Set              (Set)
 import qualified Data.Set              as Set
-import Data.Foldable (foldl')
-import           UTxO
 import           Numeric.Natural       (Natural)
+import           UTxO
 
 -- | Hash part of the ledger paylod
 class HasHash a where
@@ -52,49 +52,6 @@ txouts tx = UTxO $
   where
     transId = txid tx
 
--- |Representation of the owner of key pair.
-newtype Owner = Owner Natural deriving (Show, Eq, Ord)
-
--- |Signing Key.
-newtype SKey = SKey Owner deriving (Show, Eq, Ord)
-
--- |Verification Key.
-newtype VKey = VKey Owner deriving (Show, Eq, Ord)
-
--- |Key Pair.
-data KeyPair = KeyPair
-  {sKey :: SKey, vKey :: VKey} deriving (Show, Eq, Ord)
-
--- |Return a key pair for a given owner.
-keyPair :: Owner -> KeyPair
-keyPair owner = KeyPair (SKey owner) (VKey owner)
-
--- |A digital signature.
-data Sig a = Sig a Owner deriving (Show, Eq, Ord)
-
--- |Proof/Witness that a transaction is authorized by the given key holder.
-data Wit = Wit VKey (Sig Tx) deriving (Show, Eq)
-
--- |A fully formed transaction.
---
---     * __TODO__ - Would it be better to name this type Tx, and rename Tx to TxBody?
-data TxWits = TxWits
-              { body     :: Tx
-              , witneses :: Set Wit
-              } deriving (Show, Eq)
-
--- |Produce a digital signature
-sign :: SKey -> a -> Sig a
-sign (SKey k) d = Sig d k
-
--- |Create a witness for transaction
-makeWitness :: KeyPair -> Tx -> Wit
-makeWitness keys tx = Wit (vKey keys) (sign (sKey keys) tx)
-
--- |Verify a digital signature
-verify :: Eq a => VKey -> a -> Sig a -> Bool
-verify (VKey vk) vd (Sig sd sk) = vk == sk && vd == sd
-
 -- |Domain restriction
 --
 --     * __TODO__ - better symbol?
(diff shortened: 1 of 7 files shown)


specs/ledger/hs/src/sig/utxo/Ledger/Abstract.hs
```diff
@@ -72,6 +101,18 @@ keyPair owner = KeyPair (SKey owner) (VKey owner)
 -- |A digital signature.
 data Sig a = Sig a Owner deriving (Show, Eq, Ord)
 
+-- |Produce a digital signature
```
dnadales, line 104: Does it make sense to have these functions in some other module (like`Ledger.Witnessed`? or something like that).

specs/ledger/hs/src/witnessed/ledger/Ledger.hs
```diff
@@ -0,0 +1,82 @@
...
+utxoInductive = Transition
+  { _tPredicates =
+     [ predInsufficientWitness
+     , predValidInputs
+     , predNoIncreasedBalance
+     ]
+  , _tUncheckedApply = \st sig -> case sig of
+      NoSignal    -> st
+      TxSignal (TxWits tx _) -> st
+        { getUtxo = txins tx ⋪ (getUtxo st) ∪ txouts tx
```
dnadales, line 24: So if I understand this correctly, this extension not only adds witnesses but also caching in the state, right?

specs/ledger/hs/src/witnessed/ledger/Ledger.hs
```diff
@@ -0,0 +1,82 @@
...
+-- |Determine if a UTxO input is authorized by a given key.
+authTxin :: VKey -> TxIn -> UTxO -> Bool
+authTxin key txin (UTxO utxo) =
+  case Map.lookup txin utxo of
+    Just (TxOut (Addr pay) _) -> hash key == pay
+    _                         -> False
+
+-- |Given a ledger state, determine if the UTxO witnesses in a given
+-- transaction are sufficient.
+-- TODO - should we only check for one witness for each unique input address?
```
dnadales, line 67: Are there situations in which we have multiple witnesses for the same input?

dnadales: I mean, if we're thinking about validation scripts, then we should abstract over the witness functions. The SOS rules I wrote already present too much details, so I should abstract the witness relations there as well, so that @JaredCorduan can extend this later on.

specs/ledger/hs/src/witnessed/ledger/Ledger.hs
```diff
@@ -0,0 +1,82 @@
...
+witnessed :: TxWits -> State -> PredicateResult
```
dnadales, line 68: It seems we don't use `witnessed`. Do you plan on using this in any of the validation predicates above?

specs/ledger/hs/src/witnessed/ledger/Ledger.hs
nc6: Yes, that's true.

specs/ledger/hs/src/witnessed/ledger/Ledger.hs
```diff
@@ -0,0 +1,82 @@
...
+    predInsufficientWitness = Predicate
+      { _pName = "predInsuffcientWitness"
+      , _pValidate = \st sig -> case sig of
+          NoSignal -> Passed
+          TxSignal (TxWits tx sig) -> if txins tx `Set.isSubsetOf` unspentInputs (getUtxo st)
+              then Passed
+              else Failed InsufficientWitnesses
+            where unspentInputs (UTxO utxo) = Map.keysSet utxo
+      }
+    predValidInputs :: Predicate
```
dnadales, line 42: It'd be nice if we could reuse the `predValidInputs` and `predNoIncreaseBalance` from `simple/ledger/Ledger`...

dnadales (approved):

specs/ledger/hs/src/witnessed/ledger/Ledger.hs
nc6: Yes, indeed. The issue is that they work on different state/signal types. My first intention was to write suitable conversion functions, but this ended up quite nasty (for example, converting from the "cached" state to the "list of transactions" state is quite annoying). Hence the current approach of just re-using auxiliary functions instead.

specs/ledger/hs/src/witnessed/ledger/Ledger.hs
JaredCorduan: I don't see any reason why we would need mutliple witnesses for the same input.  But on the first draft, I was making sure that the spec matched the implementation (but left a TODO so I'd know to come back to it).  Abstracting the witness functions is definitely a great idea.  We are going to have to do it soon anyway for modeling Plutus.

JaredCorduan (approved):
Looks good @nc6 !

specs/ledger/hs/src/witnessed/ledger/Ledger.hs
nc6: Yes, not doing so was just a mistake, fixing now.

specs/ledger/hs/src/witnessed/ledger/Ledger.hs
mdimjasevic: If I'm getting this right: @JaredCorduan, you can have a transaction where two parties have to sign it for it to be valid. Hence, that's where I see an input can have multiple witnesses.

dnadales: Indeed, but aren't scripts the mechanism by which multiple signatures are supported in other currencies (like bitcoin)? In that case, having a generic witness function would subsume this case as well.

mdimjasevic: Ok, I assumed scripts in Cardano are the same as in Bitcoin. Forget about my comment :)

Issue #1, opened 2018-09-24
Import first version of blockchain types from `cardano-sl`


ruhatch:
The new implementation needs to be able to deserialise exisiting blockchains and serialise to the same format until a hard-fork is taken. The simplest way to ensure this is to import the types from the old blockchain along with their golden/round-trip tests. We can then modify and extend those types, ensuring that all tests pass and any new serialisation is done conditionally on the software version.

dnadales:
Do we use github issues now instead of YouTrack, or are these kind of issues different in some way? Just curious :)

mdimjasevic:
I've been wondering the same as @dnadales .

ruhatch:
This is a bit coarse at the moment - I'm going to add separate issues for the sub-packages

PR #10, opened 2018-10-02
[#1] Add utilities to cardano-prelude

ruhatch:
This package contains a bunch of utility functions, types, and instances. Some of them may turn out to be unnecessary (especially QuickCheck stuff), but I tried to include as little as possible to get things working.


parsonsmatt (dismissed):
I'm opposed to a `-util` package. This stuff can all go in `cardano-prelude` until we have a better place for it.

Otherwise, this appears to mostly be code copied from the old code base. Is there anything novel here? Should we take this as an opportunity to review older decisions?

ruhatch:
@parsonsmatt Seems reasonable to me to move it to `prelude`.

There isn't really anything novel here, no. This is stuff used in the `Buildable` instances and canonical JSON instances of the existing types. The current plan is to copy over the types, along with any instances we want to keep, and the binary serialisation code and tests. Then we can simplify the types knowing that we're not breaking compatibility with the existing on the wire format.

This is definitely the opportunity to review older decisions, hence my asking about which instances and libraries we want to keep over the last few days. If there's anything here that you think we should do without I'm happy to remove it!

prelude/src/Cardano/Prelude/Orphans.hs
```diff
@@ -0,0 +1,75 @@
...
+{-# LANGUAGE PolyKinds         #-}
+{-# LANGUAGE TemplateHaskell   #-}
+{-# LANGUAGE TypeApplications  #-}
+{-# LANGUAGE TypeFamilies      #-}
+
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+-- | Orphan instances for external types/classes.
+
+module Cardano.Prelude.Orphans
```
mdimjasevic, line 9: I'm relatively new to Haskell so I'd appreciate an explanation why are these instances orphans. If you have a link to a good read on orphan instances, that would be even better.

redxaxder: Instances are orphans if they don't occur in the same file as either the definition of the typeclass or the definition of the type.

We might care because the assumption of no orphan instances speeds up compilation. 

Also, multiple instances are forbidden by very strong convention, and avoiding orphans makes it easier to ensure the convention is followed.

dnadales, line 9: Isn't there a way to avoid these orphan instances?

prelude/cardano-prelude.cabal
```diff
@@ -14,8 +14,37 @@ cabal-version:       >=1.10
...
+                     , bytestring
+                     , canonical-json
+                     , cborg
+                     , containers
+                     , formatting
+                     , hashable
+                     , mtl
+                     , tagged
+                     , text
+                     , time-units
```
redxaxder: Maybe drop `time-units` in favor of `time`.

prelude/src/Cardano/Prelude/Orphans.hs
erikd: If the data type and the type class are defined external to your package and you want a type class instance of that data type, you have two options:

* Convince one of the upstream author to include the instance in their package and wait for the new package version to be released.
* Use orphan instances.

I have not real issue with orphan instances in this case as long as they are isolated to a module named something like `Cardano.Prelude.Orphans`.

prelude/cardano-prelude.cabal
dnadales: What would be the rationale for this? (just curious)

prelude/src/Cardano/Prelude/Error.hs
```diff
@@ -0,0 +1,27 @@
...
+       ) where
+
+import           Cardano.Prelude.Base
+
+import qualified Codec.CBOR.Decoding as CBOR
+import qualified Data.Aeson.Types as A
+
+-- | Convert an 'Either'-encoded failure to an 'aeson' parser failure. The
+-- return monad is intentionally specialized because we avoid 'MonadFail'.
+toAesonError :: Either Text a -> A.Parser a
```
dnadales: Are we going to ban this `Either Text` occurrence as well, or is this case allowed? If so why?

prelude/src/Cardano/Prelude/Formatting.hs
```diff
@@ -0,0 +1,111 @@
...
+base16Builder = fromString . BS.unpack . B16.encode
+
+base16F :: Format r (ByteString -> r)
+base16F = later base16Builder
+
+--------------------------------------------------------------------------------
+-- Containers
+--------------------------------------------------------------------------------
+
+pairF :: (Buildable a, Buildable b) => Format r ((a, b) -> r)
```
dnadales, line 71: I don't know if it'd make sense to add some comments about what these functions are supposed to do. I know it'd help me :)

Also, since we're starting from scratch, does it make sense to use doctests as well? (We might add this in a separate PR).

prelude/src/Cardano/Prelude/Formatting.hs
```diff
@@ -0,0 +1,111 @@
...
+pairF = later pairBuilder
+
+listJson :: (Foldable t, Buildable a) => Format r (t a -> r)
+listJson = later listBuilderJSON
+
+listJsonIndent :: (Foldable t, Buildable a) => Word -> Format r (t a -> r)
+listJsonIndent = later . listBuilderJSONIndent
+
+mapJson
+    :: (IsList t, Item t ~ (k, v), Buildable k, Buildable v)
```
dnadales: Do we keep the 4 spaces for indentation, or we use 2? I prefer the latter, since it gives more horizontal screen space.

prelude/src/Cardano/Prelude/Formatting.hs
```diff
@@ -0,0 +1,111 @@
...
+import           Prelude
+
+import           Data.ByteString (ByteString)
+import qualified Data.ByteString.Base16 as B16
+import qualified Data.ByteString.Char8 as BS
+import qualified Data.Text.Lazy as LT
+import           Data.Text.Lazy.Builder (Builder, fromLazyText, fromString)
+import           Formatting (Format, bprint, later, (%))
+import qualified Formatting as F (build)
+import           Formatting.Buildable (Buildable (..))
```
dnadales: Could we replace the implicit imports here (`(..)`) with an explicit export?

prelude/src/Cardano/Prelude/Formatting.hs
```diff
@@ -0,0 +1,111 @@
...
+
+-- | This function helps to deduce type arising from string literal
+_listBuilder
+    :: (Foldable t, Buildable a)
+    => Builder
+    -> Builder
+    -> Builder
+    -> t a
+    -> Builder
+_listBuilder = listBuilder
```
dnadales: Uh? Why do we need `_listBuilder` then?

dnadales: Nevermind, I need to read the comments :)

dnadales: Although it'd be nice to have some words about what would happen if this function wasn't there. How does it help in deducing types.

prelude/src/Cardano/Prelude/Formatting.hs
```diff
@@ -0,0 +1,111 @@
...
+    -> t a
+    -> Builder
+listBuilder prefix delimiter suffix as = mconcat
+    [build prefix, mconcat builders, build suffix]
+  where
+    builders = foldr appendBuilder [] as
+    appendBuilder a [] = [build a]
+    appendBuilder a bs = build a : build delimiter : bs
+
+-- | This function helps to deduce type arising from string literal
```
dnadales: Helps to deduce "types" maybe?

prelude/src/Cardano/Prelude/Json/Canonical.hs
```diff
@@ -0,0 +1,92 @@
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE UndecidableInstances  #-}
```
dnadales, line 5: Could `UndecidableInstances` and orphan instances be avoided somehow? We can make an issue out if this...

prelude/src/Cardano/Prelude/Json/Canonical.hs
```diff
@@ -0,0 +1,92 @@
...
+
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+-- | Canonical encoding of 'GenesisData'.
```
dnadales: "Canonical JSON encoding"?

prelude/src/Cardano/Prelude/Json/Canonical.hs
```diff
@@ -0,0 +1,92 @@
...
+instance
+    (Applicative m, Monad m, MonadError SchemaError m)
+    => ReportSchemaErrors m
+  where
+    expected expec actual = throwError SchemaError
+        { seExpected = fromString expec
+        , seActual = fmap fromString actual
+        }
+
+instance Monad m => ToJSON m Int32 where
```
dnadales, line 49: Does it make sense to submit a pull request against the package that contains `ToJSON` (I think it is [this one](https://www.stackage.org/haddock/lts-12.11/hackage-security-0.5.3.0/Hackage-Security-JSON.html#t:ToJSON)) to get rid of these orphan instances?

prelude/cardano-prelude.cabal
```diff
@@ -14,8 +14,37 @@ cabal-version:       >=1.10
...
                      , universum
+
   default-language:    Haskell2010
   default-extensions:  NoImplicitPrelude
+
+  ghc-options:         -Wall
+                       -Werror
```
dnadales: Can we please remove the `-Werror`? Also let's add other warnings that are useful and not included in `Werror`.

prelude/src/Cardano/Prelude/Json/Parse.hs
```diff
@@ -0,0 +1,39 @@
+{-# LANGUAGE LambdaCase          #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications    #-}
+
+module Cardano.Prelude.Json.Parse
+       ( tryParseString
+       , wrapConstructor
+       ) where
+
+import           Cardano.Prelude.Base
```
dnadales, line 12: What is the convention regarding imports? Normally the local packages are put at the end of the imports list (and separated from the rest by a new line). Either way I think it'd be useful to be consistent.

prelude/src/Cardano/Prelude/Json/Parse.hs
```diff
@@ -0,0 +1,39 @@
...
+    val -> expectedButGotValue typeName val
+    where typeName = show $ typeRep (Proxy @a)
+
+wrapConstructor
+    :: forall e a m
+     . (Typeable a, ReportSchemaErrors m, ToString e)
+    => Either e a
+    -> m a
+wrapConstructor = \case
+    Left  err -> expected typeName (Just $ "error occurred: " <> toString err)
```
dnadales: Could we factor out some code? For instance:

```haskell
import           Control.Arrow ((|||))

-- | TODO: add comments
tryParseString
    :: forall a m e
     . (Typeable a, ReportSchemaErrors m, ToString e)
    => (Text -> Either e a)
    -> JSValue
    -> m a
tryParseString parser = \case
    JSString str -> eitherToM (err str) . parser . toText $ str
    val -> expectedButGotValue typeName val
    where typeName = show $ typeRep (Proxy @a)
          err str = str <> ", err was: "

wrapConstructor
    :: forall e a m
     . (Typeable a, ReportSchemaErrors m, ToString e)
    => Either e a
    -> m a
wrapConstructor = eitherToM "error occurred: "

-- | TODO add comments
eitherToM
  :: forall a m e
  . (Typeable a, ReportSchemaErrors m, ToString e)
  => String
  -> Either e a
  -> m a
eitherToM msg = report ||| pure
  where
    report err = expected typeName (Just $ msg <> toString err)
    typeName = show $ typeRep (Proxy @a)

```

prelude/src/Cardano/Prelude/Orphans.hs
dnadales: What about the issues that @redxaxder pointed out?

Should we try and submit a PR to include the orphan instances in the corresponding upstream packages?

nc6: Or wrap things in newtypes all over the place.

prelude/src/Cardano/Prelude/Json/Canonical.hs
nc6: `Undecidable` isn't usually a problem, it's `Incoherent` where things start to get nasty...

prelude/test/Test/Cardano/Prelude.hs
```diff
@@ -0,0 +1,11 @@
+module Test.Cardano.Prelude
```
dnadales, line 1: It seems the tests are not hooked up yet. Is that correct?

prelude/src/Cardano/Prelude/Json/Parse.hs
nc6: I would vote in favour of "all together, alphabetically sorted". Mostly because it's the only policy that avoids all ambiguity, where you can't decide exactly what counts as "local" and what doesn't.

prelude/test/Test/Cardano/Prelude/QuickCheck/Arbitrary.hs
```diff
@@ -0,0 +1,116 @@
...
+import           Data.List.NonEmpty (NonEmpty ((:|)))
+import           Formatting (build, sformat, (%))
+import           Test.QuickCheck (Arbitrary (..), Gen, listOf, scale, shuffle,
+                     vector)
+import           Test.QuickCheck.Gen (unGen)
+import           Test.QuickCheck.Instances.ByteString ()
+import           Test.QuickCheck.Random (mkQCGen)
+
+
+makeSmall :: Gen a -> Gen a
```
dnadales, line 35: Some comments on what this does would be helpful to me :)

prelude/test/Test/Cardano/Prelude.hs
dnadales: Nevermind. I just saw the cabal file inside `test`.

prelude/test/Test/Cardano/Prelude/Tripping.hs
```diff
@@ -0,0 +1,92 @@
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+module Test.Cardano.Prelude.Tripping where
```
dnadales: Could we include some description on what the goal of the different test modules is?

dnadales: Does it make sense to explicitly state what we export? It helps to detect dead code and trace where each function is used: e.g. if we know that a function is not exported, we know that its uses are restricted to the current module.

dnadales (approved):
I time-boxed the review to two hours :) I don't know if my suggestions make sense, but I hope they help.

prelude/src/Cardano/Prelude/Json/Parse.hs
dnadales: What I usually understand by "local" is all the packages included in our `stack` file (or repo). That said, having all mixed up and alphabetically sorted makes sense.

prelude/src/Cardano/Prelude/Orphans.hs
dnadales: That's madness! :stuck_out_tongue:

prelude/src/Cardano/Prelude/Json/Canonical.hs
dnadales: Alright, I didn't know about that. I'm always scared of the compiler hanging in an endless loop, but that never happened to me with undecidable instances :)

prelude/src/Cardano/Prelude/Error.hs
```diff
@@ -0,0 +1,27 @@
...
+toAesonError = either (fail . toString) pure
```
nc6: Please avoid the use of `fail`. `fail` should only indicate a pattern matching failure, which is clearly not the case here.

prelude/src/Cardano/Prelude/Error.hs
```diff
@@ -0,0 +1,27 @@
...
+
+aesonError :: Text -> A.Parser a
+aesonError = toAesonError . Left
+
+-- | Convert an 'Either'-encoded failure to a 'cborg' decoder failure. The
+-- return monad is intentionally specialized because we avoid 'MonadFail'.
+toCborError :: Either Text a -> CBOR.Decoder s a
+toCborError = either (fail . toString) pure
```
nc6: As above

prelude/src/Cardano/Prelude/Formatting.hs
nc6: I concur about the comments; I have no idea what these things are :-)

prelude/src/Cardano/Prelude/Formatting.hs
```diff
@@ -0,0 +1,111 @@
...
+_listBuilder
+    :: (Foldable t, Buildable a)
+    => Builder
+    -> Builder
+    -> Builder
+    -> t a
+    -> Builder
```
nc6: I feel we should just be able to use `TypeApplications` here and avoid this slightly confusing function.

prelude/src/Cardano/Prelude/Json/Canonical.hs
```diff
@@ -0,0 +1,92 @@
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE UndecidableInstances  #-}
+
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
nc6, line 7: These should go in an `Orphans` module, at least.

prelude/src/Cardano/Prelude/Json/Canonical.hs
nc6: I don't see anything in here about `GenesisData`.

prelude/src/Cardano/Prelude/Json/Parse.hs
```diff
@@ -0,0 +1,39 @@
...
+       ) where
+
+import           Cardano.Prelude.Base
+
+import           Data.Typeable (typeRep)
+import           Text.JSON.Canonical (JSValue (..),
+                     ReportSchemaErrors (expected), expectedButGotValue)
+
+
+tryParseString
```
nc6: Can we have some comments here. Why would we want to use this in favour of the Aeson parsers?

prelude/src/Cardano/Prelude/Json/Parse.hs
```diff
@@ -0,0 +1,39 @@
...
+    -> m a
+tryParseString parser = \case
+    JSString str -> case parser (toText str) of
+        Right res -> pure res
+        Left err ->
+            expected typeName (Just $ str <> ", err was: " <> toString err)
+    val -> expectedButGotValue typeName val
+    where typeName = show $ typeRep (Proxy @a)
+
+wrapConstructor
```
nc6: I cannot work out how this name corresponds to what the actual code does.

prelude/stack.yaml
```diff
@@ -1,4 +1,5 @@
-resolver: snapshot.yaml
+resolver: ../snapshot.yaml
 
 packages:
   - .
+  - test
```
nc6, line 5: There really is no point having a separate stack.yaml inside here, since stack doesn't do recursive builds, and a prelude is never going to be built by itself.

prelude/test/Test/Cardano/Prelude/QuickCheck/Arbitrary.hs
```diff
@@ -0,0 +1,116 @@
...
+makeSmall = scale f
+  where
+    -- This function is the Golden Function of Testing. It is perfect
+    -- for making tested values small. There was profound research
+    -- in this area. `f 4` is 3, yes.
```
nc6, line 40: I cannot tell if this comment is sarcastic or just bizarre.

prelude/test/Test/Cardano/Prelude/QuickCheck/Property.hs
```diff
@@ -0,0 +1,225 @@
...
+splitWord :: Word64 -> Word64 -> Gen [Word64]
+splitWord total parts | total < parts =
+    error $ "splitWord: can't split " <> show total <> " into " <> show parts <> " parts."
+                      | otherwise = map succ . take iParts <$> ((<> replicate iParts 0) <$> (sumEquals (total `div` parts + 1) $ total - parts))
+  where
+    iParts = fromIntegral parts
+
+-- | Generate list of arbitrary positive integers which sum equals given sum.
+-- All elements in the list will be smaller or equal then first parameter
+sumEquals :: Word64 -> Word64 -> Gen [Word64]
```
nc6, line 157: Would be very helpful to document the parameters here.

nc6 (dismissed):
I've requested a bunch of changes. I imagine most of these were problems copied over from the existing code, so if there's a good case for merging first and fixing them after, could be persuaded of that. But we should definitely address them!

prelude/src/Cardano/Prelude/Error.hs
ruhatch: Yes, I will remove this stuff today!

prelude/src/Cardano/Prelude/Formatting.hs
ruhatch: We certainly could switch to 2. @parsonsmatt is also in favour of two, but suggested we need to design a new style guide with the increased horizontal density taken into account. Where can we discuss that?

prelude/src/Cardano/Prelude/Formatting.hs
ruhatch: I suppose so, yeah

prelude/src/Cardano/Prelude/Formatting.hs
ruhatch: This is copied from a Serokell library and I was just trying to get things compiling. I'll give it another look and see if I can clean it up some more

prelude/src/Cardano/Prelude/Formatting.hs
dnadales: No idea, maybe `#haskell-best-practice` (Am I the only one annoyed by the missing `s` in the channel name?)

prelude/cardano-prelude.cabal
redxaxder: `time-units` provides types like `Second`, `Day`, etc. Their intended use seems to be doing things like `(addTime (200 :: Second) (1 :: Day)) :: Second)`. It also provides `getCPUTimeWithUnit :: TimeUnit a => IO a`

Even though the different measurement units are given distinct types, different kinds of time are not. A value representing a point in time (it is October 5, 2018, 8:20 AM) and a value representing an interval of time that passed (the task took 5 minutes and 3 seconds) have the same type.

It's a design where values intended to be used in place of one another have distinct types (made possible with a typeclass), but values which you wouldn't want to accidentally mix up with each other have shared types.

With `time`, you would use `UTCTime`  to represent points and `NominalDiffTime` to represent intervals. If you want time units, you can write functions `seconds, days :: Integer -> NominalDiffTime`. 

Translating the initial example, we'd have `seconds 200 + days 1`. We'd lose the option to put our units after the numbers, but the verbosity goes down.

`time` is also used more pervasively. The `formatting` package already has the relevant instances for the `time` types, which would let us lose most of the orphan instances.

prelude/src/Cardano/Prelude/Json/Canonical.hs
ruhatch: Yeah, I talked to @parsonsmatt and it seems it's just GHC being conservative in its type-class-instance termination checker

prelude/src/Cardano/Prelude/Json/Canonical.hs
ruhatch: That's a good idea and @dcoutts is the maintainer, so we should be able to upstream them reasonably fast

prelude/cardano-prelude.cabal
ruhatch: Hmm, why remove the `-Werror`? Yeah, which other warnings would you recommend?

prelude/src/Cardano/Prelude/Json/Parse.hs
ruhatch: Our convention here is:

```
Prelude modules

External library modules

Internal library modules

External testing modules

Internal testing modules
```

I think avoiding ambiguity in the imports list is not a big enough win to sacrifice readability.

In this specific case `Cardano.Prelude.Base` is the prelude for the rest of the modules in `cardano-prelude`, hence its position.

prelude/src/Cardano/Prelude/Json/Parse.hs
ruhatch: I'm gonna examine this one further, but that looks like a reasonable refactoring

prelude/test/Test/Cardano/Prelude/QuickCheck/Arbitrary.hs
ruhatch: :heavy_check_mark:

prelude/test/Test/Cardano/Prelude/Tripping.hs
ruhatch: :heavy_check_mark:

ruhatch: :heavy_check_mark:

prelude/src/Cardano/Prelude/Formatting.hs
ruhatch: Or just specialise it to `Builder` as that's how we always use it!

prelude/src/Cardano/Prelude/Json/Canonical.hs
ruhatch: The `canonical-json` package is only used for the canonical encoding of `GenesisData`, which makes up the genesis block. I will make that clearer.

prelude/test/Test/Cardano/Prelude/QuickCheck/Arbitrary.hs
ruhatch: Wtf - did not look at this closely enough! I'm considering removing all the QuickCheck crap we have anyway...

prelude/cardano-prelude.cabal
dnadales: [Here](https://input-output-rnd.slack.com/archives/C48HA39RS/p1537382550000100) is the relevant slack discussion. Let me know if you think this doesn't make sense.

In short, while developing `Werror` really gets in the way, and we can always enable it in CI (with the `--pedantic` flag in stack for instance).

prelude/test/Test/Cardano/Prelude/QuickCheck/Arbitrary.hs
dnadales: No! Please don't kill "the Golden Function of Testing" :stuck_out_tongue:

prelude/src/Cardano/Prelude/Json/Parse.hs
ruhatch: There are a bunch of cases where we want to put values into `JSString`s, e.g. `Integer`s that might overflow the `JSNum` type. In those cases we need to do an extra parsing step on the contents of the `JSString`. I'll make that clear in the comments.

prelude/src/Cardano/Prelude/Error.hs
parsonsmatt: The `Aeson` interface uses `fail` to indicate parse failures.

ruhatch: Sad, but true - we can however change this to `toAesonError :: Buildable e => Either e a -> A.Parser a`

prelude/cardano-prelude.cabal
ruhatch: Thanks for pointing this out @redxaxder - I'm going to go ahead and remove those types from the prelude and propagate things up tomorrow.

PR #11, opened 2018-10-04
[CDEC-601] LaTeX blockchain spec - block validity

mdimjasevic:
https://iohk.myjetbrains.com/youtrack/issue/CDEC-601

Just want to see if this is going in the right direction. So far there is only one transition rule defined for extending a blockchain.

Based on your feedback so far, I have to incorporate delegation more deeply. I'll do that in a separate PR and I'll merge this one.

### DIFF
--- a/specs/chain/latex/blockchain-spec.tex
+++ b/specs/chain/latex/blockchain-spec.tex
@@ -69,6 +69,7 @@
 \newcommand{\HCert}{\type{HCert}}
 
 \newcommand{\SKey}{\type{SKey}}
+\newcommand{\SKeyGen}{\type{SKey_G}}
 \newcommand{\VKey}{\type{VKey}}
 \newcommand{\Sig}{\type{Sig}}
 \newcommand{\Data}{\type{Data}}
@@ -145,6 +146,7 @@ This documents defines a semantics for operations on a blockchain.
   \emph{Derived types}
   %
   \begin{align*}
+    & sk_G \in \SKeyGen \subseteq \SKey & \text{genesis block's signing key}\\
     & \BHash & \text{block hash}\\
     & hcert \in \HCert & \text{heavyweight delegation certificate}\\
     & sig \in \Sig  & \text{signature}\\
@@ -207,7 +209,7 @@ $\text{\verifyname}$ relation is given by \eqref{eq:sign-verify}:
 \begin{align}
   \label{eq:valid-block}
     \valid{b}{vk}{hcert} \implies &
-      \exists sk. (sk, vk) \in \text{\keypairname} \wedge
+      \exists sk. (sk, vk) \in \SKeyGen \times \VKey \wedge
       \delegate{sk} = hcert \wedge \\
       & \exists h, sl, d, \sigma. (h, sl, d, \sigma) = b \wedge
         \sign{sk}{h}{sl}{d} = \sigma \wedge \\
--- a/specs/chain/latex/blockchain-spec.tex
+++ b/specs/chain/latex/blockchain-spec.tex
@@ -205,7 +205,7 @@ $\text{\verifyname}$ relation is given by \eqref{eq:sign-verify}:
 \begin{equation}
   \label{eq:sign-verify}
   \forall (sk, vk) \in \text{\keypairname}, b = (h, sl, d, \sigma) \in \Block.\\
-  \verify{vk}{m}{\sigma} \iff \sign{sk}{h}{sl}{d} = \sigma.
+  \verify{vk}{d}{\sigma} \iff \sign{sk}{h}{sl}{d} = \sigma.
 \end{equation}
 
 \begin{equation}
--- a/specs/chain/latex/blockchain-spec.tex
+++ b/specs/chain/latex/blockchain-spec.tex
@@ -61,8 +61,6 @@
 %%
 %% Types
 %%
-\newcommand{\TxId}{\type{TxId}}
-\newcommand{\Ix}{\type{Ix}}
 \newcommand{\Hash}{\type{Hash}}    % addresses as verification key hashes
 \newcommand{\BHash}{\type{BHash}}  % block hashes
 \newcommand{\Addr}{\type{Addr}}
@@ -70,16 +68,10 @@
 \newcommand{\Block}{\type{Block}}
 \newcommand{\HCert}{\type{HCert}}
 
-\newcommand{\Tx}{\type{Tx}}
-\newcommand{\UTxO}{\type{UTxO}}
-\newcommand{\Value}{\type{Value}}
-\newcommand{\TxIn}{\type{TxIn}}
-\newcommand{\TxOut}{\type{TxOut}}
 \newcommand{\SKey}{\type{SKey}}
 \newcommand{\VKey}{\type{VKey}}
 \newcommand{\Sig}{\type{Sig}}
 \newcommand{\Data}{\type{Data}}
-\newcommand{\Wit}{\type{Wit}}
 
 %%
 %% Function and relation names
@@ -87,7 +79,7 @@
 \newcommand{\hashname}{hash}
 \newcommand{\signname}{sign}
 \newcommand{\signedname}{signed}
-\newcommand{\signedseq}{signedc}
+\newcommand{\signedseqname}{signedc}
 \newcommand{\verifyname}{verify}
 \newcommand{\delegatename}{delegate}
 \newcommand{\validname}{valid}
@@ -95,13 +87,10 @@
 %% 
 %% Functions and relations
 %%
-\newcommand{\witnesses}[1]{\fun{witnesses}\ \var{#1}}
 \newcommand{\sign}[4]{\fun{\signname}\ \var{#1} ~ \var{#2} ~ \var{#3} ~ \var{#4}}
 \newcommand{\signed}[2]{\fun{\signedname}\ \var{#1} ~ \var{#2}}
-\newcommand{\signedwindow}[2]{\fun{\signedseq}\ \var{#1} ~ \var{#2}}
+\newcommand{\signedwindow}[2]{\fun{\signedseqname}\ \var{#1} ~ \var{#2}}
 \newcommand{\verify}[3]{\fun{\verifyname} ~ #1 ~ #2 ~ #3}
-\newcommand{\serialised}[1]{\llbracket \var{#1} \rrbracket}
-\newcommand{\addr}[1]{\fun{addr}\ \var{#1}}
 \newcommand{\hash}[1]{\fun{\hashname}\ \var{#1}}
 \newcommand{\delegate}[1]{\fun{\delegatename}\ \var{#1}}
 \newcommand{\valid}[3]{\fun{\validname} ~ #1 ~ #2 ~ #3}
@@ -172,7 +161,7 @@ This documents defines a semantics for operations on a blockchain.
     \text{\signname} & \in \SKey \times \BHash \times \Slot \times \Data \mapsto \Sig
       & \text{signature function}\\
     \text{\signedname} & \in \SKey \times \Block & \text{signature relation}\\
-    \text{\signedseq} & \in \SKey \times \seqof{\Block} \mapsto \powerset{\Block}
+    \text{\signedseqname} & \in \SKey \times \seqof{\Block} \mapsto \powerset{\Block}
       & \text{block count function}\\
     \text{\verifyname} & \in \VKey \times \Data \times \Sig
       & \text{verification relation}\\
@@ -192,7 +181,9 @@ This documents defines a semantics for operations on a blockchain.
 A block is valid if:
 \begin{enumerate}
 \item it is signed by any key for which a current valid heavyweight delegation
-  certificate exists that is signed by any key from the genesis block, and
+  certificate exists,
+\item the corresponding certificate is signed by any key from the genesis
+  block, and
 \item in the rolling window of the last $K$ blocks the number of blocks signed
   by the key signing this block is no more than a threshold $K \cdot t$, where
   $t$ is a constant that we will pick in the range $1/5 \leq t \leq 1/4$.
--- a/specs/chain/latex/blockchain-spec.tex
+++ b/specs/chain/latex/blockchain-spec.tex
@@ -91,6 +91,7 @@
 \newcommand{\signname}{sign}
 \newcommand{\verifyname}{verify}
 \newcommand{\delegatename}{delegate}
+\newcommand{\validname}{valid}
 %% 
 %% Functions
 %%
@@ -100,12 +101,13 @@
 \newcommand{\balance}[1]{\fun{balance}\ \var{#1}}
 \newcommand{\inputs}[1]{\fun{inputs}\ \var{#1}}
 \newcommand{\witnesses}[1]{\fun{witnesses}\ \var{#1}}
-\newcommand{\sign}[2]{\fun{\signname}\ \var{#1} \times \var{#2}}
+\newcommand{\sign}[2]{\fun{\signname}\ \var{#1} ~ \var{#2}}
 \newcommand{\verify}[3]{\fun{\verifyname} ~ #1 ~ #2 ~ #3}
 \newcommand{\serialised}[1]{\llbracket \var{#1} \rrbracket}
 \newcommand{\addr}[1]{\fun{addr}\ \var{#1}}
 \newcommand{\hash}[1]{\fun{\hashname}\ \var{#1}}
-\newcommand{\delegate}[2]{\fun{\delegatename} ~ #1 ~ #2}
+\newcommand{\delegate}[2]{\fun{\delegatename}\ \var{#1} \times \var{#2}}
+\newcommand{\valid}[3]{\fun{\validname} ~ #1 ~ #2 ~ #3}
 \newcommand{\inout}[3]{\var{#1}\mapsto_{#2}\var{#3}}
 \newcommand{\utxo}[1]{\fun{utxo}\ #1}
 
@@ -169,7 +171,8 @@ This documents defines a semantics for operations on a blockchain.
     & hcert \in \HCert & \text{heavyweight delegation certificate}\\
     & sig \in \Sig  & \text{signature}\\
     & data \in \Data  & \text{data}\\
-    & (h, sl, d, \sigma) \in \BHash \times \Slot \times \Data \times \Sig
+    & \Block = \BHash \times \Slot \times \Data \times \Sig & \\
+    & (h, sl, d, \sigma) \in \Block
       & \text{block}
   \end{align*}
   %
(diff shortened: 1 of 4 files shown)


dnadales:
I get this error when building the document using nix, do you get it as well?

```text

! Package cleveref Error: cleveref must be loaded after hyperref!.

See the cleveref package documentation for explanation.
Type  H <return>  for immediate help.
 ...
```

specs/chain/latex/blockchain-spec.tex
```diff
@@ -0,0 +1,240 @@
+\documentclass[11pt,a4paper]{article}
+\usepackage{amsmath}
+\usepackage{amssymb}
+\usepackage[capitalise,noabbrev,nameinlink]{cleveref}
```
dnadales: Apparently this package should be put after hyperref :man_shrugging:

specs/chain/latex/blockchain-spec.tex
```diff
@@ -0,0 +1,240 @@
...
+  \label{eq:valid-block}
+    \valid{b}{vk}{hcert} \implies &
+      \exists sk. (sk, vk) \in \text{\keypairname} \wedge
+      \delegate{sk} = hcert \wedge \\
+      & \exists h, sl, m, \sigma. (h, sl, m, \sigma) = b \wedge
+        \sign{sk}{h}{sl}{m} = \sigma \wedge \\
+      & | \signedwindow{sk}{\var{B_{sl-K+1}} \cdots \var{B_{sl}}} | \leq K \cdot t.
+\end{align}
+
+\begin{equation}
```
dnadales, line 229: You might want to put the inference rules inside a Figure as well, since they look nicer ;)

specs/chain/latex/blockchain-spec.tex
```diff
@@ -0,0 +1,240 @@
...
+  \end{align*}
+  \caption{Definitions associated with the blockchain transition system}
+  \label{fig:state-trans-abstract}
+\end{figure}
+
+\subsubsection{Block Validity}
+\label{sec:block-valid}
+A block is valid if:
+\begin{enumerate}
+\item it is signed by any key for which a current valid heavyweight delegation
```
dnadales: This is a very long sentence. What about adding a comma to separate it: "exists, which is signed ..."

specs/chain/latex/blockchain-spec.tex
```diff
@@ -0,0 +1,240 @@
...
+\begin{equation}
+  \label{eq:chain-extension}
+  \inference[Chain-extension]
+  {\var{B_0}\var{B_1}\cdots\var{B_l} & \valid{\var{B_{l+1}}}{vk}{hcert}
+  }
+  {{\var{B_0}\var{B_1}\cdots\var{B_l}} \trans{}{\var{B_{l+1}}} {\var{B_0}\var{B_1}\cdots\var{B_l}\var{B_{l+1}}}}
+\end{equation}
+
+\end{document}
+
```
dnadales, line 238: I appreciate the short PR :)

specs/chain/latex/blockchain-spec.tex
```diff
@@ -0,0 +1,240 @@
...
+%%% Local Variables:
+%%% mode: latex
+%%% TeX-master: t
+%%% End:
```
dnadales, line 242: Are you using Emacs again :)

specs/chain/latex/blockchain-spec.tex
dnadales: @ruhatch @erikd  do you know how this delegation relation is implemented at the moment? Can you delegate from a genesis key, to another key `k_0`, and then have `k_0` delegate to `k_1`, and so on? Or only genesis keys can re-delegate?

specs/chain/latex/blockchain-spec.tex
```diff
@@ -0,0 +1,229 @@
+\documentclass[11pt,a4paper]{article}
+\usepackage{amsmath}
+\usepackage{amssymb}
+\usepackage[capitalise,noabbrev,nameinlink]{cleveref}
```
JaredCorduan: should we move this below hyperref so that it is possible to build this document with pdflatex?

specs/chain/latex/blockchain-spec.tex
```diff
@@ -0,0 +1,240 @@
...
+  \verify{vk}{m}{\sigma} \iff \sign{sk}{m} = \sigma.
+\end{equation}
+
+\begin{equation}
+  \label{eq:signed-window}
+  \signedwindow{sk}{s} = \Set{b}{b \in s \wedge \signed{sk}{b}}
+\end{equation}
+
+\begin{align}
+  \label{eq:valid-block}
```
dnadales, line 220: I think this won't work because it is not an equation...

specs/chain/latex/blockchain-spec.tex
dnadales: Done :)

specs/chain/latex/blockchain-spec.tex
JaredCorduan: yea, it's a bit long.  would it makes sense to split this criterion into two parts?  like
1. it is signed by a key for which a current valid heavyweight delegation certificate exists,
2. the corresponding certificate is signed by a key from the genesis block,
3. rolling window stuff...

specs/chain/latex/blockchain-spec.tex
```diff
@@ -0,0 +1,240 @@
...
+        \sign{sk}{h}{sl}{m} = \sigma \wedge \\
+      & | \signedwindow{sk}{\var{B_{sl-K+1}} \cdots \var{B_{sl}}} | \leq K \cdot t.
+\end{align}
+
+\begin{equation}
+  \label{eq:chain-extension}
+  \inference[Chain-extension]
+  {\var{B_0}\var{B_1}\cdots\var{B_l} & \valid{\var{B_{l+1}}}{vk}{hcert}
+  }
+  {{\var{B_0}\var{B_1}\cdots\var{B_l}} \trans{}{\var{B_{l+1}}} {\var{B_0}\var{B_1}\cdots\var{B_l}\var{B_{l+1}}}}
```
JaredCorduan: the `1`s and the `l`s are hard to tell apart.  maybe go will `\ell` or just use a different letter like `n`?

specs/chain/latex/blockchain-spec.tex
dnadales: I guess that by `2` you're implying that a "delegatee" (sorry for slaughtering the English language) cannot delegate her rights, aren't you?

specs/chain/latex/blockchain-spec.tex
JaredCorduan: alternatively, you could use a different list notation like `B -> B; b`

specs/chain/latex/blockchain-spec.tex
nc6: We should be more formal about this also, unless this just serves to motivate the formalism.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -0,0 +1,240 @@
...
+
+\subsubsection{Block Validity}
+\label{sec:block-valid}
+A block is valid if:
+\begin{enumerate}
+\item it is signed by any key for which a current valid heavyweight delegation
+  certificate exists that is signed by any key from the genesis block, and
+\item in the rolling window of the last $K$ blocks the number of blocks signed
+  by the key signing this block is no more than a threshold $K \cdot t$, where
+  $t$ is a constant that we will pick in the range $1/5 \leq t \leq 1/4$.
```
nc6: Unlikely to be relevant, but we should be careful about what we're checking here. Ultimately we want to make sure that none of the _stakeholders_ may issue more than a certain proportion of the blocks. This would allow a single stakholder to do so by making use of judicious redelegation.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -0,0 +1,240 @@
...
+\end{enumerate}
+
+
+Relationship between the $\text{\signname}$ function and the
+$\text{\verifyname}$ relation is given by \ref{eq:sign-verify}:
+
+\begin{equation}
+  \label{eq:sign-verify}
+  \forall (sk, vk) \in \text{\keypairname}, m \in \Data, \sigma \in \Sig.\\
```
nc6: `pair` is an element of `SKey \times VKey`, so I don't think we can have something in that.

JaredCorduan (approved):
looks good Marko!

specs/chain/latex/blockchain-spec.tex
JaredCorduan: that's a good point, I was just trying to split the sentence up.  we should definitely make sure we are clear about how long the chain of delegations can be.  I do not know if we want to limit it to one delegation, as in the new delegation rules, or allow multiple.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -0,0 +1,240 @@
...
+  \nextdef
+  \verify{vk}{m}{\sigma} \iff \sign{sk}{m} = \sigma.
```
nc6: Bit of an abuse of notation here. I would prefer to explicitly define `verify`, `valid` for example as actual ternary relations (or the corresponding indicator functions thereof).

specs/chain/latex/blockchain-spec.tex
```diff
@@ -0,0 +1,240 @@
...
+      \delegate{sk} = hcert \wedge \\
+      & \exists h, sl, m, \sigma. (h, sl, m, \sigma) = b \wedge
+        \sign{sk}{h}{sl}{m} = \sigma \wedge \\
+      & | \signedwindow{sk}{\var{B_{sl-K+1}} \cdots \var{B_{sl}}} | \leq K \cdot t.
+\end{align}
+
+\begin{equation}
+  \label{eq:chain-extension}
+  \inference[Chain-extension]
+  {\var{B_0}\var{B_1}\cdots\var{B_l} & \valid{\var{B_{l+1}}}{vk}{hcert}
```
nc6: There's nothing relating `B_{l+1}`, `vk` and `hcert` here - don't we need to point out that these come from the block (or from some other environment)?

nc6 (dismissed):
Requested a few changes, mainly me being fussy about things!

specs/chain/latex/blockchain-spec.tex
mdimjasevic: If I remember it correctly, the restriction is that there can be no delegation chaining, meaning only the genesis block signing keys can delegate. I'll add that to the spec.

dnadales: Bear in mind that we should reflect the current implementation to some extent. So it'd be a good idea to ask @ruhatch , @erikd or in the #sl channel.

specs/chain/latex/blockchain-spec.tex
dnadales: Indeed, if you don't need to refer to the `B_i`'s, I'd say use the notation that @JaredCorduan  suggested.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: Yes, I am! I find it very convenient for writing LaTeX. Emacs is especially useful in moderate-to-long documents with chapters, sections and references. In particular, I'm talking about the [Reftex](https://www.gnu.org/software/emacs/manual/html_node/reftex/) minor mode.

specs/chain/latex/blockchain-spec.tex
dnadales: But ultimately, even if we sign on behalf of a genesis key, I think it is the delegator identity that counts. So if you'd delegate your key to 10 stake holders each time, it'd count as if you'd signed 10 times. Does this make sense?

specs/chain/latex/blockchain-spec.tex
mdimjasevic: Thanks! Fixing this...

specs/chain/latex/blockchain-spec.tex
mdimjasevic: @nc6, I was just thinking about this! Is it that `sk` in the spec as it is now is referring to a genesis block singing key, and not a signing/staking key of a possibly delegated to party?

specs/chain/latex/blockchain-spec.tex
dnadales: My guess is that we should include a relationship that says something like `B |- vk_g ~> vk_d`, meaning "the information in the sequence of blocks `B`  says that verification key `vk_g` delegates its block issuing rights to verification key `vk_d`". This relationship would be defined in terms of the ledger rules, which would be something like `\Lambda |- vk_g ~> vk_d`, where `\Lambda` is the sequence of transactions contained in `B`.

specs/chain/latex/blockchain-spec.tex
dnadales: AucTex and RefTex rock :)

specs/chain/latex/blockchain-spec.tex
```diff
@@ -0,0 +1,232 @@
...
+  \label{fig:state-trans-abstract}
+\end{figure}
+
+\subsubsection{Block Validity}
+\label{sec:block-valid}
+A block is valid if:
+\begin{enumerate}
+\item it is signed by any key for which a current valid heavyweight delegation
+  certificate exists,
+\item the corresponding certificate is signed by any key from the genesis
```
dnadales: Personally, I find it very helpful when the things we're talking about are named. For instance, "it is signed by a key `vk_d` ... a deleg. cert `C` exists ... `C` is signed by a key from the genesis block `sk_g`..."

specs/chain/latex/blockchain-spec.tex
```diff
@@ -0,0 +1,232 @@
...
+  block, and
+\item in the rolling window of the last $K$ blocks the number of blocks signed
```
dnadales, line 199: What's the difference between saying "rolling window of the last $K$ blocks" and just "last $K$" blocks?

specs/chain/latex/blockchain-spec.tex
```diff
@@ -0,0 +1,232 @@
...
+        \sign{sk}{h}{sl}{d} = \sigma \wedge \\
+      & | \signedwindow{sk}{\var{B_{sl-K+1}} \cdots \var{B_{sl}}} | \leq K \cdot t.
+\end{align}
+
+\begin{equation}
+  \label{eq:chain-extension}
+  \inference[Chain-extension]
+  {\var{B_0}\var{B_1}\cdots\var{B_{\ell}} & \valid{\var{B_{\ell+1}}}{vk}{hcert}
+  }
+  {{\var{B_0}\var{B_1}\cdots\var{B_{\ell}}} \trans{}{\var{B_{\ell+1}}} {\var{B_0}\var{B_1}\cdots\var{B_{\ell}}\var{B_{\ell+1}}}}
```
dnadales, line 234: Do we need to label the transition by the block?

specs/chain/latex/blockchain-spec.tex
```diff
@@ -0,0 +1,232 @@
...
+    & \Block = \BHash \times \Slot \times \Data \times \Sig & \\
+    & (h, sl, d, \sigma) \in \Block
+      & \text{block}
+  \end{align*}
+  %
+  \emph{Functions and relations}
+  %
+  \begin{align*}
+    \text{\keypairname} & \in \SKey \times \VKey & \text{key pair relation}\\
+    \text{\signname} & \in \SKey \times \BHash \times \Slot \times \Data \mapsto \Sig
```
dnadales, line 169: Why do we need to a different `sign` function than the one @dcoutts introduced? I'd expect that we could get away with the old one :)

specs/chain/latex/blockchain-spec.tex
mdimjasevic: Agreed, @dnadales!

specs/chain/latex/blockchain-spec.tex
mdimjasevic: > Bear in mind that we should reflect the current implementation to some extent. So it'd be a good idea to ask @ruhatch , @erikd or in the #sl channel.

That's a good point, @dnadales. I asked in the #sl-legacy-free channel so hopefully someone will know.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: That's a good point. Now that I introduced names, I had to make it more clear what keys we're talking about in light of delegation.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: The current formulation with a rolling window puts an emphasis that what "the last K blocks" means changes over time, so to me it's a better formulation.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: What is your suggestion? I thought that this follows a convention - the thing affecting the state transition is put on top of the transition arrow.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: That's what I started with, but if you look at the Ouroboros BFT paper, the sign function there is parameterised over four variables, which I decided to go with in the end.

dnadales: Yeah, indeed. I wonder why the're needed. I need to read that paper carefully as well...

specs/chain/latex/blockchain-spec.tex
dnadales: Not necessarily, as far as I understand. You can have "anonymous" transitions.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -126,6 +126,12 @@ \section{Basic definitions}
 
 \section{Auxiliary definitions}
 \label{sec:auxil-defin}
+TODO: Put this note on serialisation somewhere more appropriate.
```
dnadales, line 129: Nice that you captured this here! Do you mind adding ducan's remark that I pinned on slack?

https://files.slack.com/files-pri/T0N639Z4N-FD7BUR48Z/image.png

dnadales (approved):

specs/chain/latex/blockchain-spec.tex
mdimjasevic: Feel free to add it in your PR as we agreed in a private chat.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: @nc6 , isn't that given by the very relation `valid`? If `B_{l+1}`, `vk` and `hcert` are related by `valid`, then there shouldn't be a need to state that fact once more. Let me know if this is nevertheless a blocker for you to proceed with merging these changes.

dnadales:
@nc6 do you still have objections about this PR being merged? We could make issues out of pending improvements.

nc6:
@dnadales Go ahead and merge, I think my issues were probably addressed.

dnadales:
Great. If there are non-addressed concerns feel free to create an issue.

PR #12, opened 2018-10-05
Dnadales/cdec 603 heavyweight delegation rules

dnadales:
Add rules for heavyweight delegation of block-issuing rights.

### DIFF
--- a/specs/ledger/latex/simple-ledger-spec.tex
+++ b/specs/ledger/latex/simple-ledger-spec.tex
@@ -208,14 +208,14 @@ the $+$ operation.
   \emph{Constraints}
   \begin{align*}
     & (\Value, +) & \text{monoid}\\
-    & (\UTxO, \cup, \emptyset, \cap, \subseteq) & \text{commutative monoid with monus}
+    & (\UTxO, \cup, \emptyset, \setminus, \subseteq) & \text{commutative monoid with monus}
   \end{align*}
   \caption{Definitions associated to the UTxO transition system}
   \label{fig:state-trans-utxo-defs}
 \end{figure*}
 
 The transition rules for unspent outputs are presented in
-Figure~\ref{fig:state-trans-utxo}. We make use of function $\balance$, which is
+Figure~\ref{fig:state-trans-utxo}. We make use of function $\balance{}$, which is
 defined below:
 
 \begin{align*}
@@ -327,6 +327,21 @@ needed.
   \caption{UTxO with witnesses inference rules}
   \label{fig:rules-utxo-witnesses}
 \end{figure}
+
+\section{Delegation}
+\label{sec:delegation}
+
+This is related to delegation of rights to sign blocks.
+
+Requirements on the delegation certificate:
+- Certificate must be signed.
+
+Do we need to keep into account the seven bootstrap keys from the genesis
+block.
+
+Why do we need to include the certificate in the chain? Why not including it
+when signing a block? This is related to the question about the fundamental
+difference between heavyweight and lightweight delegation certificates.
 
 \section{State transitions for ledgers}
 \label{sec:state-trans-ledg}
--- a/specs/ledger/latex/simple-ledger-spec.tex
+++ b/specs/ledger/latex/simple-ledger-spec.tex
@@ -386,6 +386,13 @@ which models delegation between keys in a certificate.
 \end{figure*}
 
 \begin{figure}[h]
+
+  \begin{equation}\label{eq:delegation-base}
+    \inference[Delegation-base]
+    {}
+    {\ldelegates{\epsilon}{\var{vk}}{\var{vk}}}
+  \end{equation}
+  
   \begin{equation}\label{eq:delegation-override}
     \inference[Delegation-override]
     {\delegates{c}{\var{vk}_s}{\var{vk}_d}}
@@ -394,7 +401,7 @@ which models delegation between keys in a certificate.
 
   \begin{equation}\label{eq:delegation-inductive}
     \inference[Delegation-inductive]
-    {\delegates{c}{\var{vk_s}}{\var{vk_d}} \Rightarrow \var{vk} \neq \var{vk_d}
+    {(\delegates{c}{\var{vk_s}}{\var{vk}} \Rightarrow \var{vk} \neq \var{vk_d})
       & \ldelegates{\Gamma}{\var{vk_s}}{\var{vk_d}}
     }
     {\ldelegates{\Gamma; c}{\var{vk_s}}{\var{vk_d}}}
--- a/specs/ledger/latex/simple-ledger-spec.tex
+++ b/specs/ledger/latex/simple-ledger-spec.tex
@@ -64,8 +64,6 @@
 \newcommand{\trans}[2]{\xlongrightarrow[\textsc{#1}]{#2}}
 \newcommand{\seqof}[1]{#1^{*}}
 \newcommand{\nextdef}{\\[1em]}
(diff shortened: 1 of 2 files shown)


specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -328,6 +342,67 @@ \section{Witnesses}
...
+  \begin{align*}
+    & \delegates{c}{vk_s}{vk_d} ~\dot= ~ c = (vk_s, vk_d, \sigma) \wedge \verify{vk_s}{c}{\sigma}
+  \end{align*}
+\end{figure*}
+
+\begin{figure}[h]
+  \begin{equation}\label{eq:delegation-override}
+    \inference[Delegation-override]
+    {\delegates{c}{\var{vk}_s}{\var{vk}_d}}
+    {\ldelegates{\Gamma; c}{\var{vk_s}}{\var{vk_d}}}
```
nc6: Let's have the delegation transition defined somewhere.

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -328,6 +342,67 @@ \section{Witnesses}
...
+
+\begin{figure*}[h]
+  \emph{Delegation certificates}
+  %
+  \begin{equation*}
+    \var{c} \in \VKey \times \VKey \times \Sig
+  \end{equation*}
+  \emph{Functions}
+  \begin{align*}
+    & \delegates{c}{vk_s}{vk_d} ~\dot= ~ c = (vk_s, vk_d, \sigma) \wedge \verify{vk_s}{c}{\sigma}
```
nc6: You're using `\dot=` as "is defined to be"?

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -328,6 +342,67 @@ \section{Witnesses}
...
+\begin{figure}[h]
+  \begin{equation}\label{eq:delegation-override}
+    \inference[Delegation-override]
+    {\delegates{c}{\var{vk}_s}{\var{vk}_d}}
+    {\ldelegates{\Gamma; c}{\var{vk_s}}{\var{vk_d}}}
+  \end{equation}
+
+  \begin{equation}\label{eq:delegation-inductive}
+    \inference[Delegation-inductive]
+    {\delegates{c}{\var{vk_s}}{\var{vk_d}} \Rightarrow \var{vk} \neq \var{vk_d}
```
nc6: Do we really want an implication here? This is not giving us `delegates c vk_s vk_d`.

nc6: Also, presumably want `\Rightarrow \var{vk_s} \neq \var{vk_d}`?

nc6 (changes requested):
I have some questions, a couple of suggestions.

nc6:
Also, I would suggest we have something like:

- A "delegation-empty" rule which suggests that the empty chain gives `v_k \ldelegates v_k`

This way, the blockchain layer doesn't need to deal with the case of "no delegate"

specs/ledger/latex/simple-ledger-spec.tex
JaredCorduan: should `vk` be `vk_s`?

specs/ledger/latex/simple-ledger-spec.tex
JaredCorduan: do we use dot-equals elsewhere for denominational equality?  I'm just curious...

specs/ledger/latex/simple-ledger-spec.tex
dnadales: > Also, presumably want \Rightarrow \var{vk_s} \neq \var{vk_d}?

Why would we need this? (sorry I don't follow :grimacing: )

specs/ledger/latex/simple-ledger-spec.tex
dnadales: Yeah, I just wanted to have a way to distinguish equals by definition from equality between keys. Does that make sense?

specs/ledger/latex/simple-ledger-spec.tex
dnadales: Actually `vk_d` should be `vk`. See if the changes I pushed make sense now :)

specs/ledger/latex/simple-ledger-spec.tex
JaredCorduan: I think so.  I was just wondering what was the most normal symbol for this. There's also `:=` and `≡` (\equiv).

specs/ledger/latex/simple-ledger-spec.tex
JaredCorduan: elsewhere we were using union override for this kind of thing.

dnadales: > Let's have the delegation transition defined somewhere.

Isn't the delegation transition what I'm defining here?


> elsewhere we were using union override for this kind of thing.

True. So an alternative formulation would be sth like:

image

Does that make more sense?

dnadales: I don't know how we would write a rule in the alternative formulation that says that if a key is not delegated, then it is in the delegation map. This would require us to know the set of keys, or have some sort of "bootstrap" delegation map (which will contain the genesis keys).

specs/ledger/latex/simple-ledger-spec.tex
JaredCorduan: makes total sense!  I'm used to other notations for this, but looking at wikipedia `\doteq` looks to be normal as well.  also, sorry for the typo above, I meant to write "definitional equality".

specs/ledger/latex/simple-ledger-spec.tex
nc6: @dnadales As per https://github.com/input-output-hk/cardano-chain/pull/12/files#diff-4c9378c26ca5843fc1cb0764ce515451R201, you have these prefigured for utxo transitions, but not for TxWits or delegation. I would like these to be consistent :-)

specs/ledger/latex/simple-ledger-spec.tex
nc6: Yeah, I would be more used to `:=` here

dnadales: I'm fine with both. So do I change `\dot =` to `:=` (the latter makes me think of assignments in Pascal :D)

JaredCorduan: oh, there's also good old `def` above the `=` sign.

dnadales:
@nc6 @JaredCorduan I just pushed a simplified (and now hopefully correct) version of the delegation rules. Could you check if everything makes sense now so that I can merge this, and move on?

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -328,13 +342,77 @@ \section{Witnesses}
...
+
+  \begin{equation}\label{eq:delegation}
+    \inference[Delegation]
+    {c = (vk_s, vk_d)}
+    {d \trans{delegates}{\var{c}} d \unionoverride  \{\var{vk}_s \mapsto \var{vk}_d\}}
+  \end{equation}
+
+  \caption{Delegation inference rules}
+  \label{fig:state-trans-delegation}
+\end{figure}
```
nc6, line 379: Rather strange to have this preceding the definitions for delegation.

nc6 (approved):

PR #13, opened 2018-10-05
[#18] Add cardano-binary package

ruhatch:
Add the `cardano-binary` package, which contains the typeclass `Bi`. `Bi` is a thin-wrapper around the `cborg` library that also provides symbolic size estimates. The `Generic`-deriving stuff is used very little throughout the codebase and we might want to remove it all together. There is also a bunch of `QuickCheck` stuff that could be ported to `hedgehog`, especially round-trip testing in `CborSpec`.


nc6 (approved):
Hard to review this in detail since it's so big and just contains ported code. So I will generally approve with a couple of comments:

- If the generic stuff isn't being used anywhere, I would suggest dropping it now. That way it might be plausible to actually review it if we reintroduce it here.
- One of the requirements going forth will be to perform sensible "deserialisation with offset" - in other words, allowing us to deserialise something and carry around its serialised form for validation purposes. Does this fit well with the `Bi` instances? How simple would it be to update them to handle this?

ruhatch:
@nc6 there is a PR in `cardano-sl` updating the `Bi` code to use offsets. We should be able to manually port that over fairly easily, and we'll be able to review that more easily if it's a modification to the existing code.

There are a couple of places where we use the `Generic` stuff, but I can audit and remove them, since we have the golden tests in place.

Issue #14, opened 2018-10-05
Add cardano-crypto-wrapper package


ruhatch:
The naming of this package is a bit unfortunate, but right now the crypto library [here](https://github.com/input-output-hk/cardano-crypto) is called `cardano-crypto`, so we cannot reuse the name. It might be best to merge these two libraries now that we're going for a multi-repo structure.

The goal of this issue is to pull in the `cardano-sl-crypto` package and clean it up according to the conventions of the legacy-free codebase.

PR #20, opened 2018-10-09
[#14] Add cardano-crypto-wrapper package

ruhatch:
This is an import of the `cardano-sl-crypto` package into the legacy-free codebase. This package could use some more reorganisation, but I thought I'd wait until after we remove the SSC code, as that should simplify the job.


mdimjasevic:
@ruhatch , I'd love to review this. However, there are 5000 lines changed. Would it be possible to make a smaller PR in the future? Or do you have any other suggestion?

I am not sure if and how it is related, but there's a Rust team working on crypto primitives. Are we to use their work or does that come at a latter phase?

ruhatch:
@mdimjasevic Yeah, it's tricky, because there's not an obvious way to split it up! I'd say just give it a skim as we're going to revisit it soon. I've taken the ideas from previous reviews and applied them to this package, so really just look for anything horrifying!

There'll only be one more large PR and then after that we can start having very small, quick PRs to fix little things :)

mdimjasevic (approved):
Rupert, I spent a little time browsing this PR. I can't say I understand much... I learned there are types for natural numbers in Haskell and I find their implementations peculiar.

You do have tests and I'll take it that you ran them and that they pass.

I'm still hoping to get an answer to my question on how is this related to a Rust team working on crypto primitives.

ruhatch:
@mdimjasevic Thanks for the review. RE the Rust implementation, currently this implementation is supposed to be faster, so we'll continue to use it until the Rust one is production-ready. These are also mostly bindings on top of C primitives, so we should be able to transparently replace the underlying implementation, but keep the same high-level types and structure.

PR #21, opened 2018-10-10
[CDEC-617] More precise types and delegation

mdimjasevic:
https://iohk.myjetbrains.com/youtrack/issue/CDEC-617

This PR improves types by making them more precise. Furthermore, delegation is more precisely defined, though one aspect of it is still not addressed: a certificate becomes valid not immediately, but only after a number of blocks. @dcoutts , how does that number relate to number `K` in the spec?

I am still not sure where to plug in data serialisation and deserialisation. Any thoughts?

### DIFF
--- a/specs/chain/latex/blockchain-spec.tex
+++ b/specs/chain/latex/blockchain-spec.tex
@@ -65,6 +65,8 @@
 \newcommand{\BHash}{\type{BHash}}  % block hashes
 \newcommand{\Addr}{\type{Addr}}
 \newcommand{\Slot}{\type{Slot}}
+\newcommand{\GBlock}{\type{GBlock}}
+\newcommand{\RBlock}{\type{RBlock}}
 \newcommand{\Block}{\type{Block}}
 \newcommand{\HCert}{\type{C}}
 
@@ -85,17 +87,19 @@
 \newcommand{\delegatename}{delegate}
 \newcommand{\validname}{valid}
 \newcommand{\keypairname}{pair}
+\newcommand{\precbname}{prec} % preceding block
 %% 
 %% Functions and relations
 %%
-\newcommand{\sign}[4]{\fun{\signname}\ \var{#1} ~ \var{#2} ~ \var{#3} ~ \var{#4}}
-\newcommand{\signed}[2]{\fun{\signedname}\ \var{#1} ~ \var{#2}}
-\newcommand{\signedwindow}[2]{\fun{\signedseqname}\ \var{#1} ~ \var{#2}}
+\newcommand{\sign}[4]{\fun{\signname}\ #1 ~ #2 ~ #3 ~ #4}
+\newcommand{\signed}[2]{\fun{\signedname}\ #1 ~ #2}
+\newcommand{\signedwindow}[2]{\fun{\signedseqname}\ #1 ~ #2}
 \newcommand{\verify}[3]{\fun{\verifyname} ~ #1 ~ #2 ~ #3}
-\newcommand{\hash}[1]{\fun{\hashname}\ \var{#1}}
-\newcommand{\delegate}[1]{\fun{\delegatename}\ \var{#1}}
+\newcommand{\hash}[1]{\fun{\hashname}\ #1}
+\newcommand{\delegate}[1]{\fun{\delegatename}\ #1}
 \newcommand{\valid}[3]{\fun{\validname} ~ #1 ~ #2 ~ #3}
 \newcommand{\pair}[2]{\fun{\keypairname} ~ #1 ~ #2}
+\newcommand{\precb}[1]{\fun{\prevbname} ~ #1}
 
 \begin{document}
 
@@ -152,14 +156,15 @@ design.
   \emph{Derived types}
   %
   \begin{align*}
-    & sk_G \in \SKeyGen \subseteq \SKey & \text{genesis block's signing key}\\
-    & \BHash & \text{block hash}\\
-    & c \in \HCert & \text{heavyweight delegation certificate}\\
-    & sig \in \Sig  & \text{signature}\\
-    & data \in \Data  & \text{data}\\
-    & \Block = \BHash \times \Slot \times \Data \times \Sig & \\
-    & (h, sl, d, \sigma) \in \Block
-      & \text{block}
+    sk_G & \in \SKeyGen \subseteq \SKey & \text{genesis block's signing key}\\
+    h & \in \BHash & \text{block hash}\\
+    c & \in \HCert & \text{heavyweight delegation certificate}\\
+    sig & \in \Sig  & \text{signature}\\
+    data & \in \Data  & \text{data}\\
+    (vk_1, \dotsc, vk_n) & \in \GBlock = \VKey^n & \text{genesis block} \\
+    (h, sl, d, \sigma) & \in \RBlock = \BHash \times \Slot \times \Data \times \Sig
+      & \text{non-genesis block} \\
+    b & \in \Block = \GBlock + \RBlock & \text{block}
   \end{align*}
   %
   \emph{Functions and relations}
@@ -213,7 +218,7 @@ $\text{\verifyname}$ relation is given by \eqref{eq:sign-verify}:
 
 \begin{equation}
   \label{eq:signed-window}
-  \signedwindow{sk}{s} = \Set{b}{b \in s \wedge \signed{sk}{b}}
+  \signedwindow{sk}{\beta} = \Set{b}{b \in \beta \wedge \signed{sk}{b}}
 \end{equation}
 
 \begin{align}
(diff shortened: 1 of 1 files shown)


dnadales:
The first major remark is that what the `signedc` function computes, should be in the states of the transition system: as @dcoutts pointed out, we're aiming for more "operational" rules, meaning that they are a bit closer to the implementation. I also started out in the same way as you did, but I'm changing the ledger rules to comply with this approach.

dnadales:
The second major remark is that the transition rule for the blockchain layer should rely on a delegation transition from the ledger layer.

dnadales (approved):
I'd say incorporate the changes you think make sense and can be incorporated with a small amount of effort, and let's coordinate how are we going to integrate the ledger and chain rules.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -65,12 +65,16 @@
...
+\newcommand{\RBlock}{\type{RBlock}}
 \newcommand{\Block}{\type{Block}}
 \newcommand{\HCert}{\type{C}}
 
 \newcommand{\SKey}{\type{SKey}}
 \newcommand{\SKeyGen}{\type{SKey_G}}
 \newcommand{\VKey}{\type{VKey}}
+\newcommand{\VKeyGen}{\type{VKey_G}}
 \newcommand{\Sig}{\type{Sig}}
 \newcommand{\Data}{\type{Data}}
```
JaredCorduan, line 79: should `Data`, `Sig`, and `BHash` be primitive types?

JaredCorduan (approved):
looks good to me

specs/chain/latex/blockchain-spec.tex
```diff
@@ -130,36 +161,97 @@ \section{Auxiliary definitions}
...
+
+To keep track of active certificates and when they become valid, we use a map
+from genesis block verification keys to a list of certificates.
+%
+A certificate contains an epoch in which it becomes valid.
+%
+Each list is sorted by epochs from certificates in the decreasing order (i.e.,
+a certificate from the first epoch comes last).
+%
+The map is initialised by each genesis block key delegating to itself.
```
dnadales, line 202: Does this mean that we have to use a "dummy" delegation certificate for each genesis key?

specs/chain/latex/blockchain-spec.tex
```diff
@@ -130,36 +161,97 @@ \section{Auxiliary definitions}
...
+
+
+If a new certificate is to be posted and added to the map, first it is
+checked.
```
dnadales, line 206: Which checks are performed?

dnadales: Maybe "the following checks are performed:" ?

specs/chain/latex/blockchain-spec.tex
```diff
@@ -130,36 +161,97 @@ \section{Auxiliary definitions}
...
+twice in an epoch.
+%
+Because one key's delegation can be changed only for the next epoch, but not
+for the current epoch or any epoch after the next, checking whom the key
+currently delegates to is a constant-time operation: only the head and
+possibly the second item in the list have to be examined.
+
+This approach also supports certificate revocation: it is done by posting a
+certificate in which a key delegates to itself.
+
```
dnadales, line 242: I like this approach.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -130,36 +161,97 @@ \section{Auxiliary definitions}
...
+If the certificate is valid, it is prepended to the list of the key that is
+delegating.
+%
+This also takes care of the property that no key can delegate
+twice in an epoch.
+%
+Because one key's delegation can be changed only for the next epoch, but not
+for the current epoch or any epoch after the next, checking whom the key
+currently delegates to is a constant-time operation: only the head and
+possibly the second item in the list have to be examined.
```
dnadales, line 238: One thing to keep in mind is the fact that [we don't want to keep delegation certificates forever](https://input-output-rnd.slack.com/archives/CCRB7BU8Y/p1539593581000100), so that'd mean that we need to do a cleanup after the current epoch changes.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -130,36 +161,97 @@ \section{Auxiliary definitions}
...
 If $s(\cdot)$ is a function for serialising data and $d(\cdot)$ a function for
 deserialising data, their composition $s \circ d$ is not an identity function:
+%
 this is hard to enforce and it is considered a bad practice in cryptography
 design.
 
+\section{Delegation}
+\label{sec:delegation}
+
+The genesis block contains verification keys of $n$ servers.
```
nc6, line 173: We should be clear about what this means. Cardano uses the term "genesis block" to refer to epoch boundary blocks.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -130,36 +161,97 @@ \section{Auxiliary definitions}
...
+%
+However, by the means of delegation certificates, keys from the genesis block
+can delegate signing rights to other keys.
+%
+Delegation is not transitive, i.e., there is no chaining of delegation:
+%
+only genesis block keys can delegate.
+
+
+To keep track of active certificates and when they become valid, we use a map
```
nc6: It's not clear to me what the `we` is here. This is part of the state which exists in doing a scan through the chain? If so, I would have thought it unnecessary to specify here.

specs/chain/latex/blockchain-spec.tex
nc6: Certificates get added to the delegation payload in ledger state, so the criteria here should specify when a valid ledger extension (per the delegation rules in the ledger spec) extends to a valid chain extension.

nc6 (approved):
Looks good, though I think we should be a little more formal with some of the detail here.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: Maybe. I've put them where they are now because that used to be an "abstract types" section, which I renamed later. Right, I wanted to say I'm keeping these 3 types abstract.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: No, there are no dummy delegation certificates.

One can either delegate to somebody else or run their own pool. When someone is delegating to themselves, their delegation certificate is of the form `c = (vk_g, vk_g, e, sig). Such certificates are perfectly normal and valid delegation certificates.

The point is that there is no need for a separate treatment based on whom we're delegating to - be it ourselves or someone else.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: I addressed this today. At most two certificates per delegator are kept: one that is currently valid and possibly another one for the next epoch if changing the delegatee.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: And you would agree that when a boundary block is called the "genesis block" that this is a confusing usage of the term, least to say, right? That's why I believe there should be no need to clarify the term when using it appropriately, but only when using it inappropriately.

If you look at the Ouroboros BFT paper draft, authors don't have a note along the lines: "the genesis block is not a block at an epoch boundary, rather it is the very first block in a block chain."

specs/chain/latex/blockchain-spec.tex
mdimjasevic: Yes, that's part of the state. Where would you specify it?

specs/chain/latex/blockchain-spec.tex
mdimjasevic: @nc6, I am not sure I understand. Do you think this should be part of the ledger spec or the blockchain spec?

specs/chain/latex/blockchain-spec.tex
dnadales: I think there's no notion of "pool" in the Byron version, and thus in our simplified version.

By requiring the map to be initialized by this certificates, we're constraining the chain to include this from the inset. I don't know if this is the case. Maybe @nc6 or @ruhatch know.

specs/chain/latex/blockchain-spec.tex
dnadales: Good. Just keep in mind that the cleanup might not be a constant operation anymore.

mdimjasevic: The cleanup is linear in the number of keys that can delegate. Here only 7 keys from the genesis block can delegate, so it can be considered a constant time operation.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: Ok, I should reword this then. Still, a party can delegate to itself, right?

dnadales:
Don't you need to cleanup the certificates once the current epoch is
changed? . I thought it was linear on the number of certificates.

mdimjasevic:
> Don't you need to cleanup the certificates once the current epoch is changed? . I thought it was linear on the number of certificates.

I explained it somewhere else, but to take a head of a list is a constant-time operation and I have to do this only for keys from the genesis block (there are 7 keys), which makes it a constant-time operation in total.

Issue #15, opened 2018-10-05
Add cardano-chain package


ruhatch:
This package will contain all the types necessary to represent and deserailise the `Block` and `Undo` types from `cardano-sl`. This also involves clean up of that code to bring it in line with the conventions of the legacy-free codebase.

PR #23, opened 2018-10-11
[#15] Add cardano-chain Common modules

ruhatch:
This is the first of a series of PRs to bring the `cardano-chain` package into the codebase. I've split it up, so that hopefully each PR is manageable!

Again this is a first version of this stuff, so you can definitely make it a relatively light review!

### DIFF
--- a/cardano-chain.cabal
+++ b/cardano-chain.cabal
@@ -14,7 +14,167 @@ cabal-version:       >=1.10
 
 library
   hs-source-dirs:      src
-  -- exposed-modules:
-  -- other-modules:
+  exposed-modules:
+                       Cardano.Chain.Block
+                       Cardano.Chain.Common
+                       Cardano.Chain.Constants
+                       Cardano.Chain.Delegation
+                       Cardano.Chain.Genesis
+                       Cardano.Chain.ProtocolConstants
+                       Cardano.Chain.Slotting
+                       Cardano.Chain.Ssc
+                       Cardano.Chain.Txp
+                       Cardano.Chain.Update
+
+  other-modules:
+                       Cardano.Chain.Block.Block
+                       Cardano.Chain.Block.Boundary
+                       Cardano.Chain.Block.Header
+                       Cardano.Chain.Block.Main
+                       Cardano.Chain.Block.SlogUndo
+                       Cardano.Chain.Block.Undo
+                       Cardano.Chain.Block.Util
+
+                       Cardano.Chain.Common.AddrAttributes
+                       Cardano.Chain.Common.AddrSpendingData
+                       Cardano.Chain.Common.AddrStakeDistribution
+                       Cardano.Chain.Common.Address
+                       Cardano.Chain.Common.AddressHash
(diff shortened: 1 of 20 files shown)


cardano-chain.cabal
```diff
@@ -14,7 +14,167 @@ cabal-version:       >=1.10
 
 library
   hs-source-dirs:      src
-  -- exposed-modules:
-  -- other-modules:
+  exposed-modules:
```
dnadales, line 17: I'm a bit scared of the number of modules in this package :D

src/Cardano/Chain/Common/AddrAttributes.hs
```diff
@@ -0,0 +1,95 @@
...
+'Attributes' are conceptually a map, where keys are numbers ('Word8').
+
+For address there are two attributes:
+• 0 — stake distribution, defaults to 'BootstrapEraDistr';
+• 1 — derivation path, defaults to 'Nothing'.
+
+-}
+
+instance Bi (Attributes AddrAttributes) where
+    -- FIXME @avieth it was observed that for a 150kb block, this call to
```
dnadales, line 57: I'm a bit worried that we're carrying over the FIXME's from the legacy code base. What's our plan regarding these?

src/Cardano/Chain/Common/AddrStakeDistribution.hs
```diff
@@ -0,0 +1,120 @@
...
+    -- ^ Stake distribution for bootstrap era.
+    | SingleKeyDistr !StakeholderId
+    -- ^ Stake distribution stating that all stake should go to the given stakeholder.
+    | UnsafeMultiKeyDistr !(Map StakeholderId CoinPortion)
+    -- ^ Stake distribution which gives stake to multiple
+    -- stakeholders. 'CoinPortion' is a portion of an output (output
+    -- has a value, portion of this value is stake). The constructor
+    -- is unsafe because there are some predicates which must hold:
+    --
+    -- • the sum of portions must be @maxBound@ (basically 1);
```
dnadales, line 42: Is this haddock compliant?

dnadales (approved):
Quite a big review :) I looked at it very superficially.

What I'm worried is that we're transferring big portions of the code from `cardano-sl`, which means transferring technical debt as well. Why do we need it? Do we have a plan in place to deal with this?

src/Cardano/Chain/Common/AddrAttributes.hs
ruhatch: We'll take them on a case by case basis, but there shouldn't be too many. These aren't that old if @avieth wrote this one, so we can get him to provide more context.

ruhatch:
@dnadales We will be carefully reviewing all the types to make sure that we're avoiding technical debt. The `binary`, `crypto`, `core`, and `chain` packages from `cardano-sl` were not the source of most of the issues with the old codebase anyway. I've also already removed a lot of stuff from this import. This is (almost) the bare minimum needed to get the `Bi` golden/round-trip testing of all blockchain types working.

PR #24, opened 2018-10-11
Stream (Block,Undo) from epoch files

redxaxder:


cardano-chain.cabal
```diff
@@ -12,13 +12,20 @@ build-type:          Simple
 extra-source-files:  ChangeLog.md, README.md
 cabal-version:       >=1.10
 
+flag development
```
dnadales, line 15: Thank you for including this flag instead of having `-Werror` in all type of builds :)

src/Cardano/Chain/Epoch/File.hs
```diff
@@ -0,0 +1,117 @@
...
+
+-- Epoch file format:
+--
+-- EpochFile := "Epoch data v1\n" *SlotData
+-- SlotData := "blnd" BlockLength UndoLength Block Undo
+-- BlockLength := Word32BE
+-- UndoLength := Word32BE
+-- Block := CBOR
+-- Undo := CBOR
+
```
dnadales, line 37: Some comments in this module's functions would be appreciated :)

src/Cardano/Chain/Epoch/File.hs
```diff
@@ -0,0 +1,117 @@
...
+decodeAll :: (forall s. D.Decoder s a) -> LBS.ByteString -> Either D.DeserialiseFailure a
+decodeAll decoder bytes = case D.deserialiseFromBytes decoder bytes of
+  Left err -> Left err
+  Right (remaining, result) ->
+    if LBS.null remaining
+    then Right result
+    else Left (D.DeserialiseFailure 0 "Deserialisation did not consume all input.")
```
dnadales: Why is the byte offset set to 0?

src/Cardano/Chain/Epoch/File.hs
```diff
@@ -0,0 +1,117 @@
...
+import qualified Codec.CBOR.Decoding as D
+import qualified Codec.CBOR.Read as D
+import           Control.Monad.Trans (MonadTrans (..))
+import qualified Data.Binary as B
+import           Data.Binary.Get (getWord32be)
+import qualified Data.Binary.Get as B
+import qualified Data.ByteString as BS
+import qualified Data.ByteString.Lazy as LBS
+import           Data.Word (Word32)
+import           Streaming.Prelude (Of (..), Stream)
```
dnadales, line 20: Just curious: why are we using `streaming` over `pipes` or `conduit`? Performance only? Or are there any other reasons?

dnadales (dismissed):
LGTM

Any plans on including tests?

src/Cardano/Chain/Epoch/File.hs
ruhatch: I've implemented a version of this function in `cardano-binary`, but it hasn't been merged in yet. We can rebase once it is and remove this.

ruhatch: In fact, you should just be using `decodeFull` from `Cardano.Binary.Class`, which does exactly this, using the `Bi` instance

ruhatch (commented):
Looks good to me, but we shouldn't merge this in `ruhatch/cardano-chain`, as that was just a temporary branch to expose this stuff. We should rebase this on top of `master` once the `cardano-chain` PRs have gone through.

src/Cardano/Chain/Epoch/File.hs
redxaxder: These are the primary reasons:
  - We need to pick one.
  - It doesn't seem like it will be difficult to change later.
  - They all solve the relevant problems. (As far as I know.)

Do you want to make a case for or against one of them?

dnadales: No, like I say, I'm just curious. I never know when the time will come in which I'd need to use a streaming library, and I find it useful to gather the knowledge of more experienced people.

As I pointed out the performance of `streaming` seems to better than that of `pipes` and `conduit`. Also it seems to incorporate fewer concepts.

ruhatch:
This is being merged through #40

PR #26, opened 2018-10-12
Dnadales/cdec 615 model witness for delegation

dnadales:
Adds a model for delegation in the ledger, plus a proposal on how the blockchain layer could use the ledger layer.

I'd like to know which aspects I'm missing or are wrong in the ledger model, so that we can define a list of tasks to complete for having a first draft of the ledger rules that we can send around to other teams for feedback.

Regarding the parts of the blockchain layer I modeled here, currently there are many unknowns regarding where each aspect should be modeled, and with this proposal I wanted to have some concrete grounds to start discussing. So this model was made mostly for my own understanding and for using it in our discussion next week. It can be deleted afterwards.


specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -393,98 +404,513 @@ \section{Delegation}
...
 delegation map $d$ is updated to contain the key mapping
 $\var{vk}_s \mapsto \var{vk}_d$. The symbol $\unionoverride$ denotes
-union-override, and is defined in Figure~\ref{fig:delegation-defs}.
+union-override, and is defined in Figure~\ref{fig:delegation-defs}. Another
+condition that is checked in the delegation rule is that each genesis key can
+issue at most one delegation certificate per epoch. To check this, the states
+of the transitions system maintain a map from epochs to set of keys that were
+delegated on a given epoch. Finally, the delegation epoch cannot be older than
+the current epoch. This is what allow us to clean up the maps of keys delegated
+per epoch. The $\vdash$ symbol denotes an \textit{environment} that remains
```
JaredCorduan: do you mean something like "variables to the left of the turnstile denote an environment..."?

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -393,98 +404,513 @@ \section{Delegation}
...
+    \right)
+  \end{equation*}
+  \emph{Chain extension states}
+  \begin{equation*}
+    \CEState =
+    \left(
+      \begin{array}{r@{~\in~}lr}
+        \beta & \seqof{\Block} & \text{blockchain}\\
+        \var{utxo} & \UTxO & \text{blockchain unspent outputs}\\
+        \var{dmap} & \VKeyGen \mapsto \VKey & \text{blockchain delegation map}\\
```
JaredCorduan, line 896: how is this map related the the one of the same name in `DState`?  Is it that there are two copies of the mapping, which are kept in sync since `BDELEG` uses `DELEG`?

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -393,98 +404,513 @@ \section{Delegation}
...
-    {\Lambda \trans{ledger}{tx}{\Lambda; \var{tx}}}
+    \inference[Seq-delg-ind]
+    { \var{cepoch} \vdash
+      {\left(
+        \begin{array}{r}
+          \var{dmap}\\
+          \var{ekeys}
+        \end{array}
+      \right)}
+      \trans{delegs}{\Gamma}
```
JaredCorduan, line 1002: is this supposed to be `deleg` not `delegs`?

specs/ledger/latex/simple-ledger-spec.tex
dnadales: Yes. What I understand is that inference rules can pass state around. If I'm not mistaken this was already present in the rules that @dcoutts elaborated:

image

Does that make sense? I think such feature is crucial for having modular inference rules.

specs/ledger/latex/simple-ledger-spec.tex
dnadales: That rule transition in the premise should be the same type as in the conclusion, so I guess it is `delegs`. The idea is to define the rule inductively.

specs/ledger/latex/simple-ledger-spec.tex
JaredCorduan: ah, ok, that makes sense.  I didn't realize that the `dmap` was being shared between `DState` and `CEState`.

specs/ledger/latex/simple-ledger-spec.tex
JaredCorduan: yea, I see it now, I was confused

JaredCorduan (approved):

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -133,25 +154,27 @@ \section{Introduction}
...
+a ledger, which is represented here as a list of transactions. The following
 aspects are part of such conditions:
 
 \begin{description}
 \item[Preservation of value] relationship between the total value of input and outputs
   in a new transaction, and the unspent outputs.
 \item[Witnesses] cryptographic entities needed to validate the spending
   of a transaction input.
-\item[Delegation] heavyweight delegation certificates.
+\item[Delegation] delegation certificates, which delegate the block-signing
```
nc6: It seems odd to consider the delegation as a _condition_. It rather seems as a separate part of the payload.

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -393,98 +404,514 @@ \section{Delegation}
...
+\begin{figure}
+  \centering
+  \begin{equation}
+    \label{eq:rule:chain-base}
+    \inference[Chain-base]
+    {}
+    {\left(
+        \begin{array}{l}
+          \epsilon\\
+          \var{\emptyset}\\
```
JaredCorduan: @dnadales I just realized that the `Chain-base` rule needs to allow the initial `utxo` be anything.  perhaps something like:

define `genesisID \in TxId`, and let the conclusion of `Chain-base` allow any instance `utxo \in {genesisId} -> TxOut` instead of the emptyset.

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -393,98 +404,514 @@ \section{Delegation}
...
-    & \Cert = \VKey \times \VKey
-  \end{align*}  
+    \VKeyGen \subseteq \VKey
+  \end{align*}
+  \emph{Delegation states}
+  \begin{equation*}
+    \DState =
+    \left(\begin{array}{r@{~\in~}lr}
+      \var{dmap} & \VKeyGen \mapsto \VKey & \text{delegation map}\\
+      \var{ekeys} & \Epoch \mapsto \powerset{\VKeyGen} & \text{keys delegated per epoch}
```
JaredCorduan: do we need to add `Epoch` to the delegation definitions under abstract types?

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -163,26 +186,55 @@ \section{Preliminaries}\label{sec:preliminaries}
...
+\end{figure}
 
-The states of the UTxO transition system, along with their types are defined in
-Figure~\ref{fig:state-trans-utxo-defs}. Functions on these types are defined in
-Figure~\ref{fig:utxo-auxiliary-ops}.
+Rule~\ref{eq:utxo-inductive} specifies under which conditions a transaction can
+be applied to a set of unspent outputs, and how the set of unspent output
+changes with a transaction:
+\begin{itemize}
+\item The set spent inputs in the transaction, must be in the set of unspent
```
nc6: "set of spend inputs must be a subset of"? Or "Each input spent in the transaction must be in the set of [...]"

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -163,26 +186,55 @@ \section{Preliminaries}\label{sec:preliminaries}
...
+  outputs.
+\item The balance of the unspent outputs in a transaction (i.e. the total
+  amount paid in a transaction) must be equal less than the amount of spent
```
nc6: equal or less

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -277,113 +326,75 @@ \section{State transitions for UTxO}
...
-    %
-    & \addr{} \in \TxOut \mapsto \Hash
-    & \text{output address} \\
-    %
-    & \hash{} \in \VKey \mapsto \Hash
-    & \text{hash function}
+    & \addr{}{} \in \TxIn \mapsto \UTxO \mapsto \Addr\\
+    & \addr{i}{utxo} = \var{hk} \where (hk, \wcard) = \var{utxo}~i
   \end{align*}
   \caption{Definitions associated to the UTxO transition system with witnesses}
```
nc6: Again, should define the transition system `UTXOW` here, including what state it's operating on.

specs/ledger/latex/simple-ledger-spec.tex
dnadales: That makes sense. I'll change `emptyset` to `utxo`. What I don't understand is why do we need the `utxo \in {genesisID} -> TxOut` part. My expectation was that we wouldn't need any constraint at all on UTxO, but I think I'm missing a couple of details :)

specs/ledger/latex/simple-ledger-spec.tex
dnadales: Yes! I'll add this as well. Thanks!

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -277,113 +326,75 @@ \section{State transitions for UTxO}
...
+different order). The choice here is arbitrary.
+
+\begin{figure}
+  \begin{equation}
+    \label{eq:utxo-witness-inductive}
+    \inference[UTxO-wit]
+    { \var{utxo} \trans{utxo}{tx} \var{utxo'}\\ ~ \\
+      & \forall i \in \txins{tx} \cdot
+           \exists (\var{vk}, \sigma) \in \wits{\var{tx}} \cdot
+              \verify{vk}{\serialised{\txbody{tx}}}{\sigma}  \wedge \addr{i}{utxo} = \hash{vk}
```
nc6: This serialisation notation isn't defined anywhere.

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -393,98 +404,515 @@ \section{Delegation}
 blocks in the blockchain.
 
 In the blockchain protocol only a certain number of keys can sign blocks, and
```
nc6, line 489: I don't think any of this is relevant to the _ledger_ part of delegation payload. My preference would be to avoid all discussion of e.g. "the blockchain protocol" in this document, because it's vague and will be encoded elsewhere (in Marko's chain spec)

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -393,98 +404,515 @@ \section{Delegation}
...
 
 The rule for delegation is presented in
 Figure~\ref{fig:state-trans-delegation}. It states that if $\var{c}$ is a valid
 delegation certificate from key $\var{vk}_s$ to key $\var{vk}_d$, then the
 delegation map $d$ is updated to contain the key mapping
 $\var{vk}_s \mapsto \var{vk}_d$. The symbol $\unionoverride$ denotes
-union-override, and is defined in Figure~\ref{fig:delegation-defs}.
+union-override, and is defined in Figure~\ref{fig:delegation-defs}. Another
+condition that is checked in the delegation rule is that each genesis key can
+issue at most one delegation certificate per epoch. To check this, the states
```
nc6, line 505: My intuition is that we should avoid discussing notions of _epoch_ at this level, and leave this for the chain spec. Otherwise we have to introduce quite a few unrelated notions in this document.

nc6 (approved):
I haven't thoroughly reviewed section 5, though I feel we should move it to the chain spec for discussion since it's more rightly describing that system.

specs/ledger/latex/simple-ledger-spec.tex
JaredCorduan: you are right, there are no constraints.  but we do have to pick some arbitrary Ids for the first UTxO, they won't actually be computed from previous transactions, which I was trying to make clear.

specs/ledger/latex/simple-ledger-spec.tex
dnadales: True. What about:

```latex

This specification models the \textit{conditions} that the different parts of a
transaction have to fulfill so that they can extend a ledger, which is
represented here as a list of transactions. In particular, we model the
following aspects:

\begin{description}
\item[Preservation of value] relationship between the total value of input and
  outputs in a new transaction, and the unspent outputs.
\item[Witnesses] authentication of parts of the transaction data by means of
  cryptographic entities (such as signatures and private keys) contained in
  these transactions.
\item[Delegation] validity of delegation certificates, which delegate the
  block-signing rights from a genesis key to another public key.
\item[Update validation] voting mechanism which captures the identification of
  the voters, and the participants that can post update proposals.
\end{description}
```

specs/ledger/latex/simple-ledger-spec.tex
dnadales: Done:

image

I didn't define the states, since as @dcoutts pointed out, these are implicit from the transition types.

However in some rules like this one, I do define the states, since otherwise the notation becomes too cumbersome:

image

specs/ledger/latex/simple-ledger-spec.tex
dnadales: Indeed! We need to discuss how to model "composable serialization", so I'll create a "Serialization" section where we introduce this notation, and model the former concept. [This is](https://iohk.myjetbrains.com/youtrack/issue/CDEC-628) the related issue.

specs/ledger/latex/simple-ledger-spec.tex
dnadales: Hmmm, I don't know. I do explicitly state that we do not model this in this spec, in case someone asks "why aren't the genesis keys modeled here?". If you think this adds more confusion I can remove it.

specs/ledger/latex/simple-ledger-spec.tex
dnadales: That makes sense to me as well, right now we could move the one-certificate-per-epoch constraint from Figure~13 to the blockchain rules:

image

Furthermore, we could even have two blockchain rules, one for constraining the amount of certificates per-epoch, and the other for modeling when a certificate becomes active (in this way we end up with smaller rules).

specs/ledger/latex/simple-ledger-spec.tex
dnadales: Hmmm, you have a good point there. If we leave the utxo unconstrained, then people implementing the executable specs have to choose something. If we constrain it a the rules, then we risk being too restrictive. Do you know if this `genesisID` is something present in the implementation as well?

PR #30, opened 2018-10-15
[#15] Add cardano-chain Delegation and other modules

ruhatch:
Add delegation modules from `cardano-sl`

### DIFF
--- a/src/Cardano/Chain/ProtocolConstants.hs
+++ b/src/Cardano/Chain/ProtocolConstants.hs
@@ -1,0 +1,128 @@
+{-# LANGUAGE DeriveGeneric              #-}
+{-# LANGUAGE FlexibleContexts           #-}
+{-# LANGUAGE FlexibleInstances          #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses      #-}
+{-# LANGUAGE UndecidableInstances       #-}
+
+-- | Protocol constants and some derived terms.
+
+module Cardano.Chain.ProtocolConstants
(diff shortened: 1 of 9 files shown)


src/Cardano/Chain/ProtocolConstants.hs
```diff
@@ -0,0 +1,128 @@
...
+
+       , pcEpochSlots
+       , kEpochSlots
+       ) where
+
+import           Cardano.Prelude
+
+import           Control.Monad.Except (MonadError)
+import qualified Data.Aeson as Aeson (FromJSON (..), ToJSON (..))
+import           Text.JSON.Canonical (FromJSON (..), ToJSON (..))
```
redxaxder, line 34: So there are two sets of `FromJSON` and `ToJSON`. 
The type in this module has an instance for each, and in `Cardano.Chain.Delegation.HeavyDlgIndex` we have:

 - `HeavyDlgIndex`, with an aeson instance, but no canonical-json instance
- `ProxySKHeavy`, with the opposite

What's the deal with these?

ruhatch: The `Aeson` instances were used for logging I believe, but we might remove them all together. The `canonical-json` instances are used for decoding the `GenesisData`, a bunch of initial parameters stored in the genesis block.

redxaxder: Is `Buildable` also used for logging?

ruhatch: Yeah - some decisions either have been or need to made about what to keep

mdimjasevic (approved):

PR #33, opened 2018-10-16
[#15] Add Block and test modules to cardano-chain

ruhatch:
This is the last in the series of initial imports. I have refactored the `Block` types significantly further along my branch, so save a detailed review until then! For now let's get this merged :)


src/Cardano/Chain/Block/Block.hs
```diff
@@ -0,0 +1,413 @@
...
+    (mkMainHeaderExplicit pm prevHash difficulty slotId sk pske body extraH)
+    body
+    extraB
+ where
+  extraB :: MainExtraBodyData
+  extraB = MainExtraBodyData (mkAttributes ())
+  extraH :: MainExtraHeaderData
+  extraH = MainExtraHeaderData bv sv (mkAttributes ()) (hash extraB)
+
+verifyMainBlock :: MonadError Text m => ProtocolMagic -> MainBlock -> m ()
```
dnadales, line 265: I'm wondering why do we need these validation functions here, since I'd expect that we'd write new ones...

test/Test/Cardano/Chain/Update/Gen.hs
```diff
@@ -0,0 +1,242 @@
...
+    <$> Gen.map (Range.linear 0 10) ((,) <$> genPublicKey <*> genVoteState)
+    <*> genProposal pm
+    <*> genSlotId epochSlots
+    <*> genCoin
+    <*> genCoin
+    <*> Gen.maybe genUpsExtra
+
+genUndo :: ProtocolMagic -> SlotCount -> Gen USUndo
+genUndo pm epochSlots =
+  USUndo
```
dnadales, line 166: Sometimes I wonder whether `do` notation (or applicative do in this case) might make code clearer.

dnadales (approved):

src/Cardano/Chain/Block/Block.hs
ruhatch: Yeah - I can just remove them all in another PR

test/Test/Cardano/Chain/Update/Gen.hs
ruhatch: Yeah, or just naming some intermediates, which achieves the same purpose, right?

dnadales: Indeed!

PR #34, opened 2018-10-17
Connects the delegate function to other delegation constructs

mdimjasevic:
This is a minor improvement to my delegation rules proposal.

### DIFF
--- a/specs/chain/latex/blockchain-spec.tex
+++ b/specs/chain/latex/blockchain-spec.tex
@@ -343,6 +343,7 @@ added, the delegation map is updated at $vk_s$, as given by
   {
     {
       \begin{array}{c}
+        (sk_s, vk_s) \in \SKeyGen \times \VKeyGen \\
         (vk_s \mapsto c_h; \lambda) \in \delmap_k \\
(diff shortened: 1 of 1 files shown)


dnadales (approved):
Sent the comments on slack. All in all in today's meeting we could discuss this:

# Agree on notation for the conclusions<a id="sec-1" name="sec-1"></a>

-   `e = ...` in the conclusions is not a notation we've used so far. If we
    want to use it it'd be nice to justify why. I personally prefer not to
    have this, since it involves less symbols (no need to name things). It
    might be confusing as well, since this is what we use in the premises
    for local bindings. But again, I'd be happy to adopt it if there are
    good arguments for this.

# Modeling epoch changes<a id="sec-2" name="sec-2"></a>

Do we do this explicitly by having a clock, or implicitly by using the
epoch in the new blocks.

# Delegation map rules should be in the ledger<a id="sec-3" name="sec-3"></a>

-   Marko's idea is a good one and I can incorporate in the ledger,
    replacing two maps \`dmap\` and \`ekeys\` by one.
-   However we need to be careful with how we model the initial delegation map.

Issue #22, opened 2018-10-10
Safely remove SSC types


ruhatch:
The legacy-free codebase will be using Ouroboros BFT, which doesn't require the complex slot leader schedule of Ouroboros classic. We still need to be able to decode blocks produced by the legacy codebase, so we imported the old SSC types into the new codebase. We can now remove them, but to do so safely we should make sure that old blocks decode at every step. This can be achieved by:

1) Adding a new testing primitive to check for a successful decoding using a given decoder
2) Write decoders for all deprecated types that mimic the operation of the `Bi` instances
3) Add tests for all deprecated types that decode old golden test values
4) Remove all types except `SscPayload` and `SscProof`
5) Write an encoder for an empty `SscPayload` that can be read by the legacy codebase

PR #35, opened 2018-10-17
[#22] Add Dropper to decode SscPayload/Proof

ruhatch:
Add the `Dropper s` type to decode CBOR, throwing away the results. This allows us to traverse the structure of the old blockchain, ignoring Haskell structure that we no longer care about.


binary/src/Cardano/Binary/Class/Drop.hs
```diff
@@ -0,0 +1,59 @@
+-- | A 'Dropper s' is a 'Decoder s ()', that is a decoder that returns nothing
```
dnadales, line 1: "that is, a decoder ..." (missing comma?)

binary/src/Cardano/Binary/Class/Drop.hs
```diff
@@ -0,0 +1,59 @@
...
+       , dropSet
+       , dropTuple
+       , dropTriple
+       , dropWord64
+       ) where
+
+import           Cardano.Prelude
+
+import qualified Codec.CBOR.Decoding as D
+
```
dnadales, line 21: Extra space?

binary/src/Cardano/Binary/Class/Drop.hs
```diff
@@ -0,0 +1,59 @@
...
+dropBytes :: Dropper s
+dropBytes = void D.decodeBytesCanonical
+
+-- | Drop a list of values using the supplied `Dropper` for each element
+dropList :: Dropper s -> Dropper s
+dropList dropElems = do
+  D.decodeListLenIndef
+  D.decodeSequenceLenIndef const () id dropElems
+
+dropMap :: Dropper s -> Dropper s -> Dropper s
```
dnadales, line 34: It seems obvious what `dropMap`, `dropSet`, etc do, but I was wondering whether we need to put the comments, just for the sake of consistency. I'm fine with not having them though.

binary/src/Cardano/Binary/Class/Primitive.hs
```diff
@@ -112,12 +109,12 @@ serializeWith firstChunk nextChunk =
 -- | Deserialize a Haskell value from the external binary representation
 --   (which must have been made using 'serialize' or related function).
 --
---   /Throws/: @'CBOR.Read.DeserialiseFailure'@ if the given external
+--   /Throws/: @'Read.DeserialiseFailure'@ if the given external
```
dnadales, line 112: Hmmm, do we want to make the error explicit in the types somehow, or are we ok with this?

binary/src/Cardano/Binary/Class/Primitive.hs
```diff
@@ -140,56 +137,42 @@ deserialize' = deserialize . BSL.fromStrict
...
+  :: (forall s . D.Decoder s a)
+  -> BSL.ByteString
+  -> Either (Read.DeserialiseFailure, BS.ByteString) (a, BS.ByteString)
+deserialiseDecoder decoder bs0 =
+  runST (supplyAllInput bs0 =<< Read.deserialiseIncremental decoder)
+
+supplyAllInput
+  :: BSL.ByteString
+  -> Read.IDecode s a
+  -> ST
```
dnadales, line 167: This is a weird indentation...

crypto/src/Cardano/Crypto/SecretSharing.hs
```diff
@@ -116,20 +126,29 @@ newtype Secret = Secret
...
+
 -- | Encrypted share which needs to be decrypted using 'VssKeyPair' first.
 newtype EncShare = EncShare
   { getEncShareVal :: Scrape.EncryptedSi
   } deriving (Show, Eq)
 
 deriving instance Bi EncShare
 
+dropEncShare :: Dropper s
+dropEncShare = dropBytes
```
dnadales, line 150: We have this pattern in 4 places, why not using `dropBytes` directly?

crypto/test/Test/Cardano/Crypto/Bi.hs
```diff
@@ -217,69 +228,77 @@ roundTripProxyCertAeson = eachOf 100 genUnitProxyCert roundTripsAesonBuildable
...
 golden_DecShare = goldenTestBi decShare "test/golden/DecShare"
-  where
-    Just decShare = case sharedSecretData of
-        (_, _, xs) -> deterministic "ds" . uncurry decryptShare <$> fmap fst (uncons xs)
+ where
+  Just decShare = case sharedSecretData of
+    (_, _, xs) ->
+      deterministic "ds" . uncurry decryptShare <$> fmap fst (uncons xs)
+
+golden_legacy_DecShare :: Property
```
dnadales, line 299: I'm curious about when do we use underscore and when camel-case. Maybe this was inherited from QuickCheck props, but it'd be nice to use only one style.

dnadales (approved):

binary/src/Cardano/Binary/Class/Primitive.hs
ruhatch: This is currently only used in a couple of tests - I imagine we'll remove it entirely as it's a bit weird

binary/src/Cardano/Binary/Class/Drop.hs
ruhatch: Nope - have been doing two NLs between imports and code start :)

binary/src/Cardano/Binary/Class/Primitive.hs
ruhatch: Yeah, not sure how I would make it better though - any suggestions?

crypto/src/Cardano/Crypto/SecretSharing.hs
ruhatch: I ended up dropping it in a later PR :)

crypto/test/Test/Cardano/Crypto/Bi.hs
ruhatch: You're right - I'll make an issue to change these across the whole codebase

binary/src/Cardano/Binary/Class/Primitive.hs
dnadales: Well, maybe in some cases we don't have to be that strict about the 80 chars limit :) That's the only suggestion I can think of...

PR #36, opened 2018-10-17
using multiple files for the simple ledger latex

JaredCorduan:
I've split the simple ledger latex document up into five files.  This will allow us to use a large part of this spec in the [fm-ledger-rules](https://github.com/input-output-hk/fm-ledger-rules) repo.

The new files, `frontmatter.tex`, `intro.tex`, `delegation.tex`, and `blockchain.tex` are exact copies of those sections from `simple-ledger-spec.tex`.

In order to use `simple-ledger-spec.tex` in `fm-ledger-rules`, I did have to make a few changes besides just moving things into separate files.  I added a definition to the introduction for fibre.  Even though this definition will not be used for both specs, I am including it here since I imagine that the notation from this intro will soon be making its way into a standalone notation document. I also added a bibliography.

If we decide that we do in fact like this separation, we can then decide how to share `simple-ledger-spec.tex` between the two repos in a reasonable way.


specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -480,115 +482,7 @@ \subsection{Witnesses}
...
-      \right)
-    }
-  \end{equation}
-  \caption{Delegation inference rules}
-  \label{fig:state-trans-delegation}
-\end{figure}
+\input{delegation.tex}
 
 \subsection{Witnesses}
 \label{sec:delegation-witnesses}
```
JaredCorduan: maybe it's too weird that this subsection is in the main file.  I did this since the section was mostly fine for both specs.

nc6:
My feeling is this probably goes slightly further than needed - e.g. splitting out `blockchain.tex` and `delegation.tex` makes perfect sense, but having the frontmatter and introduction split out doesn't. Otherwise a :+1: from me!

We should definitely work out how to package our own style with common definitions etc if we go down this route, though!

nc6 (approved):

dnadales:
> We should definitely work out how to package our own style with common definitions etc if we go down this route, though

Yes please! Do we add this as an issue in our workstream?

dnadales:
I guess we miss the "references.bib".

```text
=== TeX engine is 'pdfTeX'
Latexmk: Non-existent bbl file 'simple-ledger-spec.bbl'
 No file simple-ledger-spec.bbl.
Latexmk: Log file says output to 'simple-ledger-spec.pdf'
Latexmk: Failed to find one or more bibliography files
  references.bib
Latexmk: 'pdflatex': source file 'simple-ledger-spec.bbl' doesn't exist. I'll try making it...
------------
Running 'make "simple-ledger-spec.bbl"'
------------
make[1]: Entering directory '/home/damian/github/input-output-hk/cardano-chain/specs/ledger/latex'
make[1]: *** No rule to make target 'simple-ledger-spec.bbl'.  Stop.
make[1]: Leaving directory '/home/damian/github/input-output-hk/cardano-chain/specs/ledger/latex'
Latexmk: All targets (simple-ledger-spec.pdf) are up-to-date
```

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -14,7 +14,7 @@
 \usepackage{xcolor}
 \usepackage[capitalise,noabbrev,nameinlink]{cleveref}
 \hypersetup{
-  pdftitle={A Simplified Formal Specification of a UTxO Ledger},
+  pdftitle={Ledger Specification},
```
dnadales: Maybe we need to rename the file as well. Also, don't we need to be more specific about which ledger we're modelling? Should we call this "Byron Ledger"? (that sounds terrible though).

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -56,6 +56,7 @@
 \newcommand{\restrictdom}{\lhd}
 \newcommand{\subtractdom}{\mathbin{\slashed{\restrictdom}}}
 \newcommand{\restrictrange}{\rhd}
```
dnadales: I'm thinking a next step would be to extract all these in a LaTeX package, and later on we can try to add this as a nix package...

specs/ledger/latex/simple-ledger-spec.tex
```diff
@@ -194,7 +159,10 @@ \section{Notation}\label{sec:notation}
   $\Lambda$.
 \item[Functions] $A \to B$ denotes a \textbf{total function} from $A$ to $B$.
   Given a function $f$ we write $f~a$ for the application of $f$ to argument
-  $a$. 
+  $a$.
+\item[Fibre] Given a function $f: A \to B$ and $b\in B$, we write
```
dnadales: Do we use `fibre` here? If not maybe we can bundle all the definitions in a separate doc, and extract (copy+paste) the ones we need from that document.

dnadales (approved):
I agree with Nick's comments. For the rest it looks good. Thanks a lot for this!

JaredCorduan:
> My feeling is this probably goes slightly further than needed - e.g. splitting out `blockchain.tex` and `delegation.tex` makes perfect sense, but having the frontmatter and introduction split out doesn't. Otherwise a  from me!
> 
> We should definitely work out how to package our own style with common definitions etc if we go down this route, though!

originally I only split out the two files that you mentioned.  the trouble is that I will probably not want my introduction and things like date to be exactly the same as this one. My hope was to be able to use the exact same file for both specs.

specs/ledger/latex/simple-ledger-spec.tex
JaredCorduan: > Maybe we need to rename the file as well. Also, don't we need to be more specific about which ledger we're modelling? Should we call this "Byron Ledger"? (that sounds terrible though).

I agree about renaming the file!  Perhaps I can move the `\hypersetup` to the frontmatter file so that we can be more specific with the pdftitle.

specs/ledger/latex/simple-ledger-spec.tex
JaredCorduan: we do not use it in this spec, but I do use it in mine.  I'm fine with your suggestion.  I was guessing that these definitions will be moving soon anyway, to our standalone notation/definition document.

dnadales: Then, if you prefer, leave it for the time being :)

PR #37, opened 2018-10-18
iohk latex package

JaredCorduan:
here is my first attempt at packing some common latex commands that we've been using.

### DIFF
--- a/specs/ledger/latex/iohk.sty
+++ b/specs/ledger/latex/iohk.sty
@@ -1,0 +1,41 @@
+\NeedsTeXFormat{LaTeX2e}
+\ProvidesPackage{iohk}[2018/10/18]
+
+\RequirePackage{amssymb}
+\RequirePackage{extarrows}
+\RequirePackage{slashed}
+\RequirePackage[tikz]{bclogo}
+\RequirePackage{stmaryrd}
+
+\DeclareMathOperator{\dom}{dom}
+\DeclareMathOperator{\range}{range}
+
+\newcommand{\powerset}[1]{\mathbb{P}~#1}
+\newcommand{\restrictdom}{\lhd}
+\newcommand{\subtractdom}{\mathbin{\slashed{\restrictdom}}}
+\newcommand{\restrictrange}{\rhd}
+\newcommand{\subtractrange}{\mathbin{\slashed{\restrictrange}}}
+\newcommand{\union}{\cup}
+\newcommand{\unionoverride}{\mathbin{\underrightarrow\cup}}
+\newcommand{\uniondistinct}{\uplus}
+\newcommand{\var}[1]{\mathit{#1}}
+\newcommand{\fun}[1]{\mathsf{#1}}
+\newcommand{\type}[1]{\mathsf{#1}}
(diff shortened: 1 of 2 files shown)


nc6 (approved):

specs/ledger/latex/iohk.sty
```diff
@@ -0,0 +1,41 @@
...
+\newcommand{\uniondistinct}{\uplus}
+\newcommand{\var}[1]{\mathit{#1}}
+\newcommand{\fun}[1]{\mathsf{#1}}
+\newcommand{\type}[1]{\mathsf{#1}}
+\newcommand{\signed}[2]{\llbracket #1 \rrbracket_{#2}}
+\newcommand{\size}[1]{\left| #1 \right|}
+\newcommand{\trans}[2]{\xlongrightarrow[\textsc{#1}]{#2}}
+\newcommand{\seqof}[1]{#1^{*}}
+\newcommand{\nextdef}{\\[1em]}
+\newcommand{\where}{~ ~ \mathbf{where}~ ~ }
```
dnadales: Oh, you can kill `where` with fire. I added it, but I think we don't need it :)

JaredCorduan: :hocho:

specs/ledger/latex/iohk.sty
```diff
@@ -0,0 +1,41 @@
+\NeedsTeXFormat{LaTeX2e}
+\ProvidesPackage{iohk}[2018/10/18]
+
+\RequirePackage{amssymb}
+\RequirePackage{extarrows}
```
dnadales, line 5: Do we want to pull `semantics` in?

JaredCorduan: I was :thinking: a bit about that, maybe have a command for our rules.  But I wasn't sure we had much more than was already just in there already.  I guess it would save us an import?  we just use `\inference`, right?

dnadales: Yes, and I setup the space between premises. But feel free to leave `semantics` out if you think that's sensible.

dnadales (approved):
Thanks!

mgudemann (approved):
LGTM

specs/ledger/latex/iohk.sty
JaredCorduan: It's unclear to me if I can do that without renaming it, but maybe that's better anyway.  How about `\rule`?

dnadales: Hmm, I don't know. If we cannot include the `semantics` package and use it, I'd say leave it out. Let's not introduce our own custom macros for rules, since they're well documented in the semantics package.

JaredCorduan: I just tried to see the difference between using `\setpremisesspace{20pt}` or not, just so that I would know if I got it right in the latex package.  But my eyes can't tell the difference... :(

dnadales: You're right. It doesn't seem to make a difference. It does if you set it to 30 for instance (but we don't want that :)).

PR #39, opened 2018-10-18
[#22] Remove deprecated SSC types

ruhatch:
After dropping the `SscPayload` and `SscProof` we can remove all the types that we're only used in that part of the code and our dependency on `pvss` :+1:


redxaxder (commented):
looks good to me

dnadales (approved):

PR #40, opened 2018-10-18
[#16] Read epoch files

redxaxder:


src/Cardano/Chain/Epoch/File.hs
```diff
@@ -0,0 +1,81 @@
...
+    (h :> rest) <- lift $ SBS.toLazy $ SBS.splitAt len bytes
+    if h == header
+      then rest
+      else lift $ throwError (ParseErrorMissingHeader file)
+
+parseEpochFile :: FilePath -> Stream (Of (Block, Undo)) (ExceptT ParseError ResIO) ()
+parseEpochFile file =
+  let bytes = loadFileWithHeader file epochHeader
+      liftDecoderError (Right a) = pure a
+      liftDecoderError (Left e)  = throwError (ParseErrorDecoder e)
```
ruhatch, line 62: Could do `either (throwError . ParseErrorDecoder) pure` if you like that style?

ruhatch (approved):
Looks great, thanks Alex!

PR #41, opened 2018-10-19
Rename Cardano.Prelude.Parse to Base16

erikd:
Old name was too general.


dnadales (approved):
:confetti_ball: Thanks.

PR #42, opened 2018-10-19
[CDEC-617] Function notation improvement in terms of aliases

mdimjasevic:
This simply introduces two alieses:

```latex
% Partial and total function aliases
\newcommand{\totalf}{\to}
\newcommand{\partialf}{\mapsto}
```

Given that I made this PR on the same branch as another PR, the difference tab shows changes that were merged in that other PR too, which I don't know how to hide without pulling in other stuff that got merged to `master` in the meantime.


dnadales (approved):
The objection I have is that this reads a bit awkward: `A \to B` reads more natural than `A \totalf B`.

mdimjasevic:
You're right, but this one has two advantages over using these symbols directly:

* the name communicates better what this is about (given we don't have types in TeX or something alike to do this)
* in case we change our mind at some point on notation, it will be quite easy to update the notation. It will boil down to updating aliases and that's it.

Issue #27, opened 2018-10-12
Safely remove Boundary blocks


ruhatch:
Similarly to the removal of SSC types, we want to remove epoch boundary blocks, while maintaining compatibility with the old on-the-wire format. To do that we encode blocks, wrapped with the old tag for `MainBlock`s and we decode by either dropping boundary blocks or decoding `MainBlock`s as before.

To be able to check hashes of old style blocks, we only need to be able to hash the serialized representation of the header. We should be able to extract this using the new CBOR decoding-with-offsets work, without ever decoding to a Haskell type.

PR #44, opened 2018-10-19
[#27] Safely remove Boundary blocks

ruhatch:
Remove boundary blocks, using a `Dropper` to decode ones from the Byron blockchain. Once they are removed the `GenericBlock` becomes obsolete, so we can simplify all the block types a lot.


nc6:
The changes here look reasonable; however, I'm not certain that it's plausible to do this. While boundary blocks are not distributed, they are concretely a part of the blockchain: their hashes are used as the parent hash for subsequent blocks. If we drop them entirely, then the chain will not be valid by the most basic of validation semantics (unless we explicitly drop this rule).

Maybe we should discuss this more widely.

nc6:
Answered in meeting :-)

erikd:
> Answered in meeting :-)

@nc6 And the answer was?

erikd:
> Answered in meeting :-)

@nc6 And the answer was?

nc6 (approved):

nc6 (approved):

nc6:
@erikd A few options: either we overhaul the validator to make it operate over the parse structure (since it can then peek back to the underlying bytestring), or support a "trees that grow" like structure, where we map over the structure to pull in the raw bytestrings. I think there is a fair bit of detail to work out here, but I was reassured that this is at least understood and there is some plan in place.

ruhatch:
@erikd @nc6 Alex and I discussed this further and hope to return a data structure plus annotations that would contain hashes for validation. So a `Decoder s (Annotated Hash ())`  for boundary blocks and a `Decoder s (Annotated BlockValidationData Block)` for normal blocks.

PR #51, opened 2018-10-22
adding fees to ledger spec

### LINKED ISSUE #50: models fee in the leger
We should model the transaction fees in the ledger spec.

JaredCorduan:
closes #50

### DIFF
--- a/specs/ledger/latex/blockchain.tex
+++ b/specs/ledger/latex/blockchain.tex
@@ -22,6 +22,14 @@ after applying the transitions in a block. Rules for delegation and unspent
 outputs in the context of a block are given in
 Sections~\ref{sec:block-delegation} and \ref{sec:block-utxo} respectively.
 
+\begin{note}
+  The protocal constants are currently modeled as a part of the
+  chain extension environment. Once we adding voting
+  (the mechanism which controls the protocol constants)
+  to the model, it will probably make sense to move them to the
+  chain extension states.
+\end{note}
+
 \begin{figure}
   \emph{Abstract types}
   \begin{equation*}
@@ -53,6 +61,7 @@ Sections~\ref{sec:block-delegation} and \ref{sec:block-utxo} respectively.
         \var{K} & \mathbb{N} & \text{number of nodes}\\
         \var{t} & \mathbb{Q} & \text{byzantine nodes ratio}\\
         \var{d} & \Epoch & \text{delegation liveness parameter}\\
+        \var{pc} & \PrtclConsts & \text{protocol constants}\\
       \end{array}
     \right)
   \end{equation*}
@@ -138,6 +147,7 @@ Sections~\ref{sec:block-delegation} and \ref{sec:block-utxo} respectively.
       }
       \\ ~ \\
       {
+        \var{pc} \vdash
         \left(
           \begin{array}{l}
             \var{utxo}\\
@@ -158,7 +168,8 @@ Sections~\ref{sec:block-delegation} and \ref{sec:block-utxo} respectively.
          \Gkeys\\
          K\\
          t\\
-         d
+         d\\
+         pc
       \end{array}}
       \vdash
       {
@@ -240,7 +251,7 @@ is shown in Figure~\ref{fig:rules:delegation-sequence}.
   \end{equation*}
   \emph{Block delegation transitions}
   \begin{equation*}
-    \_ \vdash \_ \trans{chain}{\_} \_ \in
+    \_ \vdash \_ \trans{bdeleg}{\_} \_ \in
       \powerset (\BDEnv \times \BDState \times \Block \times \CEState)
   \end{equation*}
   \caption{Block delegation transition-system types}
--- a/specs/ledger/latex/blockchain.tex
+++ b/specs/ledger/latex/blockchain.tex
@@ -23,11 +23,21 @@ outputs in the context of a block are given in
 Sections~\ref{sec:block-delegation} and \ref{sec:block-utxo} respectively.
 
 \begin{note}
-  The protocal constants are currently modeled as a part of the
+  The protocol constants are currently modeled as a part of the
   chain extension environment. Once we adding voting
   (the mechanism which controls the protocol constants)
   to the model, it will probably make sense to move them to the
-  chain extension states.
+  chain extension states. The protocol constants will also be
+  added to the ledger state, along with some kind of rule which
+  combines voting with the UTxO rule:
+  \begin{equation*}
+    \inference[Voting-combine]
+    {
+      \var{pc} \trans{voting}{} \var{pc'} &
+      \var{pc'} \vdash \var{utxo} \trans{utxo}{} \var{utxo'}\\ ~ \\
+    }
+    {\ldots}
+  \end{equation*}
 \end{note}
 
 \begin{figure}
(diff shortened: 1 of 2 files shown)


specs/ledger/latex/ledger-spec.tex
```diff
@@ -300,8 +306,8 @@ \section{UTxO}
 \begin{figure}
   \begin{equation}\label{eq:utxo-inductive}
     \inference[UTxO-inductive]
-    { \txins{tx} \subseteq \dom \var{utxo}
-      & \balance{(\txouts{tx})} \leq \balance{(\txins{tx} \restrictdom \var{utxo})}
+    { \txins{tx} \subseteq \dom \var{utxo} & \minfee{tx} \leq \txfee{tx}\\
```
dnadales: According [this message](https://input-output-rnd.slack.com/archives/CCZSDE005/p1540233076000100), it seems `minfee` depends on parameters outside the transaction `tx`, called `a` and `b`. So it seems we'd need to add a transition environment, that contains `a` and `b` and model this is `minfee a b tx` (I need to look what `a` and `b` are).

specs/ledger/latex/ledger-spec.tex
```diff
@@ -325,6 +331,15 @@ \section{UTxO}
   $\var{tx}$.
 \end{itemize}
 
+\begin{note}
```
dnadales, line 339: That's indeed a good point! So what you're saying is that if we'd take coin to be `Z` then we could spend more than what we have by having negative outputs, right? Maybe we should either put constraints on the `Coin` type (e.g. commutative monoid), or specify in the rule that all values should be `>= 0` (but the `Coin` is not abstract anymore). Should we try to fix this in this pull request?

dnadales (commented):
Thanks a lot for this!

I think we should address:

0. Modelling the parameters on which the `minfee` function depends.
1. Restrict the possible instances of `Coin`.

specs/ledger/latex/ledger-spec.tex
nc6: At the moment, `Coin` is given no specification (which is a problem since we try to sum over it in the `balance` operation!) I feel we should constrain the coin type.

nc6 (approved):

nc6:
I had no comments aside from Damian's, so giving approval. I think we should fix the `minfee` params, though I'd be happy to address the `Coin` issue in a separate PR.

specs/ledger/latex/ledger-spec.tex
```diff
@@ -242,7 +244,11 @@ \section{UTxO}
       \txid{} & \Tx \to \TxId & \text{compute transaction id}\\
       %
       \fun{txbody} & \Tx \to \powerset{\TxIn} \times (\Ix \mapsto \TxOut)
-                                  & \text{transaction body}
+                                  & \text{transaction body}\\
+      %
+      \fun{txfee} & \Tx \to \Coin & \text{transaction fee}\\
+      %
+      \fun{minfee} & \Tx \to \Coin & \text{minimum fee}
```
mgudemann: according to a message from Duncan, close to the Damien's link below, we might make this a function from the size of `Tx` to `Coin`, as it depends solely on the size and not the content of the transaction.

mgudemann: that is having type `N -> Coin`

dnadales:
> I had no comments aside from Damian's, so giving approval. I think we should fix the `minfee` params, though I'd be happy to address the `Coin` issue in a separate PR.

Sounds good @nc6! I've created an [issue](https://github.com/input-output-hk/cardano-chain/issues/54) about this.

specs/ledger/latex/ledger-spec.tex
JaredCorduan: yea, that is my preference as well, to make it essentially N.  Commutative monoid is unfortunately not enough (Z meets that as well).  I'm happy to change this here. I just put it as a note since I wasn't sure if it required a larger conversation.  Eventually, once we introduce Chimeric types and therefore need subtraction or inverses, the decision will be trickier.

specs/ledger/latex/ledger-spec.tex
JaredCorduan: that's a great point @dnadales , thanks.

specs/ledger/latex/ledger-spec.tex
JaredCorduan: I left it as `Tx` since that we the spec reads the same whether or not the fee is implicit (ie the difference) or explicit.

JaredCorduan: but perhaps it is time to let go of trying to keep this ledger spec the same as the others yet to come.

dnadales (approved):

JaredCorduan:
@dnadales - thanks again for catching the missing protocol constants!  I've added them to the model, including the chain extension rule. it's probably worth another good look, it is not as minimal a change as it was before...

specs/ledger/latex/blockchain.tex
```diff
@@ -22,6 +22,14 @@ \subsection{Chain extension}
 outputs in the context of a block are given in
 Sections~\ref{sec:block-delegation} and \ref{sec:block-utxo} respectively.
 
+\begin{note}
+  The protocal constants are currently modeled as a part of the
```
dnadales: Protocal sounds like the first calendar humans ever used :P (sorry it is early in the morning)

specs/ledger/latex/blockchain.tex
```diff
@@ -22,6 +22,14 @@ \subsection{Chain extension}
...
+  chain extension environment. Once we adding voting
+  (the mechanism which controls the protocol constants)
+  to the model, it will probably make sense to move them to the
+  chain extension states.
+\end{note}
```
dnadales, line 41: Actually I would say that the chain extension state will include these "constants", but the ledger rules that model voting and updates will have them as well.

specs/ledger/latex/ledger-spec.tex
```diff
@@ -306,10 +310,11 @@ \section{UTxO}
 \begin{figure}
   \begin{equation}\label{eq:utxo-inductive}
     \inference[UTxO-inductive]
-    { \txins{tx} \subseteq \dom \var{utxo} & \minfee{tx} \leq \txfee{tx}\\
-      \balance{(\txouts{tx})}  + \txfee{tx} = \balance{(\txins{tx} \restrictdom \var{utxo})}
+    { \txins{tx} \subseteq \dom \var{utxo} & \minfee{pc}{tx} \leq \txfee{tx}\\
+      \balance{(\txouts{tx})}  + \txfee{tx} =
```
dnadales, line 314: So the only `minfee` depends on the protocol (fee) constants, right?

dnadales (approved):
Looks good! Thanks!

specs/ledger/latex/blockchain.tex
JaredCorduan: eventually outdated by cro-magnon-cal :calendar:

specs/ledger/latex/ledger-spec.tex
JaredCorduan: as of now, yes

specs/ledger/latex/blockchain.tex
JaredCorduan: So would you reword this as

> Once we adding voting (the mechanism which controls the protocol constants) to the model, it will probably make sense to move them to the chain extension states and add them to the rules involving votes

?

PR #55, opened 2018-10-23
[CDEC-638] Adds a description of the delegation interface

mdimjasevic:
This adds the description of the delegation interface between the ledger layer and the blockchain layer.

### DIFF
--- a/specs/chain/latex/blockchain-spec.tex
+++ b/specs/chain/latex/blockchain-spec.tex
@@ -8,6 +8,7 @@
 \usepackage[capitalise,noabbrev,nameinlink]{cleveref}
 \usepackage[utf8]{inputenc}
 \usepackage{latexsym}
+\usepackage{listings}
 \usepackage{mathpazo} % nice fonts
 \usepackage{mathtools}
 \usepackage{microtype}
@@ -174,6 +175,11 @@ design.
 \section{Delegation}
 \label{sec:delegation}
 
+\marko{This section might be out of date. There is a work in progress to come
+  up with an interface between the blockchain layer and the ledger layer.
+  %
+  See CDEC-638.}
+
 The genesis block contains verification keys of $n$ servers.
 %
 As of mid-October 2018, $n = 7$.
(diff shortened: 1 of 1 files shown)


specs/chain/latex/blockchain-spec.tex
```diff
@@ -245,6 +251,90 @@ \subsection{Delegation Design}
...
+valid and possibly invalidate other certificates in the future.
+%
+As part of the delegation interface, the ledger layer exposes functionality
+for checking if a particular verification key has the delegation right to sign
+a block in the current slot.
+%
+In particular, if the verification key has the right to sign, the
+functionality should inform who is the delegator.
+%
+In Haskell, the signature of the corresponding function is:
```
dnadales, line 283: "could be"? Since I guess there are multiple ways of modelling this.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -245,6 +251,90 @@ \subsection{Delegation Design}
...
+%
+\begin{lstlisting}
+delegator :: (VKey vk_d, VKeyGen vk_s) => vk_d -> Maybe vk_s
+\end{lstlisting}
```
dnadales, line 287: It'd be nice to have an equivalent formulation in terms of operational rules. Which are the types of the states that you envision? Then we can see how to model this in the ledger.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -245,6 +251,90 @@ \subsection{Delegation Design}
...
+
```
dnadales, line 288: So it seems that the map from `VKeyGen` to `VKey` contains the information needed to implement `delegator` already, doesn't it?

specs/chain/latex/blockchain-spec.tex
```diff
@@ -245,6 +251,90 @@ \subsection{Delegation Design}
...
+blockchain layer has to inform the ledger layer of new certificates when a new
+block is added to the blockchain.
+%
+Therefore, as part of the delegation interface, the blockchain layer exposes
+new delegation certificates functionality to the ledger layer.
+%
+This functionality should communicate at least the following information for
+each new delegation certificate:
+%
+1) the delegator's verification key,
```
dnadales: Maybe use `enumerate`?

specs/chain/latex/blockchain-spec.tex
```diff
@@ -309,6 +399,10 @@ \subsection{Properties}
 \subsubsection{Delegation}
 \label{sec:delegation-properties}
 
+\marko{This whole subsection on delegation rules is out of date.
```
dnadales, line 404: Maybe we can remove them then, since we keep the history of this document in git anyway ...

specs/chain/latex/blockchain-spec.tex
mdimjasevic: I bet there are multiple ways. This is what I came up with.

dnadales (approved):
What I'm missing is a proposal in terms of structural operational rules. I find it hard to know how should I modify the ledger layer so that the blockchain layer can use it, since I don't have blockchain rules to look at (to check for consistency).

I'm guessing how the blockchain layer could use the ledger layer in this sketch [here](https://input-output-rnd.slack.com/archives/CCRB7BU8Y/p1540294105000100?thread_ts=1540292793.000100&cid=CCRB7BU8Y). Is this what you had in mind?

I apologize for requesting such level of detail, but I'm usually quite slow in grasping these things :)

specs/chain/latex/blockchain-spec.tex
mdimjasevic: Agreed.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: That's in my plan to do next.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: Thanks! I'm putting it into `enumerate`.

mdimjasevic:
Damian, I will add rules next.

However, how I see this interface, the interface should be blind to how are specifics handled at either side of the interface, meaning it should know nothing about rules specific to the blockchain layer or to the ledger layer.

PR #57, opened 2018-10-23
[#56] Switch to `cardano-prelude` from external repo

### LINKED ISSUE #56: Use `cardano-prelude` from it's own repo
After moving `cardano-prelude` to an eponymous repo we need to make sure we can build referring to a specific git commit. We should also put the custom `stack` snapshot in this repo and other projects can refer to it via it's URL.

ruhatch:
Closes #56


parsonsmatt (approved):

nc6 (approved):
I'd really rather not have these additional stack.yaml files in subdirs, since they're just another place which needs to be kept in sync or else have non-intuitive diverging build results. But this isn't technically related to this PR, which in itself looks fine.

PR #58, opened 2018-10-24
Small step semantics

nc6:


specs/semantics/latex/small-step-semantics.tex
```diff
@@ -0,0 +1,265 @@
...
+\item [States] The underlying datatype of our system, whose validity we are
+  required to prove.
+\item [Transitions] The means by which we might move from one (valid) state to
+  another.
+\item [Signals] The means by which transitions are triggered.
+\item [Rules] Formulae describing how to derive valid states or transitions. A
+  rule has an \textit{anteccedent} and a \textit{consequent}, such that if the
+  conditions in the antecedent hold, the consequent is assumed to be a valid
+  state or transition.
+\item [Environment] Sometimes we may implicitly rely on some additional
```
dnadales, line 139: I don't know if it would help to clarify that in principle it isn't necessary to have an environment: we could model it as part of the state. However by using an environment we're explicitly stating that **no rules** using the environment can change it.

In any case, we should mention that the environment remains constant.

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -0,0 +1,265 @@
...
+    {P(A) & z = f(s) & B = g(A,z)}
+    {\tion[E1]{S2}{A}{s}{B}}
+  \]
+To read such a rule, we proceed as follows:
+\begin{enumerate}
+  \item Firstly, we look at the LHS of the consequent. This introduces the basic
+    environment, iniital state and signal.
+  \item We then expand our set of variables using let bindings in the
+    antecedent.
+  \item We evaluate all predicates in the antecedent, which should now be
```
dnadales: Maybe "We evaluate all predicates in the antecedent ($P (A)$ in the example above),"

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -0,0 +1,265 @@
...
+    multiple predicate formulae. All predicates defined in the antecedent must be
+    true; that is, the antecedent is treated as the AND of all predicates defined
+    there.
+  \item The antecedent may contain variable definitions of the form $z=f(s)$,
+    where $z$ is an additional binding to be introduced. We will refer to these
+    as ``let-bindings'' in analogy with their use in programming languages.
+  \end{itemize}
+We can give an example of such a rule:
+  \[ \inference[Example]
+    {P(A) & z = f(s) & B = g(A,z)}
```
dnadales: So far we've been using spaces to separate function arguments. Should we make this part of our conventions regarding notation?

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -0,0 +1,265 @@
...
+    {\tion[E1]{S2}{A}{s}{B}}
+  \]
+To read such a rule, we proceed as follows:
+\begin{enumerate}
+  \item Firstly, we look at the LHS of the consequent. This introduces the basic
+    environment, iniital state and signal.
```
dnadales:
```suggestion
    environment, initial state and signal.
```

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -0,0 +1,265 @@
...
+
+  \textit{Rules}
+  \[ \inference[Rule-name]
+     {\textit{antecedent}}
+     {\textit{consequent}}
+     \in R
+  \]
+  \caption{Notation for a state transition system.}
+\end{figure}
+Figure \ref{fig:notation} gives the notation for transitions and rules. For
```
dnadales, line 196: I think we need to mention somewhere that free variables in the antecedent are considered to be existentially quantified.

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -0,0 +1,265 @@
...
+conditions and how they result in the construction of valid states.
+
+\section{Transition Systems}
+\label{sec:preliminaries}
+
+In describing the semantics for a system $L$ we are principally concerned with
+five things:
+
+\begin{description}
+\item [States] The underlying datatype of our system, whose validity we are
```
dnadales, line 129: Above we mention that we're checking the validity of inductive data structures, in particular ledgers and chains. So to me it seems that we're not proving the validity of states, but rather the states are means that we use to compute the validity conditions using small-step semantics.

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -0,0 +1,265 @@
...
+  required to prove.
+\item [Transitions] The means by which we might move from one (valid) state to
+  another.
+\item [Signals] The means by which transitions are triggered.
+\item [Rules] Formulae describing how to derive valid states or transitions. A
+  rule has an \textit{anteccedent} and a \textit{consequent}, such that if the
```
dnadales:
```suggestion
  rule has an \textit{anteccedent} (sometimes called \textit{premise}) and a \textit{consequent} (sometimes called \textit{conclusion}), such that if the
```

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -0,0 +1,265 @@
...
+system $S1$.
+
+\section{Serialisation}
+Various of our specifications will rely on a serialisation function which
+encodes a variety of types to a binary format.
+
+\begin{figure}[h]
+  \label{fig:serialisation}
+  \begin{align*}
+    & \llbracket\_\rrbracket \in \textbf{Hask}\mapsto\_\to\bytes
```
dnadales, line 258: I don't understand the type of this function. What does the underscore here represents?

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -0,0 +1,265 @@
...
+    & \text{serialisation} \\
+    & \llbracket\_\rrbracket_{A} \in \var{A}\to\bytes
+  \end{align*}
+  \caption{Serialisation functions}
+\end{figure}
+
+Figure \ref{fig:serialisation} gives the definition of a serialisation function.
+We define this as a partial function since it is defined only on a subset of
+Haskell types. As an abuse of notation, we drop the subscript $A$ for the
```
dnadales, line 267: Called "Hask" in the definition above?

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -0,0 +1,265 @@
...
+serialisation function on type $A$, since it is unambiguously determined by its
+first argument.
+
+We require our serialisation function to satisfy the following property:
+\begin{definition}[Distributivity over injective functions]
```
dnadales, line 272: I like this definition. Maybe we could include some motivation on why do we need this (I guess to avoid de-serialization in order to get a field of the serialized structure).

specs/semantics/latex/small-step-semantics.tex
dnadales: I see, I guess you're defining a family of serialization functions. However I wonder whether we could replace the underscore by a concrete type (which has to depend on Hask).

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -0,0 +1,265 @@
...
+  objects in $\mathcal{C}$, such that $s_A:A\to\var{D}$ for all $A\in\dom{s}$.
+  We say that $s$ \textit{distributes over injective functions} if, for all
+  injective functions $f:A\to B$ where $A,B\in\dom{s}$, there exists a function
+  $f_s:\var{D}\to\var{D}$ such that the following diagram commutes:
+  \[
+    \begin{tikzcd}
+      A \arrow{r}{f} \arrow{d}{s_A} & B \arrow{d}{s_B} \\
+      D \arrow{r}{f_s} & D
+    \end{tikzcd}
+  \]
```
dnadales, line 283: Maybe we should write down the algebraic identity that this diagram represents, i.e. something like:

```
[| f a |]_B = f_s [| a |]_A
```

dnadales (approved):
Looking good!

specs/semantics/latex/small-step-semantics.tex
nc6: Good point. Will do so!

specs/semantics/latex/small-step-semantics.tex
nc6: Somehow I hadn't noticed that! Perhaps we should. I'll roll this with the next update, though

specs/semantics/latex/small-step-semantics.tex
nc6: Hmm. Where/when does this occur? Do we actually need free variables in the antecedent? I'd actually prefer to avoid this if possible, and force explicit quantification if needed.

specs/semantics/latex/small-step-semantics.tex
nc6: I'm not sure I understand this comment :-) Will ask you about it on slack!

specs/semantics/latex/small-step-semantics.tex
nc6: The problem is that it's a dependent type. I don't know how best to represent this in standard mathematical notation!

specs/semantics/latex/small-step-semantics.tex
dnadales: I think we don't use this at the moment, however this was a possibility that was considered in @dcoutts [document about small-steps semantics](https://github.com/input-output-hk/iohk-functional-correctness/tree/master/dcoutts-nonsense). But we could see how far we get without an implicit quantification.

specs/semantics/latex/small-step-semantics.tex
nc6: Yep. I suppose I should technically write `Ob(Hask)`, but this is often elided...

specs/semantics/latex/small-step-semantics.tex
nc6: OK, I'll add this.

specs/semantics/latex/small-step-semantics.tex
dnadales: Alright then. Maybe with a little explanation people will understand this (even I did it :P )

specs/semantics/latex/small-step-semantics.tex
JaredCorduan: similarly, is it also worth mentioning that we usually write functions with curried notation?  functional programmers are used to this, but I don't think all audiences are.

specs/semantics/latex/small-step-semantics.tex
JaredCorduan: could you just drop the first line, and only keep `[_]_A -> Bytes`?  Saving `[_]` just for the abuse?  I'm happy either way.

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -0,0 +1,274 @@
...
+  \caption{Notation for a state transition system.}
+\end{figure}
+Figure \ref{fig:notation} gives the notation for transitions and rules. For
+notational convenience, we allow for some additional constructions in the
+antecedent:
+\begin{itemize}
+  \item Rather than a single logical predicate in the antecedent, we allow
+    multiple predicate formulae. All predicates defined in the antecedent must be
+    true; that is, the antecedent is treated as the AND of all predicates defined
```
JaredCorduan: would "as the conjunction of all predicates" sound better?

JaredCorduan (approved):
This looks great, @nc6! thanks!

specs/semantics/latex/small-step-semantics.tex
dnadales: Sure! My point is that we're not checking the validity of states, but we're using states to keep track of the information that we need for validating ledger and chains. But let's discuss this on slack (it is not a big deal anyway).

specs/semantics/latex/small-step-semantics.tex
dnadales: True, that would also help I think!

specs/semantics/latex/small-step-semantics.tex
nc6: My feeling is that I would rather prohibit it, unless there's good reason we need it. Allowing existential quantification there makes creating generators harder.

dnadales: True. I realized about this yesterday :)

PR #60, opened 2018-10-24
[#56] Use base-noprelude with cardano-prelude

ruhatch:
- Updated git commit in `cardano-chain`
  - Replaced `base` with `base-noprelude` in cabal files
  - Removed `NoImplicitPrelude` and added `-fno-warn-implicit-prelude`
  - Removed references to `Cardano.Prelude`
  - Fixed up a few places where `base`s `Prelude` was imported
  - Exported `id` from `Control.Category` as `identity`
  - Made `binary` and `crypto` packages rather than extra-deps so they're tested

Maybe some people think being explicit with the import is better, so let me know!

### DIFF
--- a/binary/cardano-binary.cabal
+++ b/binary/cardano-binary.cabal
@@ -29,7 +29,7 @@ library
                        Cardano.Binary.Class.Drop
                        Cardano.Binary.Class.Primitive
 
-  build-depends:       base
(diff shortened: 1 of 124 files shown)


ruhatch:
@parsonsmatt suggested this is too confusing, so we won't add this.

PR #63, opened 2018-10-24
Add flag to disable Werror

redxaxder:
This was lost in a rebase previously


ruhatch (approved):
I thought this was missing - good spot :+1:

PR #65, opened 2018-10-25
[CDEC-636] More operational blockchain rules with delegation interface incorporated

mdimjasevic:
This makes blockchain rules more operational. Furthermore, it incorporates two functions from the delegation interface. The interface is positioned between the ledger layer and the blockchain layer.


dnadales:
Do we want to use the conventions that @nc6 drafted out in [here](https://github.com/input-output-hk/cardano-chain/pull/58). One thing that I find really helpful is to have the types of the transitions explicitly defined.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -492,92 +639,144 @@ \subsubsection{Block Validity}
...
-       b = (h, sl, d, \sigma)
-    \end{array}}
-    &
-    {\begin{array}{c}
-       \sign{sk_d}{h}{sl}{d} = \sigma \\
-       \valid{b}{vk_d} \\
-       \delegate{sk_s}{vk_d} = (vk_s, vk_d, e, \sigma') \\
-       \signedwindow{sk_s}{[\predbseq^K\var{b}; \predbseq^{(K-1)}\var{b}; \cdots; \predbseq^1{b}]} | \leq K \cdot t
+      \hash{b} = h = p \\ % types have to be fixed for hashes
+      vk_s = \delegator{vk_d}{sl} \wedge (sk_d, vk_d) \in \SKey \times \VKey \\
```
dnadales: The `delegator` function should depend on the state of the transition system. This state could be updated by the ledger rules, but I think the blockchain rules need to keep it, and pass it to this `delegator` function.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -492,92 +639,144 @@ \subsubsection{Block Validity}
...
+      \size{\signmap{vk_s}} \leq K \cdot t \wedge \sign{sk_d}{h}{sl}{(\bdata{b})} = \bsig{b} \\
```
dnadales: Why do we need `sk_d` and `sl`? I know the OBFT paper specifies this, but I think we need to use the `verify` counterpart. The reason for this is that `sk_d` is existentially quantified, and it'd be difficult to implement, since all we have at a "block-transition" is a block, which only contains the verifying keys.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -492,92 +639,144 @@ \subsubsection{Block Validity}
...
+      \signmapname' = \incixmap{vk_s}{(\trimix{\signmapname}{ix})}{ix} \\
```
dnadales, line 636: I think we should specify what `incIxMap` and `trimIx` do, otherwise they could be anything, can't they?

Maybe redefine this in terms of the set and map operations.

dnadales:
One fundamental piece of information that I'm missing is whether you could use the ledger rule I proposed:

image

I really need to know this, since otherwise I cannot move forward with the delegation rules. 

If you want we can discuss this over a video call.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -492,92 +639,144 @@ \subsubsection{Block Validity}
...
+  %
+  \begin{align*}
+    \text{qHead} & \in \Queue_\BlockIx \totalf \BlockIx^? & \text{head of queue function} \\
+    \text{qPush} & \in \Queue_\BlockIx \times \BlockIx \totalf \Queue_\BlockIx
+      & \text{queue push function} \\
+    \text{qPop} & \in \Queue_\BlockIx \totalf {\Queue_\BlockIx}^?
+      & \text{queue pop function} \\
+    \text{\trimixname} & \in \mapqueue \times \BlockIx \totalf \mapqueue
+      & \text{old block removal function} \\
+    \text{\incixmapname} & \in \VKeyGen \times \mapqueue \times \BlockIx \totalf \mapqueue
```
dnadales: The functions have different fonts here and in the rules.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -492,92 +639,144 @@ \subsubsection{Block Validity}
...
+  %
+  \emph{Type aliases}
+  %
+  \begin{align*}
+    \mapqueue = \VKeyGen \totalf \Queue_\BlockIx
+  \end{align*}
+  \emph{Functions}
+  %
+  \begin{align*}
+    \text{qHead} & \in \Queue_\BlockIx \totalf \BlockIx^? & \text{head of queue function} \\
```
dnadales: It seems `qHead`, `qPush`, `qPop` aren't used, are they?

specs/chain/latex/blockchain-spec.tex
```diff
@@ -492,92 +639,144 @@ \subsubsection{Block Validity}
...
-  \label{eq:predbseq}
-  \predbseq{b} = (\text{\toseqname} \circ \text{\predbname})(b)
-\end{equation}
 
-Analogously to the \predbname\ function, $\predbseq^m{b}$ is defined as:
+\begin{figure}
+  \emph{Abstract types}
+  %
+  \begin{align*}
+    q & \in \Queue_\BlockIx  & \text{block index queue}
```
dnadales: It seems `\Queue_\BlockIx` isn't used.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -157,98 +178,256 @@ \section{Preliminaries}
...
-possibly the second item in the list have to be examined.
+If $s(\cdot)$ is a function for serialising data and $d(\cdot)$ a function for
+deserialising data, their composition $s \circ d$ is not an identity function:
+%
+this is hard to enforce and it is considered a bad practice in cryptography
+design.
 
-This approach also supports certificate revocation: it is done by posting a
-certificate in which a key delegates to itself.
+% \section{Delegation}
```
dnadales: Can we remove the commented code?

specs/chain/latex/blockchain-spec.tex
```diff
@@ -157,98 +178,256 @@ \section{Preliminaries}
...
-Due to trimming, each list in the map has at most two certificates at any
-point in time.
-%
-At the beginning of an epoch, a list has only one element and at any other
-point it can have at most two elements (i.e., two certificates) if there was a
-new delegation certificate for the next epoch posted.
 
+\section{Auxiliary definitions}
+\label{sec:auxil-defin}
+TODO: Put this note on serialisation somewhere more appropriate.
```
dnadales: @nc6 already took care of this in [this branch](https://github.com/input-output-hk/cardano-chain/tree/nc/semantics-spec). Maybe we could update this, and also discuss with him whether we need to do anything in the SOS rules.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -267,70 +446,59 @@ \section{Delegation Interface}
...
 for checking if a particular verification key has the delegation right to sign
-a block in the current slot.
+a block in a given slot.
 %
 In particular, if the verification key has the right to sign, the
 functionality should inform who is the delegator.
 %
 In Haskell, the signature of the corresponding function is:
 %
 \begin{lstlisting}
```
dnadales: This is not render correctly, I think you need to configure the `listings` style. We could make an issue to have a common file with this, so that we can reuse it across documents, like the `iohk.sty` that @JaredCorduan made.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -157,98 +178,256 @@ \section{Preliminaries}
...
+  \emph{Abstract types}
+  %
+  \begin{align*}
+    c & \in \HCert  & \text{heavyweight delegation certificate} \\
+  \end{align*}
+  %
+  \emph{Functions}
+  %
+  \begin{align*}
+    \text{cFrom} & \in \HCert \totalf \VKeyGen & \text{certificate delegator key projection} \\
```
dnadales: Maybe we don't need to mention "projection" every time?

specs/chain/latex/blockchain-spec.tex
```diff
@@ -157,98 +178,256 @@ \section{Preliminaries}
...
+  %
+  \begin{align*}
+    ix & \in \BlockIx & \text{block index}\\
+     h & \in \BHash   & \text{block hash}\\
+  \end{align*}
+  %
+  \emph{Derived types}
+  %
+  \begin{align*}
+    (vk_1, \dotsc, vk_n, h) & \in \GBlock = \VKeyGen^n \times \BHash & \text{genesis block} \\
```
dnadales: Why not using "projections" for the genesis block as well?

specs/chain/latex/blockchain-spec.tex
dnadales, line 636: I think `ix = bIx`, right? And probably you also need to define `sl`.

dnadales (commented):
My main concerns are:

0. The `delegator` function should be a function of the state of the transition system.
1. We need not use `sk_d` in the rules.
2. Functions `incIxMap` and `trimIx` should be explicitly defined.

We can discuss this over a video call together with @nc6, @JaredCorduan , and @dcoutts .

mdimjasevic:
> Do we want to use the conventions that @nc6 drafted out in [here](https://github.com/input-output-hk/cardano-chain/pull/58). One thing that I find really helpful is to have the types of the transitions explicitly defined.

I'll definitely use whatever conventions and notation we use. I'll read what Nick has and update this spec accordingly.

mdimjasevic:
> One fundamental piece of information that I'm missing is whether you could use the ledger rule I proposed:
> 
> image
> 

The type of `pdlgs` seems reasonable.

A type dictates a semantics, but the specifics of a value implementing the type (e.g., inference rules) should be irrelevant for the interface. If it wasn't that way, then there would be no interface and no separation of concerns. Instead, there would be a dependent implementation of different things, in our example a combined implementation of the blockchain and the ledger layers.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: You're right. An alternative is to use the function symbol without any arguments applied, which puts unwelcome white space after the symbol. If you have a different proposal to handle this, I'd be more than happy to implement it!

specs/chain/latex/blockchain-spec.tex
mdimjasevic: I'll provide definitions for `trimIx` and `incIxMap` next, which will use these queue operations.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: Right, I'll remove it.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: I noticed Haskell code doesn't render properly and that I'll have to fix it. Is it already customised properly in Jared's style file?

specs/chain/latex/blockchain-spec.tex
mdimjasevic: That's a good point. I'll change that.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: Right. Fixed. Thanks!

mdimjasevic:
@dnadales , I don't see how could the `delegator` function depend on a state of the blockchain transition system. Can you please explain how?

As described in the document, this function is part of the interface between the ledger layer and the blockchain layer and it is provided by the ledger layer. Per our discussions and an agreement, the blockchain layer doesn't keep certificates as part of its state, hence the `delegator` function cannot depend on a non-existing part of the blockchain state. That part of the state is maintained by the ledger. Don't you agree?

specs/chain/latex/blockchain-spec.tex
JaredCorduan: not yet, but that would be a great thing to add to it

dnadales:
> > One fundamental piece of information that I'm missing is whether you could use the ledger rule I proposed:
> > image
> 
> The type of `pdlgs` seems reasonable.
> 
> A type dictates a semantics, but the specifics of a value implementing the type (e.g., inference rules) should be irrelevant for the interface. If it wasn't that way, then there would be no interface and no separation of concerns. Instead, there would be a dependent implementation of different things, in our example a combined implementation of the blockchain and the ledger layers.

We should aim for all the best things formal methods and good software engineering practices advocate, however note that we're using the small-steps semantics framework, which means that we have to play according to its rules. Here we have no concepts of type-classes, modules, dependent types, or what have you. If there are improvements to the small-steps semantics framework you think we can introduce within a reasonable time frame we should by any means do it, but we should keep in mind that:

0. Operational rules should be simple: the idea behind the operational rules is that we can communicate with other researchers (who might not speak FP or Haskell).
1. We have to deliver something sooner rather than later.

All in all, the chain rules have to refer to the ledger transitions at some point, and so far I have no idea how do you plan on doing this. Without knowing this I don't see which kind of transitions I have to provide. What I'd expect to see is something like:

```
... some conditions ... ledgerEnv |- ledgerSt --L--> ledgerSt'
... maybe some conditions on ledgerEnv/ledgerSt/LedgerSt'
--------------------------------------------------------------------
chainEnv |- chainSt --C--> chainSt'
```

Where `--C-->` is the chain transition, and `--L-->` is the ledger transition.

dnadales:
> @dnadales , I don't see how could the `delegator` function depend on a state of the blockchain transition system. Can you please explain how?
> 
> As described in the document, this function is part of the interface between the ledger layer and the blockchain layer and it is provided by the ledger layer. Per our discussions and an agreement, the blockchain layer doesn't keep certificates as part of its state, hence the `delegator` function cannot depend on a non-existing part of the blockchain state. That part of the state is maintained by the ledger. Don't you agree?

Let's discuss this over video call. Otherwise we'll never agree on anything like this.

dnadales (approved):
I think discussing the open issues on github is neither efficient nor effective. Let's have a video call with all the parties that might be interested (besides @mdimjasevic and @dnadales of course), open a google doc, and try to come to an agreement.

nc6:
Concur, let's aim to have an agreement today!

Issue #45, opened 2018-10-19
Track location of original bytestring


redxaxder:
Tag deserialized types with an annotation that allows the original bytes to be recovered.

PR #70, opened 2018-10-25
[#45] Track offsets into bytestring when decoding

redxaxder:


ruhatch:
Changing direction on this one, so this PR is now unneeded.

PR #77, opened 2018-10-25
[#38] Change golden_ to camel case in tests

### LINKED ISSUE #38: Change 'golden_' to camel case in golden tests
We use `roundTrip` as a prefix for round trip tests, so we should be consistent. This also satisfies some hlint suggestions, so it should be good.

ruhatch:
Update golden tests to use camel case instead of underscores in the prefix.

Closes #38

### DIFF
--- a/binary/test/Test/Cardano/Binary/Helpers/GoldenRoundTrip.hs
+++ b/binary/test/Test/Cardano/Binary/Helpers/GoldenRoundTrip.hs
@@ -7,7 +7,7 @@ module Test.Cardano.Binary.Helpers.GoldenRoundTrip
        , roundTripsBiShow
        , roundTripsBiBuildable
        , compareHexDump
-       , legacyGoldenDecode
+       , deprecatedGoldenDecode
        ) where
 
 import           Cardano.Prelude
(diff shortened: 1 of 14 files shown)


binary/test/Test/Cardano/Binary/Helpers/GoldenRoundTrip.hs
```diff
@@ -7,7 +7,7 @@ module Test.Cardano.Binary.Helpers.GoldenRoundTrip
        , roundTripsBiShow
        , roundTripsBiBuildable
        , compareHexDump
-       , legacyGoldenDecode
+       , deprecatedGoldenDecode
```
redxaxder, line 10: Both "legacy" and "deprecated" have negative connotations. Maybe `byronGoldenDecode`?

redxaxder (approved):

binary/test/Test/Cardano/Binary/Helpers/GoldenRoundTrip.hs
ruhatch: Damian suggested this too, but we might use this for other things that we deprecate. I'm open to other suggestions, but I think deprecating is pretty normal (and therefore not too negative) term.

PR #85, opened 2018-10-31
State transition systems

nc6:
Still WIP, I need to update witnessing and add delegation. But this should give a flavour.


dnadales:
Would it be possible to include information in the README about how to:

  0. Test the code
  1. Fire up a REPL

I guess we'd want to use Nix for this...

specs/semantics/hs/src/Control/State/Transition.hs
```diff
@@ -0,0 +1,103 @@
...
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE GADTs                 #-}
+{-# LANGUAGE KindSignatures        #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE TypeFamilies          #-}
+
+-- | Small step state transition systems.
+module Control.State.Transition where
+
+import           Control.Lens
```
dnadales, line 12: Could we use implicit imports? They help me a lot when tracing where a function comes from.

specs/semantics/hs/src/Control/State/Transition.hs
```diff
@@ -0,0 +1,106 @@
...
+data PredicateResult sts
+  = Passed
+  | Failed (PredicateFailure sts)
+
+-- | The antecedent to a transition rule.
+data Antecedent sts where
+
+  -- | This rule is predicated upon a transition under a (sub-)system.
+  SubTrans
+    :: Embed sub sts
```
dnadales, line 76: Since `sub` and `sts` are both `STS`'s, maybe for the sake of consistency we might want to call them `subSts` and `sts` (or `superSts`, but that doesn't look nice).

dnadales (approved):
Looks really nice!

JaredCorduan (approved):
looks great, thanks @nc6 !

specs/ledger/hs/src/witnessed/ledger/Ledger.hs
```diff
@@ -1,59 +1,47 @@
...
-    predNoIncreasedBalance = Predicate
-      { _pName = "predValidInputs"
-      , _pValidate = \st sig -> case sig of
-          NoSignal -> Passed
-          TxSignal (TxWits tx _) -> liftResult $ Ledger.Simple.noIncreasedBalance tx (getUtxo st)
-      }
+  rules =
+    [ Rule [] $ Base (UTxO Map.empty)
+    , Rule
+      [ SubTrans _1 (_3 . to body) utxoInductive
```
dnadales, line 35: I missed this aspect: it seems that we can reuse the predicates, but we have to do some copying to be able to reuse the state changes caused by the premise :( I only realized about this when trying to write my own instances.

I wonder whether we could solve this problem (while keeping the elegance of the framework you defined).

PR #86, opened 2018-10-31
Update ledger delegation rules according to blockchain-ledger interface

### LINKED ISSUE #66: Update specification of epochs in delegation certificates

dnadales:
- Remove most of the prose as that is difficult to maintain when the rules are not stable. 
- Add additional notation and explanations.
- Update the delegation rules.
- Add the blockchain-ledger interface.
- Change `input` by `include` so that the `includeonly` command can be used when working on a draft.

Closes #66.


specs/ledger/latex/delegation.tex
```diff
@@ -1,60 +1,157 @@
...
-Variables to the left of the $\vdash$ (turnstile) symbol are the
-\textit{environment} of a rule, which remains constant. This environment (the
-current epoch in this case) would be set by the (chain) rules that use
-Rule~\ref{eq:delegation}.
+There are several restrictions on a certificate posted on the blockchain:
+\begin{enumerate}
+\item Only genesis keys can delegate.
+\item Certificates must be properly signed by the delegator.
+\item Any given key can delegate at most once per-epoch.
+\item The epochs in certificate cannot refer to past epochs. This mechanism
```
nc6:
```suggestion
\item The epochs in certificates cannot refer to past epochs. This mechanism
```

specs/ledger/latex/delegation.tex
```diff
@@ -1,60 +1,157 @@
...
+  prevents replay attacks.
+\item Two certificates for the same key issued in the same (current) slot
+  cannot contain different epochs.
```
nc6, line 23: I don't understand this point. How can one key post two certificates in the same slot?

specs/ledger/latex/delegation.tex
```diff
@@ -1,60 +1,157 @@
...
-  \caption{Delegation definitions}
-  \label{fig:defs:delegation}
+  \begin{equation*}
+    \begin{array}{r@{~\in~}lr}
+      \fun{dbody} & \DCert \to (\VKey \times \Epoch)
+      & \text{body of the delegation certificate}\\
+      \fun{dwit} & \DCert \to (\VKeyGen \times \Sig)
+      & \text{witness for the delegation certificate}\\
+      \fun{dwho} & \DCert \mapsto (\VKeyGen \times \VKey)
+      & \text{who delegates to who in the certificate}\\
```
nc6:
```suggestion
      & \text{who delegates to whom in the certificate}\\
```

specs/ledger/latex/delegation.tex
```diff
@@ -1,60 +1,157 @@
...
+\begin{figure}[htb]
+  \emph{Delegation scheduling environments}
+  \begin{equation*}
+    \DSEnv =
+    \left(
+      \begin{array}{r@{~\in~}lr}
+        \mathcal{K} & \powerset{\VKeyGen} & \text{allowed delegators}\\
+        \var{e} & \Epoch & \text{epoch}\\
+        \var{s} & \Slot & \text{slot}\\
+        \var{d} & \Slot & \text{certificate liveness parameter}
```
nc6: I wonder if it would be better to have a separate type for `SlotCount`. This would avoid problems where we get confused between absolute slots and relative slots.

specs/ledger/latex/delegation.tex
```diff
@@ -1,60 +1,157 @@
...
+      \end{array}
+    \right)
+  \end{equation*}
+
+  \emph{Delegation scheduling states}
+  \begin{equation*}
+    \DSState
+    = \left(
+      \begin{array}{r@{~\in~}lr}
+        \var{sds} & \seqof{(\Slot \times (\VKeyGen \times \VKey))} & \text{scheduled delegations}\\
```
nc6, line 93: Particular reason for the nested tupling here?

specs/ledger/latex/delegation.tex
```diff
@@ -1,60 +1,157 @@
...
+  \caption{Delegation scheduling transition-system types}
+  \label{fig:ts-types:delegation-scheduling}  
+\end{figure}
+
+\begin{figure}[htb]
+  \begin{equation}
+    \label{eq:rule:delegation-scheduling}
+    \inference
+    {
+      \dwit{c} = (\var{vk_s},~ \sigma)
```
nc6: I somehow feel these would be easier to read the other way around, making it clear that the binding is happening to the variables on the _left_. But this is just nitpicking

specs/ledger/latex/delegation.tex
```diff
@@ -64,44 +161,63 @@
...
         \begin{array}{r@{~\in~}lr}
-          \var{dmap} & \VKeyGen \mapsto \VKey & \text{delegation map}\\
-          \var{ekeys} & \Epoch \mapsto \powerset{\VKeyGen} & \text{keys delegated per epoch}
+          \var{dms} & \VKeyGen \mapsto \VKey & \text{delegation map}\\
+          \var{dws} & \VKeyGen \mapsto \Slot & \text{when last delegation occurred}\\
         \end{array}\right)
   \end{align*}
   \emph{Delegation transitions}
   \begin{equation*}
     \_ \vdash \_ \trans{deleg}{\_} \_ \in
```
nc6: Should this be under "adeleg" now?

specs/ledger/latex/blockchain-interface.tex
```diff
@@ -0,0 +1,152 @@
...
+        \left(
+          \begin{array}{l}
+            \var{dms}\\
+            \var{dws}\\
+            \var{sds}\\
+            \var{eks}
+          \end{array}
+        \right)
+      }
+      \trans{sdeleg}{\Gamma}
```
nc6:
```suggestion
      \trans{deleg}{\Gamma}
```

nc6 (approved):
Minor comments and a few things which need to be fixed, but in general I think this is looking great. I figure we will finagle out any final details as we write the Haskell equivalent!

specs/ledger/latex/blockchain-interface.tex
```diff
@@ -0,0 +1,152 @@
...
+        \var{sds} & \seqof{(\Slot \times (\VKeyGen \times \VKey))} & \text{scheduled delegations}\\
+        \var{eks} & \powerset{(\Epoch \times \VKeyGen)} & \text{key-epoch delegations}
+      \end{array}
+    \right)
+  \end{equation*}
+
+  \emph{Delegation transitions}
+  \begin{equation*}
+    \_ \vdash \_ \trans{deleg}{\_} \_ \in
+    \powerset (\DIEnv \times \DIState \times \seqof{\DCert} \times \DIState)
```
mdimjasevic, line 40: Does it matter that certificates come in sequence, i.e., in a particular order? If not, I'd suggest to make this a set instead.

specs/ledger/latex/delegation.tex
mdimjasevic: I got confused by this as well.

mdimjasevic:
I'd recommend using more different names for `DState` and `DSState`. It gets hard to keep track which one is which.

dnadales:
Sure. Which names do you recommend?


On Nov 2, 2018 11:11, "Marko Dimjašević" <notifications@github.com> wrote:

I'd recommend using more different names for DState and DSState. It gets
hard to keep track which one is which.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/input-output-hk/cardano-chain/pull/86#issuecomment-435331706>,
or mute the thread

specs/ledger/latex/delegation.tex
```diff
@@ -1,60 +1,157 @@
...
+
+\begin{figure}[htb]
+  \begin{equation}
+    \label{eq:rule:delegation-scheduling}
+    \inference
+    {
+      \dwit{c} = (\var{vk_s},~ \sigma)
+      & \verify{vk_s}{\serialised{\dbody{c}}}{\sigma} & vk_s \in \mathcal{K}\\ ~ \\
+      \dwho{c} = (\var{vk_s},~ \var{vk_d}) & \depoch{c} = e_d
+      & (e_d,~ \var{vk_s}) \notin \var{eks} & e < e_d \\ ~ \\
```
mdimjasevic, line 117: What epoch does `e` represent? Is it the current epoch?

mdimjasevic:
I'd recommend adding text and referencing all figures you have in the document. That will make it easier to see how they all relate to each other.

specs/ledger/latex/delegation.tex
```diff
@@ -1,60 +1,157 @@
...
+      (s + d,~ (\var{vk_s},~ \wcard)) \notin \var{sds}\\
```
mdimjasevic, line 118: Instead of calculating when the certificate becomes active yourself, I'd recommend extracting that info from the certificate and, of course, checking it before adding to `sds`.

specs/ledger/latex/delegation.tex
```diff
@@ -64,44 +161,63 @@
...
+        \var{dms} & \unionoverride & \{\var{vk_s} \mapsto \var{vk_d}\}\\
+        \var{dws} & \unionoverride & \{\var{vk_s} \mapsto s \}
+      \end{array}
+      \right)
+    }
+  \end{equation}
+  \nextdef
+  \begin{equation}\label{eq:rule:delegation-nop}
+    \inference
+    {\var{kv_s} \mapsto s_p  \in \var{dws}  \wedge s \leq s_p
```
mdimjasevic: Did you mean `vk_s` instead of `kv_s` here?

mdimjasevic:
Furthermore, I just noticed there is `DIState` as well. When it comes to environments, there is a similar confusion. Having names that are easy to distinguish in an as dense document as this one would definitely be of help.

specs/ledger/latex/blockchain-interface.tex
```diff
@@ -0,0 +1,152 @@
...
+
+\subsubsection{Delegation interface functions}
+\label{sec:delegation-interface-functions}
+
+\begin{figure}[htb]
+  \begin{align*}
+    & \fun{delegates} \in \DIState \to (\VKeyGen \times \VKey) \to \Bool & \text{delegation relationship}\\
+    & \fun{delegates}~s~(\var{vk_s}, \var{vk_d}) = \var{vk_s} \mapsto \var{vk_d} \in (\var{dms}~s)\\
+    \nextdef
+    & \fun{initialDS} \in \VKeyGen \to \DIState & \text{initial delegation state}\\
```
mdimjasevic, line 140: How does this initial delegation state fit into existing rules? My guess is that the state shouldn't be a function of a verification key, but that it should be initialised for all verification keys from the genesis block, which would change the state's type to `DIState`.

mdimjasevic (commented):
I am not sure what are all states and environments that you use. They have almost the same names, and some of them have the same types, which makes it tricky to review. I'd be happy to have another run if you make changes that address these issues.

specs/ledger/latex/blockchain-interface.tex
nc6: The issue is that they're _consumed_ in a sequence, so this makes sense, otherwise we have to pick an item from the set in these rules.

dnadales: Another reason for choosing sequences is that they have an inductive definition, i.e. you can pattern match in the rules using sequences (`\epsilon`, `xs; x`), but there's no inductive definition of a set (I think `{x} \cup X`  will give you `|X| + 1` possible matches).

specs/ledger/latex/delegation.tex
dnadales: In principle there's nothing that prevents a client of the ledger layer from submitting two delegation certificates with the same delegator (genesis key), but having different epochs. In this case both certificates would become active at `s + d` (`s` is the current slot). We don't perform any checking on the relationship between epochs and slots. Hence the constraint.

specs/ledger/latex/delegation.tex
dnadales: That makes sense. I'll introduce a `SlotCount = N` type.

specs/ledger/latex/delegation.tex
dnadales: Just to:

- Read it as a relation between two types.
- Pattern match on the second element as `(s, w)`.

specs/ledger/latex/delegation.tex
dnadales: Done!

specs/ledger/latex/delegation.tex
dnadales: Yes! Thanks.

specs/ledger/latex/delegation.tex
dnadales: Yes.

specs/ledger/latex/delegation.tex
dnadales: But as far as I understand we don't have the information about when a certificate becomes active in the certificate itself. Or am I wrong?

specs/ledger/latex/delegation.tex
dnadales: Yes! Nice catch. Thanks.

specs/ledger/latex/blockchain-interface.tex
dnadales: The blockchain layer could pass the keys of the genesis block to this function, initializing the initial delegation map with the identity delegations. There might be another options. Please feel free to submit a PR if you can think of them. Thanks!

dnadales:
Thank you @mdimjasevic and @nc6 for your reviews. I'll try to address all the comments I could. There are two pending remarks by Marko:

- The `VKeyGen` parameter in the `initialDS` function.
- The names of the states and environments.

Since I don't see a clear way to address these remarks @mdimjasevic feel free to either open separate issues or submit a PR against the ledger rules. In particular, I understand that we have very similar names, but I wanted to avoid long ones in the spec. But do feel free to propose better ones!

Thanks again.

mdimjasevic:
@dnadales , can you tell me which of your `D*State`s is `DelegState` from our last week's discussion and that is part of the interface than the blockchain layer interacts with? I need this information to update the blockchain specification accordingly.

dnadales:
> @dnadales , can you tell me which of your D*States is DelegState from our last week's discussion and that is part of the interface than the blockchain layer interacts with? I need this information to update the blockchain specification accordingly.

Delegation-interface state: `DIState`. Feel free to submit a PR renaming it to whatever you think is appropriate. Also feel free to bring the ledger and chain types in sync. If you do this, take a look at the types that @nc6 already implemented in Haskell.

PR #92, opened 2018-11-01
[DEVOPS-1118] get cs-ledger building with nix

cleverca22:


dnadales (approved):
Thanks for this!

Maybe we need to update the README on how to do `build`, `test`, and `repl` with Nix. Do we use `nix-shell --run "cabal build"` for instance?

cleverca22:
there is currently no `cabal` binary, you need to either `runhaskell Setup.hs` or `ghc Setup.hs -o Setup` and then use `./Setup` as the cabal binary

dnadales:
Thank you @cleverca22

PR #94, opened 2018-11-01
Candidate design for tracking original bytes in deserialized data

redxaxder:
The goal of this changeset is to (eventually) stop deserializing-then-reserializing during validation of incoming data.
This PR only does this for the header data and not the entire block.

For review only. Do not merge.

### DIFF
--- a/binary/src/Cardano/Binary/Class/Annotated.hs
+++ b/binary/src/Cardano/Binary/Class/Annotated.hs
@@ -1,0 +1,36 @@
+{-# LANGUAGE DeriveAnyClass     #-}
+{-# LANGUAGE DeriveFunctor      #-}
+{-# LANGUAGE DeriveGeneric      #-}
+{-# LANGUAGE DerivingStrategies #-}
+
+module Cardano.Binary.Class.Annotated
+  ( Annotated (..)
+  , ByteSpan (..)
+  , slice
+  , decodeAnnotated
+  )
+  where
+
+import qualified Codec.CBOR.Decoding as D
+import           Codec.CBOR.Read (ByteOffset)
+import qualified Data.ByteString.Lazy as BSL
+
+import           Cardano.Binary.Class.Core (Bi (..))
+import           Cardano.Prelude
+
+
+slice :: BSL.ByteString -> ByteSpan -> BSL.ByteString
+slice bytes (ByteSpan start end) = BSL.take (end - start) $ BSL.drop start $ bytes
+
+data ByteSpan = ByteSpan !ByteOffset !ByteOffset
+
+data Annotated b a = Annotated { unAnnotated :: !b, annotation :: !a }
+  deriving (Eq, Show, Functor, Generic)
+  deriving anyclass NFData
+
+annotatedDecoder :: D.Decoder s a -> D.Decoder s (Annotated a ByteSpan)
+annotatedDecoder d = flip fmap (D.decodeWithByteSpan d) $ \(x, start, end) ->
+  Annotated x (ByteSpan start end)
+
+decodeAnnotated :: (Bi a) => D.Decoder s (Annotated a ByteSpan)
+decodeAnnotated = annotatedDecoder decode
--- a/binary/src/Cardano/Binary/Class/Annotated.hs
+++ b/binary/src/Cardano/Binary/Class/Annotated.hs
@@ -31,7 +31,7 @@ data Annotated b a = Annotated { unAnnotated :: !b, annotation :: !a }
   deriving anyclass NFData
 
 annotatedDecoder :: D.Decoder s a -> D.Decoder s (Annotated a ByteSpan)
-annotatedDecoder d = flip fmap (D.decodeWithByteSpan d) $ \(x, start, end) ->
+annotatedDecoder d = D.decodeWithByteSpan d <&> \(x, start, end) ->
   Annotated x (ByteSpan start end)
 
 decodeAnnotated :: (Bi a) => D.Decoder s (Annotated a ByteSpan)
--- a/binary/src/Cardano/Binary/Class/Annotated.hs
+++ b/binary/src/Cardano/Binary/Class/Annotated.hs
@@ -2,12 +2,14 @@
 {-# LANGUAGE DeriveFunctor      #-}
 {-# LANGUAGE DeriveGeneric      #-}
 {-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE Rank2Types #-}
 
 module Cardano.Binary.Class.Annotated
   ( Annotated (..)
   , ByteSpan (..)
   , slice
   , decodeAnnotated
+  , decodeFullAnnotatedBytes
   )
   where
 
@@ -15,9 +17,9 @@ import qualified Codec.CBOR.Decoding as D
 import           Codec.CBOR.Read (ByteOffset)
 import qualified Data.ByteString.Lazy as BSL
 
-import           Cardano.Binary.Class.Core (Bi (..))
+import           Cardano.Binary.Class.Core (Bi (..), DecoderError)
+import           Cardano.Binary.Class.Primitive (decodeFullDecoder)
 import           Cardano.Prelude
-
 
 slice :: BSL.ByteString -> ByteSpan -> BSL.ByteString
 slice bytes (ByteSpan start end) = BSL.take (end - start) $ BSL.drop start $ bytes
@@ -34,3 +36,12 @@ annotatedDecoder d = flip fmap (D.decodeWithByteSpan d) $ \(x, start, end) ->
 
 decodeAnnotated :: (Bi a) => D.Decoder s (Annotated a ByteSpan)
 decodeAnnotated = annotatedDecoder decode
+
+decodeFullAnnotatedBytes
+  :: (Functor f)
+  => Text
+  -> (forall s. D.Decoder s (f ByteSpan))
+  -> BSL.ByteString
+  -> Either DecoderError (f ByteString)
+decodeFullAnnotatedBytes lbl decoder bytes = (fmap.fmap) (BSL.toStrict . slice bytes) $
+  decodeFullDecoder lbl decoder bytes
--- a/crypto/src/Cardano/Crypto/Signing/Check.hs
+++ b/crypto/src/Cardano/Crypto/Signing/Check.hs
@@ -10,6 +10,7 @@
 
 module Cardano.Crypto.Signing.Check
        ( checkSig
+       , checkSigAnnotated
        , checkSigRaw
        , verifyProxyCert
        , validateProxySecretKey
@@ -22,7 +23,7 @@ import qualified Cardano.Crypto.Wallet as CC
 import           Control.Monad.Except (MonadError, throwError)
 import           Data.Coerce (coerce)
 
-import           Cardano.Binary.Class (Bi, Raw)
+import           Cardano.Binary.Class (Annotated (..), Bi, Raw)
 import qualified Cardano.Binary.Class as Bi
 import           Cardano.Crypto.ProtocolMagic (ProtocolMagic)
 import           Cardano.Crypto.Signing.Tag (SignTag (..), signTag)
@@ -37,6 +38,10 @@ import           Cardano.Crypto.Signing.Types.Signing (ProxyCert (..),
 checkSig
   :: Bi a => ProtocolMagic -> SignTag -> PublicKey -> a -> Signature a -> Bool
 checkSig pm t k x s = checkSigRaw pm (Just t) k (Bi.serialize' x) (coerce s)
+
+checkSigAnnotated
+  :: ProtocolMagic -> SignTag -> PublicKey -> (Annotated a ByteString) -> Signature a -> Bool
+checkSigAnnotated pm t k x s = checkSigRaw pm (Just t) k (annotation x) (coerce s)
 
 -- CHECK: @checkSigRaw
 -- | Verify raw 'ByteString'
--- a/crypto/src/Cardano/Crypto/Signing/Signing.hs
+++ b/crypto/src/Cardano/Crypto/Signing/Signing.hs
@@ -24,6 +24,7 @@ module Cardano.Crypto.Signing.Signing
        , validateProxySecretKey        -- reexport
        , proxySign
        , proxyVerify
+       , proxyVerifyAnnotated
 
        , module Cardano.Crypto.Signing.Types.Signing
        ) where
@@ -33,11 +34,10 @@ import           Cardano.Prelude
 import qualified Cardano.Crypto.Wallet as CC
 import           Crypto.Random (MonadRandom, getRandomBytes)
 import           Data.ByteArray (ScrubbedBytes)
-import qualified Data.ByteString as BS
 import           Data.Coerce (coerce)
 import           Formatting (build, sformat)
 
-import           Cardano.Binary.Class (Bi, Raw)
+import           Cardano.Binary.Class (Bi, Raw, Annotated(..))
 import qualified Cardano.Binary.Class as Bi
 import           Cardano.Crypto.ProtocolMagic (ProtocolMagic)
 import           Cardano.Crypto.Signing.Check (checkSig, checkSigRaw,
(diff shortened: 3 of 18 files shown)


binary/src/Cardano/Binary/Class/Annotated.hs
```diff
@@ -0,0 +1,47 @@
...
+slice bytes (ByteSpan start end) = BSL.take (end - start) $ BSL.drop start $ bytes
+
+data ByteSpan = ByteSpan !ByteOffset !ByteOffset
+
+data Annotated b a = Annotated { unAnnotated :: !b, annotation :: !a }
+  deriving (Eq, Show, Functor, Generic)
+  deriving anyclass NFData
+
+annotatedDecoder :: D.Decoder s a -> D.Decoder s (Annotated a ByteSpan)
+annotatedDecoder d = flip fmap (D.decodeWithByteSpan d) $ \(x, start, end) ->
```
ruhatch: `D.decodeWithByteSpan <&> \(x, start end) -> ...`?

src/Cardano/Chain/Block/Header.hs
```diff
@@ -66,26 +77,45 @@ import           Cardano.Chain.Slotting (SlotId (..), slotIdF)
...
 -- Header
 --------------------------------------------------------------------------------
 
-data Header = Header
-  { headerProtocolMagic :: !ProtocolMagic
-  , headerPrevHash      :: !HeaderHash
+type Header = AHeader ()
+
+header :: ProtocolMagic -> HeaderHash -> Proof -> ConsensusData -> ExtraHeaderData -> Header
+header pm hh p cd ehd = AHeader
```
ruhatch: I'm not sure I like this being called `header`, but I know `mkHeader` is also taken. We probably shouldn't have been exposing the `Header` constructor before and this opens that up too. You could simply inline this to it's only usage site and then you can change back `h`s to `header`s.

src/Cardano/Chain/Block/Header.hs
```diff
@@ -345,11 +374,31 @@ instance Bi BlockSignature where
...
+      , (annotation . aHeaderPrevHash) h
+      , (annotation . aHeaderProof) h
+      , (annotation . aConsensusSlot . headerConsensusData) h
+      , (annotation . aConsensusDifficulty . headerConsensusData) h
+      , (annotation . aHeaderExtraData) h
+      ]
+    toSign = ToSign
+      (headerPrevHash h)
+      (headerProof h)
+      (consensusSlot . headerConsensusData $ h)
```
ruhatch: I think there's a `headerSlot` and `headerDifficulty` to shorten these

test/Test/Cardano/Chain/Block/Header.hs
```diff
@@ -0,0 +1,43 @@
...
+
+data Error = ErrorH HeaderError | ErrorD DecoderError
+  deriving (Show, Eq)
+
+validateAnnotatedHeader :: Property
+validateAnnotatedHeader = H.property $ do
+  pm <- H.forAll genProtocolMagic
+  sc <- H.forAll genSlotCount
+  header <- H.forAll $ genHeader pm sc
+  let bytes = (serializeEncoding . encode) header
```
ruhatch: Couldn't you just use `serialize` here?

test/Test/Cardano/Chain/Block/Header.hs
```diff
@@ -0,0 +1,43 @@
...
+  recover pm bytes === Right header
+  where
+  recover :: ProtocolMagic -> BSL.ByteString -> Either Error Header
+  recover pm bytes = do
+    header <- first ErrorD $ decodeFullAnnotatedBytes "Header" decodeAHeader bytes
+    first ErrorH $ verifyAHeader pm header
+    pure $ map (const ()) header
```
ruhatch: Maybe `pure (() $> header)`? Not sure if that's actually more readable though...

ruhatch (commented):
Looks really good. I'd like to get some more eyes on this and also some more documentation, but should be able to get this in tomorrow.

crypto/src/Cardano/Crypto/Signing/Check.hs
```diff
@@ -38,6 +39,10 @@ checkSig
   :: Bi a => ProtocolMagic -> SignTag -> PublicKey -> a -> Signature a -> Bool
 checkSig pm t k x s = checkSigRaw pm (Just t) k (Bi.serialize' x) (coerce s)
 
+checkSigAnnotated
+  :: ProtocolMagic -> SignTag -> PublicKey -> (Annotated a ByteString) -> Signature a -> Bool
+checkSigAnnotated pm t k x s = checkSigRaw pm (Just t) k (annotation x) (coerce s)
+
```
redxaxder, line 45: This function should ultimately replace all uses of checkSig above.

crypto/src/Cardano/Crypto/Signing/Signing.hs
```diff
@@ -147,9 +147,31 @@ proxySign pm t sk@(SecretKey delegateSk) psk m
...
+proxyVerifyAnnotated pm t psig omegaPred m = predCorrect && sigValid
+ where
+  psk = psigPsk psig
+  PublicKey issuerPk        = pskIssuerPk psk
+  PublicKey pdDelegatePkRaw = pskDelegatePk psk
+  predCorrect               = omegaPred (pskOmega psk)
+  sigValid                  = CC.verify
+    pdDelegatePkRaw
+    (mconcat ["01", CC.unXPub issuerPk, signTag pm t, annotation m])
+    (psigSig psig)
```
redxaxder, line 169: Replaces `proxyVerify`

src/Cardano/Chain/Block/Header.hs
```diff
@@ -66,26 +77,38 @@ import           Cardano.Chain.Slotting (SlotId (..), slotIdF)
...
+  , aHeaderProof        :: !(Annotated Proof a)
   -- ^ Proof of body
-  , headerConsensusData :: !ConsensusData
+  , headerConsensusData :: !(AConsensusData a)
   -- ^ Consensus data to verify consensus algorithm
-  , headerExtraData     :: !ExtraHeaderData
+  , aHeaderExtraData    :: !(Annotated ExtraHeaderData a)
   -- ^ Any extra data
-  } deriving (Eq, Show, Generic, NFData)
+  } deriving (Eq, Show, Generic, NFData, Functor)
```
redxaxder, line 102: All values which need their hash or signature verified are replaced with an annotated version.
The types we use in place of `a` are:
 - `()` for headers constructed locally
 - `ByteSpan` for the return type of a decoder
 - `ByteString` for checkable values.

src/Cardano/Chain/Block/Header.hs
```diff
@@ -345,11 +371,31 @@ instance Bi BlockSignature where
...
+      , (annotation . aConsensusDifficulty . headerConsensusData) h
+      , (annotation . aHeaderExtraData) h
+      ]
+    toSign = ToSign
+      (headerPrevHash h)
+      (headerProof h)
+      (headerSlot h)
+      (headerDifficulty h)
+      (headerExtraData h)
+
```
redxaxder, line 398: We can't move entirely away from constructing things during verification since the bytestring signed during the construction of a header is the above, rather than a substring of the encoded header.

binary/src/Cardano/Binary/Class/Annotated.hs
```diff
@@ -0,0 +1,47 @@
...
+decodeAnnotated = annotatedDecoder decode
+
+decodeFullAnnotatedBytes
+  :: (Functor f)
+  => Text
+  -> (forall s. D.Decoder s (f ByteSpan))
+  -> BSL.ByteString
+  -> Either DecoderError (f ByteString)
+decodeFullAnnotatedBytes lbl decoder bytes = (fmap.fmap) (BSL.toStrict . slice bytes) $
+  decodeFullDecoder lbl decoder bytes
```
parsonsmatt, line 47: This is strictly a style thing, but I usually like the way that:

```haskell
fmap (BSL.toStrict . slice bytes) <$> decodeFullDecoder lbl decoder bytes
```

as it saves a few characters and avoids the cluttered `fmap.fmap` thing. I'd prefer to see `(fmap . fmap)`, but I don't have a strong opinion.

crypto/src/Cardano/Crypto/Signing/Signing.hs
```diff
@@ -147,9 +147,31 @@ proxySign pm t sk@(SecretKey delegateSk) psk m
...
+
 -- CHECK: @proxyVerify
 -- | Verify delegated signature given issuer's pk, signature, message
 --   space predicate and message itself.
+{-# WARNING proxyVerify "to remove" #-}
```
parsonsmatt, line 174: Not strictly necessary but I'd make this a `DEPRECATED` pragma instead and include the note "use `proxyVerifyAnnotated` instead"

src/Cardano/Chain/Block/Header.hs
```diff
@@ -66,26 +77,38 @@ import           Cardano.Chain.Slotting (SlotId (..), slotIdF)
...
+  } deriving (Eq, Show, Generic, NFData, Functor)
+
+headerPrevHash :: AHeader a -> HeaderHash
+headerPrevHash = unAnnotated . aHeaderPrevHash
+
+headerProof :: AHeader a -> Proof
+headerProof = unAnnotated . aHeaderProof
+
+headerExtraData :: AHeader a -> ExtraHeaderData
+headerExtraData = unAnnotated . aHeaderExtraData
```
parsonsmatt, line 111: I love this change, as it extends the codebase in new and interesting ways, but not in a way that requires a massive and invasive edit to every bit of code that uses the old variant! :smile:

src/Cardano/Chain/Block/Header.hs
```diff
@@ -220,48 +247,47 @@ instance B.Buildable HeaderError where
...
-    (headerExtraData header)
+  signed = recoverSignedBytes header_
 
   epochId   = siEpoch $ consensusSlot consensus
 
-  consensus = headerConsensusData header
+  consensus = headerConsensusData header_
+
+verifyHeader :: MonadError HeaderError m => ProtocolMagic -> Header -> m ()
+verifyHeader = undefined
```
parsonsmatt, line 287: Is this a `TODO`? I usually prefer `error "some informative note here"` for that sort of thing :slightly_smiling_face:

src/Cardano/Chain/Block/Header.hs
```diff
@@ -345,11 +371,31 @@ instance Bi BlockSignature where
...
 --------------------------------------------------------------------------------
 
+-- | Produces the ByteString that was signed in the block
+recoverSignedBytes :: AHeader ByteString -> Annotated ToSign ByteString
+recoverSignedBytes h = Annotated toSign bytes
+  where
+    bytes = BS.concat
+      [ "\133"
+      -- This is the value of Codec.CBOR.Write.toLazyByteString (encodeListLen 5)
+      -- It is hard coded here because the signed bytes included it as an implementation artifact
```
parsonsmatt, line 385: Is there any way we can factor this magic constant out, or re-derive it somehow? I'd hate for a regression to sneak in if that value ever changes. I imagine it would be extremely difficult to debug.

stack.yaml
```diff
@@ -14,6 +14,11 @@ extra-deps:
     subdirs:
       - .
       - test
+  - git: https://github.com/well-typed/cborg
+    commit: 80fbe0ee5e67a5622e2cb9eaa9d8594a2214322d
+    subdirs:
+      - cborg
```
parsonsmatt, line 20: Can we include a note on why we are pulling in a Git dependency, along with a link to the PR that upstreams this?

test/Test/Cardano/Chain/Block/Header.hs
parsonsmatt: This is also `void :: Functor f => f a -> f ()` :smile:

parsonsmatt (changes requested):
There's a term that is still `undefined`. Otherwise, the design and implementation here looks fantastic.

src/Cardano/Chain/Block/Header.hs
redxaxder: This is slated to be deleted entirely. It's still in use in other code, so this just makes the PR compile.

crypto/src/Cardano/Crypto/Signing/Signing.hs
redxaxder: I would like to delete the current `proxyVerify` an replace it with a new `proxyVerify` which has the type/implementation of `proxyVerifyAnnotated`.

binary/src/Cardano/Binary/Class/Annotated.hs
redxaxder: I have a mental 'chunk' for chained fmaps `(fmap . fmap . fmap)` to recognize what they do. N of them goes N functors deep.

The mix between `fmap` and `<$>` doesn't trigger it for me the same way, so it takes a bit longer to parse it. This might be just a personal mental quirk, though.

binary/src/Cardano/Binary/Class/Annotated.hs
```diff
@@ -0,0 +1,47 @@
+{-# LANGUAGE DeriveAnyClass     #-}
```
dnadales, line 1: Maybe some comments on what the `Annotated` module contains, and what the exposed functions are would be nice.

src/Cardano/Chain/Block/Header.hs
redxaxder: My interpretation of the situation we're in is the following.

Two decisions were made earlier:
  - The Byron implementors decided that a CBOR list gets signed instead of a substring of the header
  - The `cborg` implementors chose this particular byte as their encoding of the list length. (The CBOR specification has multiple encodings of the same integer)

If the value used in the `cborg` library ever changes, we will still want to use the earlier value to maintain compatibility with older blocks.

We're permanently saddled with caring about the particular byte `"\133"` :(.

Do you have an idea for how to make this better?

ruhatch: We could use `serializeEncoding (encodeListLen 5)`? Then it's clearer what's going on and uses the code that generated the original encoding.

EDIT: @redxaxder explained why this doesn't work. Going forward we don't want to enforce a canonical CBOR encoding. There's more than one encoding of the number 5 in CBOR, so we can't rely on this exact byte being used by third-party implementations of a node. 

The real issue is that this byte doesn't appear in the actual serialised `Header`, because the `ToSign` structure is not in the `Header`, but is instead synthesised from it. We therefore cannot know which byte to sign when checking the signature.

So we have to include this *specific* byte in our validation spec to make sure that third-party implementations can check signatures. We could drop it if we change the format, and only enforce full validation for a tail of the blockchain that no longer uses the old format.

crypto/src/Cardano/Crypto/Signing/Signing.hs
```diff
@@ -147,9 +147,31 @@ proxySign pm t sk@(SecretKey delegateSk) psk m
...
+  -> Bool
+proxyVerifyAnnotated pm t psig omegaPred m = predCorrect && sigValid
+ where
+  psk = psigPsk psig
+  PublicKey issuerPk        = pskIssuerPk psk
+  PublicKey pdDelegatePkRaw = pskDelegatePk psk
+  predCorrect               = omegaPred (pskOmega psk)
+  sigValid                  = CC.verify
+    pdDelegatePkRaw
+    (mconcat ["01", CC.unXPub issuerPk, signTag pm t, annotation m])
```
dnadales, line 168: No more re-serialization. Nice!

dnadales (commented):
LGTM!

src/Cardano/Chain/Block/Header.hs
redxaxder: The plan for Shelly is to stop requiring canonical CBOR.

(nb: "canonical" here does not refer to a standard. It refers to a particular set of choices made for our protocol.)

There are multiple bytes that implementors can choose from to encode the same CBOR integer, and these are used to encode the length of lists. For this particular CBOR list, the byte chosen by the node isn't transmitted as part of the payload. 

So it isn't enough to specify that this is a byte corresponding to list length 5. It has to be this exact byte. Otherwise the receiving node cannot validate the signature.

ruhatch:
This was a PR for review only, so I'm closing it.

PR #95, opened 2018-11-01
[#93] Add cardano-mainnet-mirror as a git submodule

### LINKED ISSUE #93: Add cardano-mainnet-mirror as a submodule for testing
We want to run validation over the real blockchain to make sure we're not breaking anything as we move away from the Byron types, serialisation, and validation. We can imagine that the whole blockchain is in a sub-directory by using a submodule. Then tests can just read those files directly as long as the submodule has been updated.

ruhatch:
`cardano-mainnet-mirror` contains a copy of the mainnet blockchain. We can use this to perform tests for validation, deserialisation, etc., without running a full node.

Closes #93

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "cardano-mainnet-mirror"]
+	path = cardano-mainnet-mirror
+	url = git@github.com:input-output-hk/cardano-mainnet-mirror.git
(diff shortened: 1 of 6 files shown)


redxaxder (approved):

redxaxder:
This is way more sensible than just pulling over the first two.

Issue #108, opened 2018-11-06
ledger_test_validate_old


ruhatch:
Test the blockchain layer validation and the ledger layer validation with
1. existing mainnet chain; and
2. executable spec and generator

Issue #96, opened 2018-11-02
Implement UTxO inference rule as validation function


ruhatch:
The UTxO inference rule [here](https://github.com/input-output-hk/cardano-chain/blob/85adbe4d7ca6796eb03c4b2f28a7ee64ab15f8b7/specs/ledger/latex/ledger-spec.tex#L311) involves `Coin`s and the calculation of fees. We would like to change `Coin`s to `Lovelace` and calculate fees in `Lovelace` to simplify things. To make sure the changes don't break validation, I will implement the rule with the old types, check it works on mainnet blocks, and then update the types making sure it is still correct.

PR #109, opened 2018-11-06
[#96] Implement UTxO inference rule

ruhatch:
Added the UTxO inference rule, written to reflect the structure of the corresponding rule in the spec. Feedback on how easy it is to compare and alternative ways of laying things out would be much appreciated.

The tests for this run over the first 15 mainnet epochs, but I have also run it over the whole set of epoch files mirrored at `cardano-mainnet-mirror`. This testing forms part of #108, so we're working on multiple parts of the Shelley plan at once.

I needed to import some more types from `cardano-sl` to get `GenesisUTxO`. In case those are adding noise during review, the important design decisions are happening in:
- Cardano.Chain.Txp.UTxO
- Cardano.Chain.Txp.Validation
- Test.Cardano.Chain.Txp.Validation


src/Cardano/Chain/Txp/Validation.hs
```diff
@@ -0,0 +1,122 @@
...
+  fee <-
+    liftEither
+    .         first (TxValidationCoinError "Fee")
+    $         balanceIn
+    `subCoin` balanceOut
+
+  -- Check that the fee is greater than the minimum
+  unless (minFee <= fee) $ throwError $ TxValidationFeeTooSmall tx minFee fee
+ where
+  txSize              = biSize tx
```
ruhatch, line 89: This will be an `Annotation` once @redxaxder's latest work is merged in

src/Cardano/Chain/Txp/Validation.hs
```diff
@@ -0,0 +1,122 @@
...
+
+  inputUTxO           = S.fromList (NE.toList (_txInputs tx)) <| utxo
```
ruhatch, line 91: I wondered about changing the `Tx` to use a `Set` for the `TxIn`s. Could use [Data.NonEmpty.Set](http://hackage.haskell.org/package/non-empty-0.3.0.1/docs/Data-NonEmpty-Set.html) to keep the non-emptiness guarantees.

src/Cardano/Chain/Txp/UTxO.hs
```diff
@@ -0,0 +1,70 @@
...
+fromList = UTxO . M.fromList
+
+member :: TxIn -> UTxO -> Bool
+member txIn = M.member txIn . unUTxO
+
+union :: MonadError UTxOError m => UTxO -> UTxO -> m UTxO
+union (UTxO m) (UTxO m') = do
+  let m'' = M.union m m'
+  unless (M.size m'' == M.size m + M.size m')
+    $ throwError UTxOOverlappingUnion
```
JaredCorduan, line 47: :+1: :smile:

src/Cardano/Chain/Txp/Validation.hs
```diff
@@ -0,0 +1,122 @@
...
+
+  -- Calculate the 'fee' as the difference of the balances
+  fee <-
+    liftEither
+    .         first (TxValidationCoinError "Fee")
+    $         balanceIn
+    `subCoin` balanceOut
+
+  -- Check that the fee is greater than the minimum
+  unless (minFee <= fee) $ throwError $ TxValidationFeeTooSmall tx minFee fee
```
JaredCorduan, line 87: is "Output balance + fee = input balance" being checked?

.stylish-haskell.yaml
```diff
@@ -130,7 +130,7 @@ steps:
       #
       # - module_name: align under start of module name.
       #   Useful for 'file' and 'group' align settings.
-      list_padding: 4
+      list_padding: 2
```
dnadales, line 133: :confetti_ball: :champagne:

src/Cardano/Chain/Genesis/ProtocolConstants.hs
```diff
@@ -31,4 +40,21 @@ data GenesisProtocolConstants = GenesisProtocolConstants
   , gpcVssMinTTL     :: !VssMinTTL
   } deriving (Show, Eq, Generic)
 
+instance Monad m => ToJSON m GenesisProtocolConstants where
+  toJSON gpc = mkObject
+  -- 'k' definitely won't exceed the limit
```
dnadales, line 45: Why? :D

src/Cardano/Chain/Txp/GenesisUTxO.hs
```diff
@@ -0,0 +1,35 @@
...
+import qualified Cardano.Chain.Txp.UTxO as UTxO
+import Cardano.Crypto
+  (hash)
+
+
+genesisUtxo :: GenesisData -> UTxO
+genesisUtxo genesisData = UTxO.fromList $ utxoEntry <$> preUtxo
+ where
+  preUtxo :: [(Address, Coin)]
+  preUtxo = mappend
```
dnadales: maybe replace [`mappend` by `<>`](https://www.stackage.org/haddock/lts-12.17/base-4.11.1.0/Prelude.html#v:mappend)?

src/Cardano/Chain/Txp/Validation.hs
```diff
@@ -0,0 +1,122 @@
...
+ where
+  txSize              = biSize tx
+
+  inputUTxO           = S.fromList (NE.toList (_txInputs tx)) <| utxo
+
+  calculateMinimumFee = \case
+    TxFeePolicyTxSizeLinear txSizeLinear ->
+      integerToCoin . ceiling $ calculateTxSizeLinear txSizeLinear txSize
+    _ -> panic "Unknown TxFeePolicy"
```
dnadales: Why aren't we using `throwError` here?

src/Cardano/Chain/Txp/Validation.hs
```diff
@@ -0,0 +1,122 @@
...
+
+-- | Validate that 'TxIn' is in the domain of 'UTxO'
+validateTxIn :: MonadError TxValidationError m => UTxO -> TxIn -> m ()
```
dnadales, line 104: I'm just wondering whether we'd like to see any remaining missing inputs.

src/Cardano/Chain/Txp/Validation.hs
dnadales: This seems like a reasonable change. I would expect that we're most of the time using the `TxIns` for lookups, so I don't know if having a list is useful at all. We should definitely change it or make an issue out of it.

src/Cardano/Chain/Txp/Validation.hs
dnadales: Indeed, it seems that we're missing a check between inputs and outputs.

test/test.hs
```diff
@@ -20,6 +21,7 @@ main = runTests
     , Test.Cardano.Chain.Ssc.Bi.tests
     , Test.Cardano.Chain.Txp.Bi.tests
     , Test.Cardano.Chain.Txp.Json.tests
+    , Test.Cardano.Chain.Txp.Validation.tests
```
dnadales, line 24: One thing that I noticed is that we're using the [`runTests`](https://github.com/input-output-hk/cardano-prelude/blob/ac617ef246e09eee2901d75c210e872eed341da8/test/Test/Cardano/Prelude/Tripping.hs#L44) from `Tripping`, however I don't think we're executing only round trip tests.

But the main problem is that when developing (or reviewing :D) we cannot decide which tests to run. Do we want to replace the way we invoke our tests?

test/Test/Cardano/Chain/Txp/Validation.hs
```diff
@@ -0,0 +1,85 @@
...
+  (getEpochFiles)
+
+
+tests :: IO Bool
+tests = do
+  genesisData <- either (panic . sformat build) identity
+    <$> runExceptT (readGenesisData "test/mainnet-genesis.json")
+  let utxo = genesisUtxo genesisData
+  utxoRef <- newIORef utxo
+  files   <- take 15 <$> getEpochFiles
```
dnadales, line 53: Any reason why we choose the number 15?

test/Test/Cardano/Chain/Txp/Validation.hs
```diff
@@ -0,0 +1,85 @@
...
+    properties :: [(PropertyName, Property)]
+    properties = zip (fromString <$> files) (epochValid utxoRef <$> files)
+  checkSequential $ Group "Test.Cardano.Chain.Txp.Validation" properties
+
+data Error
+  = ErrorParseError ParseError
+  | ErrorUTxOValidationError SlotId UTxOValidationError
+  deriving (Eq, Show)
+
+epochValid :: IORef UTxO -> FilePath -> Property
```
dnadales, line 64: I'm just curious about why do we need `IORef's` instead of passing the UTxO around. Would the tests become slower?
My naive expectation would be that the performance would be the same...

dnadales (dismissed):
Looks good, thanks! We only need to address the concern about checking for the relation between inputs and outputs.

src/Cardano/Chain/Txp/Validation.hs
ruhatch: I'm calculating the fee as `input balance - output balance` as @JaredCorduan suggested we could do, so we shouldn't need to check it

src/Cardano/Chain/Genesis/ProtocolConstants.hs
ruhatch: Good question, I assume because it's a parameter from the configuration that we'd never set to be that large, but I don't know if we should assume that.

src/Cardano/Chain/Txp/GenesisUTxO.hs
ruhatch: It was `(<>)` before, but it made it less clear I think. Only because the subexpressions we're so large, so maybe I could give them names and use `(<>)`...

src/Cardano/Chain/Txp/Validation.hs
ruhatch: Good point, this was temporary, but it does represent a `TxValidationError`

src/Cardano/Chain/Txp/Validation.hs
ruhatch: Ah, so collect all missing inputs and return them in the error?

src/Cardano/Chain/Txp/Validation.hs
JaredCorduan: ok, that's fair. :)  you even mentioned that in comment above, I must have missed that.

JaredCorduan (approved):

test/test.hs
ruhatch: `runTests` is generic and should probably be moved from `Tripping` to a more general set of `hedgehog` helpers.

I'm not sure I understand the second part of this! How might you change things?

test/Test/Cardano/Chain/Txp/Validation.hs
ruhatch: Just seemed like a reasonable trade-off between test run time and epoch coverage - open to changing it or even running on the whole chain

test/Test/Cardano/Chain/Txp/Validation.hs
ruhatch: The problem is that `Property` does not return a value! We could lump all of the tests into a single `Property`, but then we cannot see as it progresses through the epochs. I had that originally and was frustrated by the lack of progress indication. So not a performance concern at all :)

ruhatch:
bors +r

src/Cardano/Chain/Txp/GenesisUTxO.hs
dnadales: I'm fine with both :)

src/Cardano/Chain/Txp/Validation.hs
dnadales: Yeah. I'm not saying that we need to do this in this PR, but if this will be a useful feature we could make an issue out of it.

test/test.hs
dnadales: Sorry, I was not clear in my second paragraph :upside_down_face: What I meant was that when working on a specific test suite you only want to run the tests for that test suite. In `hspec` the main function accepts a match parameter `-m`. I don't know if `hedgehog` (or `tasty`?) have equivalent main functions, but I think it'd be a useful feature to have.

But maybe there's another way of running only certain tests that I'm not aware of...

test/Test/Cardano/Chain/Txp/Validation.hs
dnadales: Clear. Thanks!

test/Test/Cardano/Chain/Txp/Validation.hs
mdimjasevic: @ruhatch, it would be useful to have a sentence like this in comments. Anyone reading this source code will ask themselves: "Why 15?"

PR #114, opened 2018-11-07
Thoughts on composition

polinavino:
My thoughts on what building a total state transition out of its parts should look like


mdimjasevic:
@polinavino , I'm trying to understand your proposal. What do you mean when you say _transitions of distinct parts (substates)_? Do you mean `\SubSo` and `\SubSt` are disjoint sets?

polinavino:
> @polinavino , I'm trying to understand your proposal. What do you mean when you say _transitions of distinct parts (substates)_? Do you mean `\SubSo` and `\SubSt` are disjoint sets?

polinavino:
Yes! Like the delegation-related part of the ledger state is made up of DState (for individual keys and their delegations) and PState (for stake pools and retiring stake pools).

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -248,6 +250,47 @@ \section{Composition of Transition Systems}
 acts as an introduction rule for $B$ as determined by the transition rules for
 system $S1$.
 
+When we want to describe a complete transition of a state with separately
+defined rules governing the transitions of distinct parts (substates), one (most
+unambiguous) approach we can take is to define the total transition
+by having a separate rule for each of the substate transitions we need this
+rule to reflect. In particular, let us consider two substates, $\SubSo$ and
```
dnadales: Why is `SubSo` not called `SubS1` and `SubSt` called `SubS1`?

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -248,6 +250,47 @@ \section{Composition of Transition Systems}
...
+$\SubSt$, with $A, B \in \SubSo$ and $A', B' \in \SubSt$, with all other
```
dnadales: Maybe renaming `A` to `A_1` and `A'` to `A_2` make variable types easier to track. Or maybe you could rename `A` to `S_1` and `B` to `S_1'`? Just an idea, feel free to ignore it :D

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -248,6 +250,47 @@ \section{Composition of Transition Systems}
...
+    {\tion[E1]{S2}{(A,A')}{s}{(B,A')}}
+\]
+and
+\[
+  \inference[ExampleS2]
+    {P(A') & z' = f' ~ s' & \tion[E1_{s2}]{S1'}{A'}{z'}{B'}}
+    {\tion[E1]{S2'}{(A,A')}{s'}{(A,B')}}
+\]
+
+where $E1 = E1_{s1} \cup E1_{s2}$. These rules can be applied in sequence,
```
dnadales: The "where ..." part appears indented in the document.

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -248,6 +250,47 @@ \section{Composition of Transition Systems}
...
+to give the full $(A,A') \to (B,B')$ transition.
```
dnadales: Is having two rules different from having a single rule with the two premises that involve `S1` and `S2`?

dnadales: I think a potential problem I see with this is that you might trigger a lot of `ExampleS1` transitions, and only one `ExampleS2`, then the won't proceed in a lockstep manner (if that's what you intended).

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -248,6 +250,47 @@ \section{Composition of Transition Systems}
...
+
+\begin{todo}
+Note that unless $s \wedge s'$ or some other representation of composition
+of the signals for the consequents in ExampleS1 and ExampleS2 is of the
+same type as $s$ and $s'$ (which is not the case if, for example,
+$s$ and $s'$ are certificates and we do not specify some way to build
+a new certificate that includes both),
+it does not seem like we can get away with
+having just one rule (or it may be the most straighforward option to
+describe the system, anyway).
```
dnadales: We might need to make this a bit more precise.

specs/semantics/latex/small-step-semantics.tex
polinavino: I will change this

specs/semantics/latex/small-step-semantics.tex
polinavino: I will rename!

specs/semantics/latex/small-step-semantics.tex
polinavino: It is in the new commit (and the one I am about to make which includes your above comments.

specs/semantics/latex/small-step-semantics.tex
polinavino: If there is a single trigger that causes a total $(A,A') \to (B,B')$
transition, that should be explicitly indicated in a separate rule, i think.

dnadales:
Maybe we should spend half an hour tomorrow to discuss this via video call,
I think it might be easier that way :)

specs/semantics/latex/small-step-semantics.tex
polinavino: Sure! before or after the meeting tomorrow? (I would prefer after)

polinavino: Although, the single signal total transition can be done with the two rules too

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -248,6 +250,79 @@ \section{Composition of Transition Systems}
...
+it is important to make precise what we mean when we have two or more
+transitions in the antecedent, possibly triggered by different signals.
+Let us first consider the case where we define two transitions of two
+sets of state variables,
+$A_1 \to B_1$ and $A_2 \to B_2$. Suppose these are both triggered by the same signal $c$,
+and we wish to express that upon arrivan of a signal $c$,
+there is indeed a complete and inseparable state change, with no intermediate
+$(A_1, B_2)$ state. Here, we can give the complete state transition as a
+rule where the consequent transition $(A_1, A_2, ...) \to (B_1, B_2, ...)$ is triggered
+by that same signal $c$ (or some related signal $f(c)$).
```
nc6, line 270: I don't think this paragraph makes any sense. When we have a transition in the antecedent, we are saying that the existence of such a valid transition is a required condition for the transition in the consequent to be valid. If there are two such transitions, then we have two conditions, but there is no _ordering_ relationship between them. One could impose a _sequencing_ between them (for example, by having the signal for one derived from the RHS of the other), but this is equivalent to writing "let x = 1, y = x" in a mathematical paper; it should not cause us to worry about cases where `x` has been set but `y` has not.

There's also no particular reason why we have to _use_ the results on the RHS of the transitions in the antecedent; I don't think there are any cases where we do it, but it would be fine to for example rely on the transition $A_2 \to B_2$ being valid, but do nothing with the result $B_2$.

In a real implementation, of course, there may be occasion to worry about atomicity, but I think it's not likely to cause problems at this level of granularity (we will presumably commit only at the block level), an is somewhat removed from what we're discussing here.

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -248,6 +250,79 @@ \section{Composition of Transition Systems}
...
+two separate rules with only one of $A_1 \to B_1$ or $A_2 \to B_2$ in
+the antecedent in each.
+
+Now, when we define an inference rule for a complete state transition that
+is defined in terms of two or more transitions of subsets of state variables,
+the first thing we must consider is the signal. Suppose the triggers for the
+two antecedent transitions are $c$ and $c'$. The type of the total transition
+signal is likely to be some combination of $c$ and $c'$, or more generally,
+$f(c,c')$, which is necessarily of the type of the complete state transition
+signal.
```
nc6, line 285: This is often the case, but not necessarily. For example, earlier versions of the delegation spec had something like $A \to^{c} B$ and $C \to^{f(B)} D$ in the antecedent.

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -248,6 +250,79 @@ \section{Composition of Transition Systems}
...
+triggered by the particular value $c = V$, as opposed to one firing as
+a result of $c = V$, and the other only when $c = V' \neq V$. In this case,
+the rule will still not be wrong, only redundant, and should be replaced by
+two separate rules with only one of $A_1 \to B_1$ or $A_2 \to B_2$ in
+the antecedent in each.
+
+Now, when we define an inference rule for a complete state transition that
+is defined in terms of two or more transitions of subsets of state variables,
+the first thing we must consider is the signal. Suppose the triggers for the
+two antecedent transitions are $c$ and $c'$. The type of the total transition
```
nc6, line 282: By `total` you mean the signal in the consequent?

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -248,6 +250,79 @@ \section{Composition of Transition Systems}
...
+signal is likely to be some combination of $c$ and $c'$, or more generally,
+$f(c,c')$, which is necessarily of the type of the complete state transition
+signal.
+
+Note here that if a total state transition is triggered by different types of
+events, such as either by the arrival of
```
nc6, line 288: That would violate the condition earlier this paper, which gives each system a particular signal type.

specs/semantics/latex/small-step-semantics.tex
```diff
@@ -248,6 +250,79 @@ \section{Composition of Transition Systems}
...
+
+To summarize, in most cases, a set of rules governing the transitions of a
+state of a system should be made up of inference rules with only one
+transition in the antecedent, and the signal for the consequent transition
+should be triggered by some embedding of the antecedent transition signal $c \in C$
+into a more general type of transition signal for the whole system, as in
+$C \subseteq C'$. Only in the cases where a signal is intended to trigger a
+complete and inseparable (no intermediate states) transition do we want to
+have two transitions of subsets of state variables with signal $c$ in the antecedent,
+and a $c$-triggered transition in the consequent.
```
nc6, line 311: Hmm, I think you perceive the "flow" of reading these rules differently to me (and differently to the 'how to read a rule' section earlier at https://github.com/input-output-hk/cardano-chain/pull/114/files#diff-f592a79369f418b4c5a597a1208271ebR216). I would say that it's the sub-signals (signals for transitions in the antecedent) which are derived from the signal (and other components) on the LHS of the consequent, not the other way around. There is only one 'triggering' signal, that on the consequent. Everything above is then merely a condition on the resulting state (or transition, depending on how you look at it) in the consequent being valid. Thinking about this way, I think, is somewhat simpler and doesn't give any situations such as you describe here.

nc6 (changes requested):
I would rather not merge this; I think it's not describing the system as described earlier in the paper. It's possible that some of the problems described might persist when re-cast into the existing interpretation framework, but my intuition is that they will all disappear. 

If such problems do persist, then we should at least re-cast this into the same framework as described elsewhere in the paper, or else this is likely to just cause confusion (or maybe this is just me being easily confused!)

specs/semantics/latex/small-step-semantics.tex
polinavino: Yes, sub-signals definitely is the way to describe what the signals in the antecedent should be, I was just thinking that yesterday. Should I take another another crack at writing about this or just leave it?

polinavino: Like, maybe nail down the definition of sub-signal?

ruhatch:
@polinavino @nc6 What's the state of this?

nc6:
@ruhatch I think I am still failing to understand the core motivation behind this. @polinavino might it be worth having another chat about this?

JaredCorduan:
We can open this PR up again if/when we get some examples where it is needed. But thanks @polinavino !

Issue #115, opened 2018-11-07
Add witnesses to UTxO validation


ruhatch:
This should be added on top of the existing UTxO inference rule to make it compositional like the ledger rule.

PR #120, opened 2018-11-08
[#115] Add witness to UTxO validation

ruhatch:
Added `updateUTxOWitness` to perform witness validation commposed with `updateUTxO` to match the style of the specs. This does not perform validation on `ScriptWitness`es.

I also cleaned up some of the other validation code and added more documentation. I have run this validation over the whole of mainnet and it was successful.

### DIFF
--- a/src/Cardano/Chain/Txp/Validation.hs
+++ b/src/Cardano/Chain/Txp/Validation.hs
@@ -2,48 +2,79 @@
 {-# LANGUAGE FlexibleContexts  #-}
 {-# LANGUAGE LambdaCase        #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RankNTypes        #-}
 {-# LANGUAGE TypeApplications  #-}
 
 module Cardano.Chain.Txp.Validation
        ( validateTx
        , updateUTxO
+       , updateUTxOWitness
        , TxValidationError
        , UTxOValidationError
        ) where
 
+
 import Cardano.Prelude
+
 
 import Control.Monad.Except
   (MonadError, liftEither)
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Set as S
+import qualified Data.Vector as V
+
 
 import Cardano.Binary.Class
   (biSize)
 import Cardano.Chain.Common
-  ( Coeff (..)
+  ( Address
+  , Coeff (..)
   , Coin
   , CoinError
+  , Script (..)
   , TxFeePolicy (..)
   , TxSizeLinear (..)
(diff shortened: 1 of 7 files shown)


JaredCorduan (approved):
LGTM, thanks!

src/Cardano/Chain/Txp/Validation.hs
```diff
@@ -65,61 +97,121 @@ validateTx feePolicy utxo tx = do
...
-    utxo
-    tx
-  liftEither . first UTxOValidationUTxOError $ UTxO.union
-    (S.fromList (NE.toList (_txInputs tx)) </| utxo)
-    (txOutputUTxO tx)
+
+  validateTx hardcodedTxFeePolicy utxo tx
+    `wrapError` UTxOValidationTxValidationError
+
+  UTxO.union (S.fromList (NE.toList (_txInputs tx)) </| utxo) (txOutputUTxO tx)
```
dnadales, line 186: This looks (and reads) much better!

src/Cardano/Chain/Txp/Validation.hs
```diff
@@ -65,61 +97,121 @@ validateTx feePolicy utxo tx = do
...
+      `wrapError` UTxOValidationUTxOError
+
+  -- Validate witnesses and their signing addresses
+  mapM_
+      (uncurry $ validateWitness pm sigData)
+      (zip addresses (V.toList witness))
+    `wrapError` UTxOValidationTxValidationError
+
+  -- Update 'UTxO' ignoring witnesses
+  updateUTxO utxo tx
```
dnadales, line 217: The only difference w.r.t. the rules is that we perform the UTxO check after the witness one, which I think is fine. The alternative would have been more awkward:

```haskell
   utxo' <- updateUTxO utxo tx

   -- check witnesses

   return utxo'
```

src/Cardano/Chain/Txp/Validation.hs
```diff
@@ -65,61 +97,121 @@ validateTx feePolicy utxo tx = do
...
+  validateTx hardcodedTxFeePolicy utxo tx
+    `wrapError` UTxOValidationTxValidationError
+
+  UTxO.union (S.fromList (NE.toList (_txInputs tx)) </| utxo) (txOutputUTxO tx)
+    `wrapError` UTxOValidationUTxOError
+
  where
+
   hardcodedTxFeePolicy =
     TxFeePolicyTxSizeLinear $ TxSizeLinear (Coeff 155381) (Coeff 43.946)
```
dnadales, line 192: Are we gonna change this to `44`?

Issue #100, opened 2018-11-05
Implement in Haskell the blockchain specification


mdimjasevic:
A blockchain specification with its inference rules can be found in [specs/chain/latex](https://github.com/input-output-hk/cardano-chain/tree/master/specs/chain/latex). Write a Haskell implementation for it and place it in [specs/chain/hs](https://github.com/input-output-hk/cardano-chain/tree/master/specs/chain/hs).

mdimjasevic:
The code for this will be in this branch: mdimjasevic-100/haskell-prototype.

PR #125, opened 2018-11-12
Initial Haskell implementation of the blockchain specification

mdimjasevic:
This is an initial implementation of the blockchain specification. It also slightly updates the accompanying LaTeX document. There are bits that are still not finished, all related to the delegation interface to the ledger layer, which should be implemented in coordination with whomever is implementing the ledger side of the delegation interface:

1. Implementing the initial `DSIState`, i.e., `initDSIState`.
2. Implement or import delegation interface functions, namely: `delegates`, `newCertsRule` and `updateCerts`,

This PR closes issue #100.


cardano-chain.cabal
```diff
@@ -51,7 +51,6 @@ library
                        Cardano.Chain.Common.Attributes
                        Cardano.Chain.Common.BlockCount
                        Cardano.Chain.Common.ChainDifficulty
-                       Cardano.Chain.Common.Coeff
```
nc6: I'm not seeing what this has to do with this PR?

specs/chain/hs/src/Chain/Blockchain.hs
```diff
@@ -0,0 +1,179 @@
...
+              else Failed SignedMaximumNumberBlocks
+      -- checks that delegation certificates in the signal block are legal
+      -- with regard to delegation certificates in the delegation state
+      legalCerts :: Embed Interf BC => Antecedent BC
+      -- legalCerts = SubTrans env signal rule where
+      legalCerts = SubTrans (_1 . _1) (_3 . to rbCerts) newCertsRule
+
+instance Embed Interf BC where
+  -- stateLens :: Lens' (State BC) (State Interf)
+  stateLens = lens getter setter where
```
nc6: You can just use `_3` as the lens here, although the updated STS framework no longer has this anyway.

specs/chain/hs/src/Data/Queue.hs
```diff
@@ -0,0 +1,42 @@
+-- | Provides a First In First Out queue with amortised constant-time operations
+module Data.Queue
```
nc6, line 2: Any particular reason for rolling your own queue rather than just using `Data.Sequence` from containers (http://hackage.haskell.org/package/containers-0.6.0.1/docs/Data-Sequence.html) which you already have imported?

nc6 (dismissed):
This branch seems to contain a large number of completely unrelated changes (to the code being worked on by Ru and Alex). I've started doing a review, but it would be much easier if the unrelated changes could be excised from this PR first.

cardano-chain.cabal
mdimjasevic: @nc6, I guess the `master` branch was updated between me requesting a PR and the moment you started working on it. I'll rebase my changes on top of the latest `master` and let you know once I'm done.

mdimjasevic:
@nc6, I rebased my branch on top of the latest `master` (which changed again since my comment two hours ago) and aligned the implementation with your latest STS library version.

I will have to check the rest of your comments you made so far.

specs/chain/hs/src/Data/Queue.hs
mdimjasevic: I was surprised not to find easily an up-to-date library with FIFO queues. I believe `Data.Sequence` from `containers` might support FIFO queues in some way, but I took a look at it and it seems to have lots of functions for sequences, but I couldn't spot those for FIFO queues. I am not saying they're not there, it's just that I couldn't find a subset to use. Anyhow, this can easily be replaced if the `containers` library does provide FIFO queue functionality. The functionality is pretty straight-forward and quick to implement so I just rolled my own implementation. This is prototyping anyway.

mdimjasevic:
> @nc6, I rebased my branch on top of the latest `master` (which changed again since my comment two hours ago) and aligned the implementation with your latest STS library version.
> 
> I will have to check the rest of your comments you made so far.

The pull request is ready for reviewing now.

dnadales (approved):

specs/chain/hs/src/Data/Queue.hs
nc6: `Sequence` has O(1) access and insert at both the start and end, which is presumably all you need for a queue?

mdimjasevic: That's good to know. I'll take a look at it in the future. For now I'm leaving it as it is given it's a prototype.

Issue #117, opened 2018-11-08
Change `TxSizeLinear` to use `Lovelace` coefficients


ruhatch:
Right now `TxSizeLinear` is using `Nano` precision coefficients. I have validated that changing from `43.946` to `44` does not break the mainnet validation, so we should be able to safely make this change. This simplifies our accounting and removes some unnecessary rounding issues.

PR #126, opened 2018-11-12
[#117] Change TxSizeLinear coeffs to Lovelace

ruhatch:
This validates the whole of `cardano-mainnet-mirror`. Validation currently uses a hardcoded `TxSizeLinear`, but it should match the one used on mainnet. Once we've decided what the config looks like we should be able to get the `TxSizeLinear` from the `mainnet-genesis.json` file.


JaredCorduan (approved):
LGTM

PR #128, opened 2018-11-15
Updates to the STS framework.

nc6:
- Various functions for applying rules and testing predicates.
- Allow the results of `SubTrans` transitions to be shared more easily.

### DIFF
--- a/specs/ledger/hs/src/witnessed/ledger/Ledger.hs
+++ b/specs/ledger/hs/src/witnessed/ledger/Ledger.hs
@@ -22,23 +22,25 @@ instance STS UTXOW where
   type Signal UTXOW = TxWits
   type Environment UTXOW = Ledger.Simple.ProtocolConstants
   data PredicateFailure UTXOW
-    = UtxoFailure (PredicateFailure UTXO)
+    = UtxoFailure [PredicateFailure UTXO]
     | InsufficientWitnesses
-    deriving Show
+    deriving (Eq, Show)
 
   rules =
     [ Rule [] $ Base (UTxO Map.empty)
     , Rule
-      [ SubTrans _1 (_3 . to body) utxoInductive
-      , Predicate $ \pc utxo tw -> witnessed tw utxo
+      [ SubTrans st1
+      , Predicate $ \(pc, utxo, tw) -> witnessed tw utxo
       ]
       ( Extension . Transition $
-        \pc utxo (TxWits tx _) -> (txins tx ⋪ utxo) ∪ txouts tx
+        \jc -> subTransResult jc st1
       )
     ]
+    where
+      st1 = EmbeddedTransition (to $ \(env, st, sig) -> (env, st, body sig)) utxoInductive
 
 instance Embed UTXO UTXOW where
-  stateLens = id
+  wrapFailed = UtxoFailure
 
 -- |Determine if a UTxO input is authorized by a given key.
 authTxin :: VKey -> TxIn -> UTxO -> Bool
--- a/specs/semantics/hs/src/Control/State/Transition.hs
+++ b/specs/semantics/hs/src/Control/State/Transition.hs
@@ -4,15 +4,23 @@
 {-# LANGUAGE KindSignatures        #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE RankNTypes            #-}
+{-# LANGUAGE ScopedTypeVariables   #-}
+{-# LANGUAGE StandaloneDeriving    #-}
+{-# LANGUAGE TypeApplications      #-}
 {-# LANGUAGE TypeFamilies          #-}
+{-# LANGUAGE UndecidableInstances  #-}
 
 -- | Small step state transition systems.
 module Control.State.Transition where
 
-import Control.Lens
+import           Control.Lens
+import           Data.Bifunctor (first)
+import           Data.Maybe     (catMaybes)
 
 -- | State transition system.
-class (Show (PredicateFailure a))
+class ( Eq (PredicateFailure a)
+      , Show (PredicateFailure a)
+      )
   => STS a where
   -- | Type of the state which the system transitions between.
   type State a :: *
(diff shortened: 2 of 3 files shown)


specs/semantics/hs/src/Control/State/Transition.hs
```diff
@@ -35,66 +43,126 @@ initialRules = filter isInitial rules
...
+initialStates = catMaybes $ applyBase <$> rules @s
+ where
+  applyBase (Rule _ (Base b)) = Just b
+  applyBase _                 = Nothing
+
+-- | Rules applying transitions
+transitionRules :: STS a => [Rule a]
+transitionRules = filter (not . isInitial) rules
+ where
+  isInitial (Rule _ (Base _)) = True
```
mdimjasevic: You already have the identical `isInitial` function in the `where` clause of `initialRules`. Therefore, if you're to keep the definition of `initialRules` separate from that of `transitionRules`, I suggest making the `isInitial` function top-level.

mdimjasevic: To reflect the duality of `initialRules` and `transitionRules`, I suggest defining them using the `Data.List.partition` function: `partition :: (a -> Bool) -> [a] -> ([a], [a])`. Then you'd partition them with a single function call and the `isInitial` function can stay in a `where` clause.

specs/semantics/hs/src/Control/State/Transition.hs
```diff
@@ -35,66 +43,126 @@ initialRules = filter isInitial rules
...
 
+deriving instance (Eq (PredicateFailure sts)) => Eq (PredicateResult sts)
+
+gatherFailures :: [PredicateResult sts] -> [PredicateFailure sts]
+gatherFailures xs = catMaybes $ prToMaybe <$> xs
+ where
+  prToMaybe Passed     = Nothing
+  prToMaybe (Failed x) = Just x
+
+data EmbeddedTransition sub sts where
```
mdimjasevic, line 82: A brief description of the data type would be great to have.

specs/semantics/hs/src/Control/State/Transition.hs
```diff
@@ -35,66 +43,126 @@ initialRules = filter isInitial rules
...
-        -> State sts
-        -> Signal sts
+    :: (   JudgmentContext sts
         -> PredicateResult sts
        )
     -> Antecedent sts
 
+-- | Check the antecedent
+checkAntecedent :: JudgmentContext sts -> Antecedent sts -> PredicateResult sts
+checkAntecedent jc ant = case ant of
```
mdimjasevic, line 106: Can you make this function total by pattern matching on both data constructors of `ant` of the `Antecedent sts` type? I guess you could simply add a case like this:

```haskell
  Predicate p -> p jc
```

If not, I suggest changing the signature of the `checkAntecedent` function to make it total.

specs/semantics/hs/src/Control/State/Transition.hs
```diff
@@ -35,66 +43,126 @@ initialRules = filter isInitial rules
...
+applyRule
+  :: STS s
+  => JudgmentContext s
+  -> Rule s
+  -> Either [PredicateFailure s] (State s)
+applyRule jc (Rule ants con) = case checkAntecedent jc <$> ants of
+  xs | all (== Passed) xs -> Right $ applyConsequent jc con
+  xs                      -> Left $ gatherFailures xs
+
+-- | Apply a rule even if its predicates fail.
```
mdimjasevic, line 151: I'm not getting this one. Why forcefully apply a rule if its antecedents did not pass? Maybe a call would be great to have.

mdimjasevic (commented):
I am not convinced that applying rules regardless of antecedents makes sense. Therefore, I'm leaving my review as comments only at the time being.

specs/ledger/hs/src/witnessed/ledger/Ledger.hs
```diff
@@ -22,23 +22,25 @@ instance STS UTXOW where
...
     [ Rule [] $ Base (UTxO Map.empty)
     , Rule
-      [ SubTrans _1 (_3 . to body) utxoInductive
-      , Predicate $ \pc utxo tw -> witnessed tw utxo
+      [ SubTrans st1
+      , Predicate $ \(pc, utxo, tw) -> witnessed tw utxo
       ]
       ( Extension . Transition $
-        \pc utxo (TxWits tx _) -> (txins tx ⋪ utxo) ∪ txouts tx
+        \jc -> subTransResult jc st1
```
dnadales, line 36: Yes! :) 

What would be the difference between `SubTrans st1` and `subTransResult js st1`? I guess the first one will be used to evaluate the predicate, whereas the second will be used to compute the next state. In this case, will we compute the next state twice?

dnadales (approved):
Looking good. Thanks!

The only thing I wonder is what an STS framework bring us over good old functions. Maybe this will become clear once we incorporate the generators you're working on, but we might need to explain this eventually.

nc6:
To my mind, these _are_ just functions. The wrapping just provides machinery to make it easier to _only_ write the functions for the basic components and then have the system as a whole work, rather than having to assemble everything from scratch for each system.

specs/semantics/hs/src/Control/State/Transition.hs
nc6: Because we need this for generating invalid states.

specs/ledger/hs/src/witnessed/ledger/Ledger.hs
nc6: At the moment, yes. Though we could get around this by changing the calculation in the antecedent to not actually evaluate the result. I guess the alternative would be to switch this to evaluate to `Maybe`? That might be nicer...

specs/semantics/hs/src/Control/State/Transition.hs
nc6: Oh, whoops. For some reason I missed that. Good catch!

nc6:
bors +r

specs/semantics/hs/src/Control/State/Transition.hs
mdimjasevic: If you're talking about the `type State a :: *` from the `STS` class, I could use an explanation how is generating such states conditioned by having illegal rule applications. Can't you generate illegal states regardless of rules?

Issue #131, opened 2018-11-16
Remove unused light delegation


ruhatch:
There is not currently any light delegation or `BlockPSignatureLight`s on mainnet, so we can safely remove these.

PR #132, opened 2018-11-16
[#131] Remove unused light delegation

ruhatch:
This removes lightweight delegation, which was unused. I commented where the `SignTag` and encoded `Word8` tag were used in case this is reintroduced at some point.


crypto/src/Cardano/Crypto/Signing/Tag.hs
```diff
@@ -26,7 +26,7 @@ signTag protocolMagic = \case
   SignCommitment     -> "\x05" <> network
   SignUSVote         -> "\x06" <> network
   SignMainBlock      -> "\x07" <> network
-  SignMainBlockLight -> "\x08" <> network
+  -- "\x08" was used for SignMainBlockLight, but was never used in mainnet
```
dnadales, line 29: Great that you documented this!

dnadales (approved):

PR #142, opened 2018-11-20
Added annotated types and associated hashing/decoding

redxaxder:


binary/cardano-binary.cabal
```diff
@@ -28,8 +28,10 @@ library
                        Cardano.Binary.Class.Core
                        Cardano.Binary.Class.Drop
                        Cardano.Binary.Class.Primitive
+                       Cardano.Binary.Class.Annotated
 
-  build-depends:       base
+  build-depends:       aeson
```
ruhatch: I like to keep `base` above `aeson` because we'll never change it

binary/src/Cardano/Binary/Class/Annotated.hs
```diff
@@ -0,0 +1,83 @@
...
+  compare = compare `on` unAnnotated
+
+instance ToJSON b => ToJSON (Annotated b a) where
+  toJSON = toJSON . unAnnotated
+
+instance FromJSON b => FromJSON (Annotated b ()) where
+  parseJSON j = flip Annotated () <$> parseJSON j
+
+-- | A decoder for a value paired with an annotation specifying the start and end
+-- of the sconsumed bytes.
```
ruhatch: Small type here `sconsumed -> consumed`

binary/src/Cardano/Binary/Class/Annotated.hs
```diff
@@ -0,0 +1,83 @@
...
+annotatedDecoder :: D.Decoder s a -> D.Decoder s (Annotated a ByteSpan)
+annotatedDecoder d = D.decodeWithByteSpan d
+  <&> \(x, start, end) -> Annotated x (ByteSpan start end)
+
+-- | A decoder for a value paired with an annotation specifying the start and end
+-- of the sconsumed bytes.
```
ruhatch: Same type and maybe mention the `Bi` constraint on this version

binary/src/Cardano/Binary/Class/Annotated.hs
```diff
@@ -0,0 +1,83 @@
...
+  :: (Functor f)
+  => Text
+  -> (forall s . D.Decoder s (f ByteSpan))
+  -> BSL.ByteString
+  -> Either DecoderError (f ByteString)
+decodeFullAnnotatedBytes lbl decoder bytes =
+  (fmap . fmap) (BSL.toStrict . slice bytes)
+    $ decodeFullDecoder lbl decoder bytes
+
+class Decoded f where
```
ruhatch, line 77: Maybe comment what this is for and why we chose this implementation

cardano-chain.cabal
```diff
@@ -21,6 +21,7 @@ library
   hs-source-dirs:      src
   exposed-modules:
                        Cardano.Chain.Block
+                       Cardano.Chain.Block.Header
```
ruhatch: Why do you need to expose this?

stack.yaml
```diff
@@ -14,6 +14,12 @@ extra-deps:
     subdirs:
       - .
       - test
+  - git: https://github.com/well-typed/cborg # https://github.com/well-typed/cborg/pull/185
```
ruhatch, line 17: You should probably upstream this to the `snapshot.yaml` in `cardano-prelude` and update the resolver references here

src/Cardano/Chain/Block/Block.hs
```diff
@@ -139,16 +169,18 @@ encodeBlock block = encodeListLen 2 <> encode (1 :: Word) <> encode block
 --   now deprecated these explicit boundary blocks, but we still need to decode
 --   blocks in the old format. In the case that we find a boundary block, we
 --   drop it using 'dropBoundaryBlock' and return a 'Nothing'.
-decodeBlock :: Decoder s (Maybe Block)
-decodeBlock = do
+decodeABlockOrBoundary :: Decoder s (Maybe (ABlock ByteSpan))
```
ruhatch, line 172: Aha, so we want the bytes of the boundary block even though we're dropping the actual block! I think this is going to be a pain, but we need to annotate the dropping function to recover the offsets and then return an `Either (Annotated () ByteSpan) (ABlock ByteSpan)`. Let me know if this needs more clarification!

ruhatch (approved):
Looks really good in general, just a few small things. I think we can do the annotation of the boundary block in another PR.

Issue #146, opened 2018-11-20
Import and clean up configuration functions


ruhatch:
We will probably end up changing the configuration stuff later on, but for now we can use clean up versions of `cardano-sl` stuff. We should bring over the `Config` type which contains `GenesisData`, `GenesisHash`, and `Maybe GeneratedSecrets` for testnet, as well as constructors `mkConfig` and `mkConfigFromStaticConfig`.

PR #147, opened 2018-11-20
[#146] Import and clean up configuration functions

ruhatch:
This imports some functions for constructing configurations from `cardano-sl`. This will probably be reworked later down the line, but it is definitely cleaner than the old version.

### DIFF
--- a/src/Cardano/Chain/Genesis/Config.hs
+++ b/src/Cardano/Chain/Genesis/Config.hs
@@ -1,12 +1,31 @@
+{-# LANGUAGE BangPatterns      #-}
+{-# LANGUAGE FlexibleContexts  #-}
+{-# LANGUAGE LambdaCase        #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeApplications  #-}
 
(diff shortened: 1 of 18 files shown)


src/Cardano/Chain/Genesis/Config.hs
```diff
@@ -106,3 +139,175 @@ instance FromJSON StaticConfig where
...
+configNonAvvmBalances :: Config -> GenesisNonAvvmBalances
+configNonAvvmBalances = gdNonAvvmBalances . configGenesisData
+
+configBlockVersionData :: Config -> BlockVersionData
+configBlockVersionData = gdBlockVersionData . configGenesisData
+
+configAvvmDistr :: Config -> GenesisAvvmBalances
+configAvvmDistr = gdAvvmDistr . configGenesisData
+
+canonicalGenesisJson :: GenesisData -> (LByteString, Hash Raw)
```
redxaxder: Why not ` :: GenesisData ->  (LByteString, GenesisHash)` ?

redxaxder (approved):

PR #151, opened 2018-11-21
Preliminary FromCBOR/ToCBOR

redxaxder:


binary/src/Cardano/Binary/Class/FromCBOR.hs
```diff
@@ -0,0 +1,20 @@
+
```
ruhatch, line 1: Random newline!

binary/src/Cardano/Binary/Class/FromCBOR.hs
```diff
@@ -0,0 +1,20 @@
...
+{-# LANGUAGE DefaultSignatures    #-}
```
ruhatch, line 2: Shouldn't need this I don't think

binary/stack.yaml
```diff
@@ -10,6 +10,11 @@ extra-deps:
     subdirs:
       - .
       - test
+  - git: https://github.com/well-typed/cborg # https://github.com/well-typed/cborg/pull/185
```
ruhatch, line 13: Yeah, we should definitely upstream this. I'll add a small PR tomorrow to do that though!

ruhatch (approved):

PR #153, opened 2018-11-22
Bump snapshot for cborg peekByteOffset

ruhatch:
Update the snapshot to the latest version from `cardano-prelude`, which contains the `peekByteOffset` code in `cborg`.


ruhatch:
bors try

Issue #137, opened 2018-11-19
Specify the block size


mdimjasevic:
As Duncan said, now that we have the [| x |] serialised syntax, we can talk about sizes of blocks.

Add to the blockchain layer the specification for the block size. In particular, in validating the old blocks, we must apply different size limits for normal blocks vs epoch boundary blocks (EBBs). For normal blocks we use the limit that comes from the protocol parameter for that purpose. For EBBs we apply a fixed 2Mb limit.

nc6:
Do we want this in the chain spec? This feels like something that should be in the consensus layer. cc @dcoutts

mdimjasevic:
The PR #154 addressed this issue.

Issue #138, opened 2018-11-19
Specify the block header size


mdimjasevic:
The updateable protocol parameters have limits for block sizes, header sizes, and transaction sizes. Specify the block header size as part of the blockchain layer specification.

mdimjasevic:
The PR #154 addressed this issue.

PR #154, opened 2018-11-22
Specify block size and block header size

mdimjasevic:
This extends the blockchain layer specification with checks on the block size and the block header size.

The pull request closes issue #137 and issue #138.


specs/chain/latex/blockchain-spec.tex
```diff
@@ -292,6 +319,61 @@ \subsection{Heavyweight Delegation Certificates}
 \subsection{Blocks}
 \label{sec:blocks}
 
+\begin{figure}
+  \emph{Primitive types}
+  %
+  \begin{align*}
+    ix & \in \BlockIx & \text{block index}\\
+    h  & \in \Hash   & \text{hash}\\
+    pp & \in \ProtParams & \text{protocol parameters}
```
dnadales, line 328: In the ledger spec I modeled the protocol parameters as a map from parameter names to their values. This allows me to update the map in the ledger spec, when an update proposal comes in. Should we create an issue to align the two specs?

specs/chain/latex/blockchain-spec.tex
```diff
@@ -292,6 +319,61 @@ \subsection{Heavyweight Delegation Certificates}
 \subsection{Blocks}
 \label{sec:blocks}
 
+\begin{figure}
+  \emph{Primitive types}
```
dnadales, line 323: Is there a difference between abstract and primitive types? Nowadays I just use "abstract types".

specs/chain/latex/blockchain-spec.tex
```diff
@@ -292,6 +319,61 @@ \subsection{Heavyweight Delegation Certificates}
...
+  \emph{Projection functions}
+  %
+  \begin{align*}
+    \fun{gbKeys} & \in \GBlock \totalf \VKeyGen^n & \text{block keys} \\
+    \fun{\rbixname} & \in \RBlock \totalf \BlockIx & \text{block index} \\
+    \fun{\hashname} & \in \Block \totalf \Hash
+      & \text{previous block's hash} \\
+    \fun{\rbslname} & \in \RBlock \totalf \Slot & \text{block slot} \\
+    \fun{\rbsignername} & \in \RBlock \totalf \VKey & \text{block signer} \\
+    \fun{\rbcertsname} & \in \RBlock \totalf \powerset{\HCert}
```
dnadales, line 353: Are block certificates present in the types in our implementation? As far as I can see they are not part of the body (but they might be somewhere else):

```haskell
-- | 'Body' consists of payloads of all block components
data ABody a = ABody
  { bodyTxPayload     :: !(ATxPayload a)
  -- ^ Txp payload
  , bodySscPayload    :: !SscPayload
  -- ^ Ssc payload
  , bodyDlgPayload    :: !(Delegation.APayload a)
  -- ^ Heavyweight delegation payload (no-ttl certificates)
  , bodyUpdatePayload :: !(Update.APayload a)
  -- ^ Additional update information for the update system
  } deriving (Eq, Show, Generic, Functor, NFData)
```

Maybe it is good to check this with @ruhatch to make sure we can implement this spec.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -322,71 +404,30 @@ \subsection{Blocks}
 projection functions for the following:
 %
 \begin{itemize}
-\item a block index,
 \item a block hash pointing to the block's predecessor,
-\item a slot it is in,
+\item a block index,
```
dnadales, line 408: Is the block index the same as the block number?

specs/chain/latex/blockchain-spec.tex
```diff
@@ -322,71 +404,30 @@ \subsection{Blocks}
...
 \item a verification key of the block signer,
-\item a set of heavyweight delegation certificates in the body, and
-\item a cryptographic signature.
+\item a set of heavyweight delegation certificates in the body,
+\item a slot it is in,
+\item data that is the block body,
```
dnadales, line 412: The certificates (described above) are also part of the block body.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -322,71 +404,30 @@ \subsection{Blocks}
...
 
 
-%
 For every block, it is possible to compute its hash with the
 $\fun{\hashofblockname}$ function.
 %
 In the case of the genesis block, this simply returns the hash field in the
 block.
 %
+Furthermore, note that the $\fun{\maxblocksizename}$ parameter of a
```
dnadales, line 424: If I recall correctly we'd have a maximum size for regular and EB blocks. Would it make sense to have two functions so that the reader doesn't have to remember that "maxblocksize" refers to regular blocks? Maybe `maxRegBS` and `maxEBBS` (or their long version names).

specs/chain/latex/blockchain-spec.tex
```diff
@@ -472,7 +513,12 @@ \subsection{Block Validity}
   corresponding verification key in the genesis block.
   %
   The certificate therefore carries a a cryptographic signature $\sigma'$,
-  and
+\item new delegation certificates that are in the body of $b$ are not in
```
dnadales, line 516: Consider explaining what "certificate in conflict" means in this context.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -481,6 +527,16 @@ \subsection{Block Validity}
   in the range $1/5 \leq t \leq 1/4$.
 \end{enumerate}
 
+\begin{align}
+  \label{eq:blocksizelimit}
+  \blocksizelimit{b}{pp} = \
+  \begin{cases}
+    \maxblocksize{pp} & \text{if $b$ is the genesis block or } \isebb{b} \equiv \bot \\
```
dnadales: The two `p`'s appear separated in the generated doc. Consider using `\var`.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -481,6 +527,16 @@ \subsection{Block Validity}
...
+    2^{21} & \text{otherwise}
```
dnadales, line 535: Why not leaving this abstract?

specs/chain/latex/blockchain-spec.tex
```diff
@@ -644,6 +691,8 @@ \subsection{Blockchain Extension}
     {\begin{array}{c}
       ix = \rbix{b} \wedge vk_d = \rbsigner{b} \wedge sl = \rbsl{b} \\
       \hash{b} \equiv \hashofblock{p} \\
+      \bsize{b} \leq \blocksizelimit{b}{pp} \\
+      \bhdrsize{b} \leq \maxheadersize{pp} \\
       \delegation{ds}{vk_s}{vk_d} \\
```
dnadales, line 696: BTW, I just realized `vk_s` seems to be a free variable in the rule, isn't it?

dnadales (approved):

specs/chain/latex/blockchain-spec.tex
ruhatch: Yeah, won't we need an interface similar to the delegation one to share the `BlockVersionData` or whatever that's renamed to?

mdimjasevic:
Thank you for your comments, @dnadales and @ruhatch! Given that I'll be away for the rest of the week, I'll merge this as it is now so it doesn't drift away from the branch. When I get back next week, I'll address your concerns in a separate PR as they don't sound like blockers.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: @dnadales , I got the impression you're suggesting that your map allows you to do something that one cannot do with the abstract `ProtParams` type. Is that right? I don't see why one wouldn't be able to update a `ProtoParams`.

dnadales: If you don't need to do anything with the `\ProtParams` then they can remain abstract.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: What is a block number?

specs/chain/latex/blockchain-spec.tex
mdimjasevic: I admit there's room for improvement. I should work on making these more specific and non-overlapping.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: I don't see a strong case for either way, so I'll keep it the way it is.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: I intentionally did not discuss this in detail in the blockchain specification. It is the ledger layer that decides what is a "certificate in conflict" in a given delegation context.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: Is there anything wrong with having it specific?

specs/chain/latex/blockchain-spec.tex
mdimjasevic: I'm addressing this in a separate PR.

specs/chain/latex/blockchain-spec.tex
mdimjasevic: Because it is a straightforward specific number. Is there any harm in having it this way?

PR #156, opened 2018-11-22
First draft of the small step rules for the update mechanism

### LINKED ISSUE #64: Formalize the Byron update rules
# Main Goal

Model the evolution of the [`BlockVersionData`](https://github.com/input-output-hk/cardano-sl/blob/a791f11bc3443628fd168f896dba93fe8d1f1d59/chain/src/Pos/Chain/Update/BlockVersionData.hs#L30) parameters in the ledger state.

## Aspects to be modeled

0. Authentication: the proposal needs to be properly signed.
1. Authorization: the proposal needs to be signed by the genesis keys or one of the delegates.
2. Voting: genesis key holders or delegates can only vote on a proposal after it is posted.
3. Voting deadlines: voting ends with a majority of the voters (4/7 if we assume no state) agree on a proposal. Bear in mind that a rule that finishes the voting period after 4/7 positive votes will succeed to validate the existing chain only if the distribution of the stake is roughly equal (which might not be the case, and we'd have to revisit this rule).
4. Block-version changes: when the voting ends, we need to update the block-chain parameters according to the proposed version.

# Current subtasks

- [x] Determine which information do we need to keep track of in the ledger state:
    - [x] A subset of the information in `BlockVersionData`.
    - [x] Something else?
- [x] Determine whether software updates are related to protocol (`BlockVersionData`) updates.

# References

- https://cardanodocs.com/technical/updater/
- https://cardanodocs.com/cardano/update-mechanism/
- https://github.com/input-output-hk/cardano-sl/blob/develop/docs/auxx.md#proposing-an-update
- https://github.com/input-output-hk/cardano-sl/blob/develop/wallet-new/docs/updates.md
- https://github.com/input-output-hk/cardano-sl/blob/develop/docs/proposals/update-downloader.md

dnadales:
This is a first draft of the formalization of the Byron update mechanism (in the simplified OBFT setting) although all the aspects of it should be covered, including:

0. Submitting new protocol-update proposals.
1. Confirming proposals.
2. Adoption new protocol versions, as specified by a proposal.

I have made several simplifications w.r.t. the current implementation of the update mechanism, specially where it was not clear to me why the current system was implemented in the way it was, and I could not get a clarification about it. If these simplifications are correct, we'd be removing a great deal of complexity from the current implementation. However I doubt I will get this right on the first iteration, but nevertheless I hope we have a solid body of rules that is simple and extensible enough, so that the necessary modifications can be made without making the specification much more complex.

Closes #64.


specs/ledger/latex/update-mechanism.tex
```diff
@@ -0,0 +1,1237 @@
...
+      \var{upa} & \type{UpdAttr} & \text{update attributes}\\
+    \end{array}
+  \end{equation*}
+  %
+  \emph{Derived types}
+  \begin{equation*}
+    \begin{array}{r@{~\in~}l@{~=~}r@{~\in~}lr}
+      \var{pv} & \ProtVer & (\var{maj}, \var{min}, \var{alt})
+      & (\mathbb{N}, \mathbb{N}, \mathbb{N}) & \text{protocol version}\\
+      \var{pps} & \PPMMap & \var{pps} & \ProtPm \mapsto \Value
```
JaredCorduan, line 88: `Value` can be any type, and varies from parameter to parameter, right?

specs/ledger/latex/update-mechanism.tex
```diff
@@ -0,0 +1,1237 @@
...
+        \var{dms} & \VKeyGen \mapsto \VKey & \text{delegation map}\\
+        \var{cps} & \UPropId \mapsto \BkNr & \text{confirmed proposals}\\
+        \var{rups} & \Rups
+                             & \text{registered update proposals}\\
+      \end{array}\right)
+  \end{align*}
+  %
+  \emph{Block protocol version registration states}
+  \begin{align*}
+    & \BVRState
```
JaredCorduan, line 838: We've already chatted about this today, but I thought I'd leave a note here just to make sure we are on the same page.  I don't think we need to store which nodes are ready to update in the ledger state, but can instead take this from the environment given by the blockchain layer.

JaredCorduan (approved):
This is great @dnadales ! I know today we talked about some thing regarding how to handle the versioning, and separating the software updating from the protocol constants, but I think this can be done next in a new PR. thanks!

specs/ledger/latex/update-mechanism.tex
dnadales: As they are now, the ledger rules need to keep track of how many nodes are saying "I'm ready to upgrade to this version" to be able to apply the [block version adoption algorithm](https://github.com/input-output-hk/cardano-sl/blob/develop/docs/block-processing/us.md#block-version-adoption-algorithm). So maybe we should discuss this a bit further...

specs/ledger/latex/update-mechanism.tex
dnadales: Indeed. In Haskell we'd have to model this as a product type, as it is now. However for the formal spec it is handier in this way :)

JaredCorduan: sounds good!

specs/ledger/latex/update-mechanism.tex
```diff
@@ -0,0 +1,1241 @@
...
+    version, then the proposed minor version must be incremented by one.
+  \item If the proposed major version is larger than the current major version,
+    then the proposed minor version must be zero.
+  \end{itemize}
+\item must be consistent with the current protocol parameters:
+  \begin{itemize}
+  \item the proposal size must not exceed the maximum size specified by
+    the current protocol parameters,
+  \item the proposed new maximum block should be not greater than twice current
+    maximum block size, and
```
nc6, line 208: What's the motivation for this restriction? Might be nice to reference it here.

specs/ledger/latex/update-mechanism.tex
```diff
@@ -0,0 +1,1241 @@
...
+\end{itemize}
+
+\begin{figure}[htb]
+  \begin{equation}
+    \label{eq:rule:voting}
+    \inference
+    {
+      \var{pid} = \vPropId{v} & \var{vk} = \vCaster{v} &
+      \var{vts}_{\var{pid}} =
+      \{ (\var{pid}, \var{vk_s}) \mid \var{vk_s} \mapsto \var{vk} \in \var{dms} \}\\
```
nc6, line 592: So the "voting only by genesis keys" is just satisfied here by vts_pid being empty otherwise?

specs/ledger/latex/update-mechanism.tex
```diff
@@ -0,0 +1,1241 @@
...
+associated with the registration of candidate protocol versions present in
+blocks. Some clarifications are in order:
+\begin{itemize}
+\item The $k$ parameter is used to determined when a confirmed proposal is
+  stable. Given we are in a current block $b_n$, all update proposals confirmed
+  at or before block $b_n - k$ are deemed stable.
+\item The type $\Rups$ is equal to
+  $\UPropId \mapsto (\ProtVer \times \PPMMap)$, we use this local
+  definition to be able to fit the description of the environment in the page
+  width.
```
nc6, line 813: Add to #overlyhonestmethods?

nc6 (approved):
A veritable _tour de force_! I haven't followed the discussions here too closely, but from a general inspection this looks great!

specs/ledger/latex/update-mechanism.tex
dnadales: Good question :) There are a lot of such constraints that are implemented like this without giving any rationale :/ We should definitely revisit these!

specs/ledger/latex/update-mechanism.tex
dnadales: Indeed! Although this silently ignores the fact that a non-delegate of a genesis key tried to vote. I don't know if we want to specify the `vts_pid \neq \emptyset` here, or leave the reporting to the implementation.

specs/ledger/latex/update-mechanism.tex
dnadales: Yeah, it sounds silly... but I didn't want to use the `Rups` throughout since people would have to go back and forth between definitions, but the page width constraints are ruthless. And you complain about the format tyrant :P

PR #159, opened 2018-11-26
[#158] Implement blockchain validation spec without delegation

### LINKED ISSUE #134: Validate the size of unused epoch boundary blocks
We should validate that epoch boundary blocks are less than or equal to 2MB in size. This is different to how we will be validating normal blocks.

### LINKED ISSUE #152: Get bytes of epoch boundary blocks
The annotation work didn't add annotations to the `decodeBlockOrBoundaryBlock` for the boundary block case, so we cannot get the hashes of the boundary blocks. This should add some sort of `AnnotatedDropper` to allow dropping, but returning the bytes from things.

### LINKED ISSUE #158: Implement blockchain validation spec without delegation
This comes before the implementation of the delegation state controlled by the ledger layer, so it ignores the concerns of that part of the validation.

ruhatch:
This is a first implementation of the blockchain validation spec that ignores the ledger side of the delegation state. It just assumes that any delegation we see is valid. This PR also adds the `BlockValidationData` necessary to validate boundary blocks and bumps the version of `cardano-mainnet-mirror` to bring in the boundary files, which weren't stored inside the epoch files.

Closes #158, closes #152, closes #134


src/Cardano/Chain/Block/Block.hs
```diff
@@ -162,36 +170,60 @@ decodeABlock = do
 encodeBlock :: Block -> Encoding
 encodeBlock block = encodeListLen 2 <> encode (1 :: Word) <> encode block
 
+
+data ABlockOrBoundary a
+  = ABOBBlock (ABlock a)
```
dnadales, line 175: Why not just `Block = RegularBlock ... | BoundaryBlock ...`?

mdimjasevic:
@ruhatch, my apologies, but I won't be able to review your PR until the next week.

src/Cardano/Chain/Block/Block.hs
```diff
@@ -162,36 +170,60 @@ decodeABlock = do
...
-  dropBoundaryBody
-  dropBoundaryExtraBodyData
+  Annotated (Annotated hh bs) (ByteSpan start end) <- annotatedDecoder $ do
+    enforceSize "BoundaryBlock" 3
+    aHeaderHash <- annotatedDecoder dropBoundaryHeader
+    dropBoundaryBody
+    dropBoundaryExtraBodyData
+    pure aHeaderHash
+  pure $ BoundaryValidationData
+    { boundaryBlockLength = end - start
```
dnadales, line 223: It is annoying but would it make sense to assert that `end <= start`?

src/Cardano/Chain/Block/Block.hs
```diff
@@ -304,6 +342,9 @@ blockLeaderKey = headerLeaderKey . blockHeader
 blockDifficulty :: ABlock a -> ChainDifficulty
 blockDifficulty = headerDifficulty . blockHeader
 
+blockToSign :: ABlock a -> ToSign
```
dnadales, line 345: One question regarding this `ToSign` time: I've seen it in the `cardano-sl` being used in such a way that what it denotes is not data to sign, but data that has been signed. Is its meaning here different? (sorry for the small digression)

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -0,0 +1,240 @@
...
+-- | A mapping from 'VerificationKey's to the 'SlotId's of any slots they have
+--   signed in the last k blocks
+newtype SigningHistory = SigningHistory
+  { unSigningHistory :: Map StakeholderId [SlotId]
+  }
+
+lookup :: PublicKey -> SigningHistory -> Maybe [SlotId]
+lookup pk = M.lookup (mkStakeholderId pk) . unSigningHistory
+
+restrictSlotId :: BlockCount -> SlotId -> [SlotId] -> [SlotId]
```
dnadales: Maybe add some comments on what this does exactly?

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -0,0 +1,240 @@
...
+
+
+--------------------------------------------------------------------------------
+-- SigningHistory
+--------------------------------------------------------------------------------
+
+-- | A mapping from 'VerificationKey's to the 'SlotId's of any slots they have
+--   signed in the last k blocks
+newtype SigningHistory = SigningHistory
+  { unSigningHistory :: Map StakeholderId [SlotId]
```
dnadales: Here the `SlotId` differs from the one from the spec, I wonder whether we should use `FlatSlotId` here (and make an issue to rename this to something like `GlobalSlotId`).

src/Cardano/Chain/Block/Block.hs
ruhatch: This is because we won't have boundary blocks anymore, so we want `Block` to refer to an actual block. The boundary stuff will only appear during validation of the old chain and won't be referred to elsewhere, so it will be nice to assume the a block is just a block throughout the rest of the code.

src/Cardano/Chain/Block/Validation.hs
dnadales: The other question is whether we need to keep the `SlotId`'s. I'd expect that it is enough to keep the signers of the last `K * t` slots. So we'd only need a sequence of `SlotId`'s.

src/Cardano/Chain/Block/Block.hs
ruhatch: Sure, although it should be impossible - one thing we could do is add a length function for `ByteSpan` and bounds checking there to keep this cleaner.

src/Cardano/Chain/Block/Block.hs
ruhatch: Yeah, it is both the data that should be signed and also the data that was signed (ideally ;). It could probably do with a better name!

src/Cardano/Chain/Block/Validation.hs
ruhatch: It's not quite clear to me what you mean here!

ruhatch: Also should it be `2K * t` slots?

dnadales: > It's not quite clear to me what you mean here!

Do we need a to know which slot was signed by which key? Isn't enough to know the sequence of the last `K * t` signers? 

> Also should it be 2K * t slots?

I was referring to the `K * t` in the spec.

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -0,0 +1,240 @@
...
+  let signingHistory = cvsSigningHistory cvs
+
+  -- Retrieve the signing history for the delegator
+  delegatorSlots <- length <$> maybe
+    (throwError $ ChainValidationMissingDelegator delegator)
+    pure
+    (lookup delegator signingHistory)
+
+  -- Check that 'delegator' hasn't delegated too many previous blocks
+  let threshold = k `div` 5
```
dnadales: Should we make 5 a named constant?

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -0,0 +1,240 @@
...
+        (proxyVerifyDecoded
+          pm
+          SignMainBlockHeavy
+          signature
+          (const True)
+          (recoverSignedBytes $ blockHeader b)
+        )
+        (throwError $ ChainValidationInvalidSignature (blockSignature b))
+      let psk = psigPsk signature
+      pure (pskIssuerPk psk, pskDelegatePk psk)
```
dnadales, line 235: So it seems that the the information of who is signing on behalf of whom is present in the block, shouldn't we update our spec to match this?

dnadales (approved):
Looking good.

My main concerns are:

- Do we need a map for keeping the history of the signers, I'd expect a sequence would suffice.
- Maybe it'd be better to work with global slot id's (as in the spec), so we make the conversion from local to global slot id when reading the data from the blocks.

src/Cardano/Chain/Block/Validation.hs
ruhatch: So we need to calculate the number of blocks signed by any individual signer in the last `K * t` blocks. If we just keep the sequence then we need to run over the whole sequence of `K * t` signers and count the signatures from each one. If we wanted to just keep the list of signers then we could also keep a `Map StakeholderId SlotCount` that we update incrementally. That seems like a reasonable solution actually. What do you think?

dnadales: The problem with that might be that you need to trim the map, so you wouldn't know where to subtract.

PR #177, opened 2018-11-29
Build Haskell specs in CI.

nc6:
This doesn't currently exploit any caching (sounds like they should be built in hydra for that), but at least means we are building the specs. It takes less time than to build the cardano-chain package, so given they're done in parallel this shouldn't add any time to CI.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -19,6 +19,12 @@ steps:
     agents:
       system: x86_64-linux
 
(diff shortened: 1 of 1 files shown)


ruhatch (approved):
LGTM

nc6:
OK. Going to label this for merge, and then make sure we revert it once the hydra build is working correctly.

nc6:
Closing in favour of the other approach!

Issue #165, opened 2018-11-26
Implement delegation state and validation rules


ruhatch:
Implement the delegation part of the ledger spec. This provides a `DelegationState` and `delegates` function for use in the chain validation. It also implements the validation rules and updates that state.

PR #180, opened 2018-11-29
[#165] Implement delegation state and validation rules

ruhatch:
This implements the delegation state and validation rules. It also implements the blockchain-ledger interface and integrates that with chain validation.

Please pay special attention to the `initialInterfaceState` implementation, as I would like feedback on whether that looks reasonable. In order to get that working I had to add an extra condition to `activateDelegation`.


cardano-chain.cabal
```diff
@@ -60,9 +60,10 @@ library
                        Cardano.Chain.Common.TxFeePolicy
                        Cardano.Chain.Common.TxSizeLinear
 
-                       Cardano.Chain.Delegation.HeavyDlgIndex
+                       Cardano.Chain.Delegation.Certificate
                        Cardano.Chain.Delegation.Payload
                        Cardano.Chain.Delegation.Undo
+                       Cardano.Chain.Delegation.Validation
```
dnadales, line 66: Just a minor thing, we have `Cardano.Chain.Delegation` for coding the delegation  aspect, but `Cardano.Chain.Txp` for coding the UTxO related logic. I don't know if this is what we want or we want to restructure this.

crypto/src/Cardano/Crypto/Signing/Proxy/Cert.hs
```diff
@@ -56,6 +56,11 @@ safeCreateProxyCert pm ss (PublicKey delegatePk) o = coerce sig
     $ mconcat ["00", CC.unXPub delegatePk, serialize' o]
 
 -- | Checks if certificate is valid, given issuer pk, delegate pk and ω
+--
+--   We prepend the bytes "XC" because all our current certs were signed with
+--   that ByteString length prefix.
+--
+--   TODO: Make a special case for genesis and use `verifySignatureRaw` here
```
dnadales: Should we turn this TODO into an issue, so that we make sure we'll tackle it.

crypto/src/Cardano/Crypto/Signing/Proxy/Cert.hs
```diff
@@ -56,6 +56,11 @@ safeCreateProxyCert pm ss (PublicKey delegatePk) o = coerce sig
...
 verifyProxyCert
```
dnadales, line 62: Is this the right name? Is `verifyProxyCert` verifying a delegation certificate?

specs/ledger/latex/delegation.tex
```diff
@@ -17,10 +17,9 @@ \section{Delegation}
 \item Only genesis keys can delegate.
 \item Certificates must be properly signed by the delegator.
 \item Any given key can delegate at most once per-epoch.
+\item Any given key can issue at most one certificate in a given slot.
```
dnadales, line 20: This reads much better now. Thanks!

src/Cardano/Chain/Block/Block.hs
```diff
@@ -238,7 +237,7 @@ mkBlock
   -> Either GenesisHash Header
   -> SlotId
   -> SecretKey
-  -> ProxySKBlockInfo
+  -> Maybe Delegation.Certificate
```
dnadales, line 241: Maybe we should add some comments on what `SecretKey` and the `Maybe Delegation.Certificate` parameters represent. My guess is that the former is the key that is signing the block, and the latter a delegation certificate in case the key is not a genesis one. Is that correct?

src/Cardano/Chain/Block/Header.hs
```diff
@@ -213,30 +214,29 @@ mkHeaderExplicit
...
-        pske
-  leaderPk  = maybe (toPublic sk) snd pske
+  proof     = mkProof body
+
+  toSign    = ToSign prevHash proof slotId difficulty extra
+
+  signature = case mDlgCert of
+    Nothing -> BlockSignature $ sign pm SignMainBlock sk toSign
+    Just dlgCert ->
+      BlockPSignatureHeavy $ proxySign pm SignMainBlockHeavy sk dlgCert toSign
```
dnadales, line 242: I guess `BlockPSignatureHeavy` is a block signature issued by the delegate of a genesis key. Is this a good name? If not we could make an issue to rename this.

src/Cardano/Chain/Block/Header.hs
```diff
@@ -393,7 +393,7 @@ wrapBoundaryBytes = mappend "\130\NUL"
 --   interval).
 data BlockSignature
   = BlockSignature (Signature ToSign)
-  | BlockPSignatureHeavy (ProxySigHeavy ToSign)
+  | BlockPSignatureHeavy (ProxySignature EpochIndex ToSign)
```
dnadales, line 402: Does this mean that we're signing the epoch index as well now? I guess not since we wouldn't be able to validate the existing chain otherwise. Could you add a comment that explains this? :grin:

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -113,25 +125,26 @@ updateSigningHistory pk sh
...
-      . map (, 0)
-      . M.keys
-      . getGenesisWStakeholders
-      $ configBootStakeholders config
-    , shSigningQueue      = Empty
-    }
-  , cvsPreviousHash    = Nothing
-  , cvsDelegationState = DelegationState
-  }
+  , cvsDelegationState :: InterfaceState
```
dnadales: I don't know whether `InterfaceState` is the most representative name for the chain-ledger interface state. What about something like `ChainLedgerInterfaceState`? Too long? Maybe `ChainLedgerIState`? Too Java? :) Anyway, just a suggestion.

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -190,11 +204,10 @@ updateChainBoundary config cvs bvd = do
 updateChain
```
dnadales, line 217: Does it make sense to include a comment here about the fact that this function is implementing a transition rule from the spec? Same for the other functions that implement transitions.

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -144,6 +157,7 @@ data ChainValidationError
   | ChainValidationInvalidHash HeaderHash HeaderHash
   | ChainValidationInvalidSignature BlockSignature
   | ChainValidationMissingDelegator PublicKey
+  | ChainValidationSchedulingError SchedulingError
```
dnadales: It's the comments guy again: maybe add some comments explaining what this error means, since it might not be immediately obvious.

src/Cardano/Chain/Delegation/Certificate.hs
```diff
@@ -0,0 +1,44 @@
...
+  (FromJSON(..), Int54, JSValue(..), ToJSON(..), fromJSField, mkObject)
+
+import Cardano.Chain.Slotting (EpochIndex)
+import Cardano.Crypto (AProxySecretKey(..), pskOmega, unsafeProxySecretKey)
+
+
+-- | A delegation certificate is a `ProxySecretKey` tagged with an `EpochIndex`
+type Certificate = ACertificate ()
+
+type ACertificate a = AProxySecretKey EpochIndex a
```
dnadales, line 27: Why the `A` prefix?

src/Cardano/Chain/Delegation/Certificate.hs
```diff
@@ -0,0 +1,44 @@
...
+
+instance Monad m => ToJSON m Certificate where
+  toJSON psk = mkObject
+    -- omega is encoded as a number, because in genesis we always set it to 0
```
dnadales, line 31: Would it make sense to clarify here (between brackets) what `omega` is?

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -252,9 +265,21 @@ updateChain config delegates cvs b = do
   -- Update the signing history
   let signingHistory' = updateSigningHistory delegator signingHistory
 
+  -- Update the delegation state
```
dnadales, line 274: Something that I missed in the previous review: we have a lot of this `unless ... throwError` patterns. Does it make sense to write a function similar to `wrapError` (say `wrapError_`)?

src/Cardano/Chain/Delegation/Payload.hs
```diff
@@ -32,23 +32,23 @@ import Cardano.Binary.Class
...
+  { getPayload    :: [ACertificate a]
   , getAnnotation :: a
   } deriving (Show, Eq, Generic, Functor)
 
 instance (NFData a) => NFData (APayload a) where
 
 type Payload = APayload ()
 
-unsafePayload :: [ProxySKHeavy] -> Payload
+unsafePayload :: [Certificate] -> Payload
```
dnadales: If the `Certificate` here represents a delegation certificate maybe we'd want to make this type more specific.

src/Cardano/Chain/Delegation/Validation.hs
```diff
@@ -0,0 +1,269 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE TupleSections    #-}
+
+-- | Validation and update of the delegation ledger state.
```
dnadales: Are we validating the ledger state or the delegation certificates?

src/Cardano/Chain/Delegation/Validation.hs
```diff
@@ -0,0 +1,269 @@
...
+  -> SchedulingState
+  -> ACertificate ByteString
+  -> m SchedulingState
+scheduleCertificate config slot d ss cert = do
+
+  -- Check that the delegator is a genesis key
+  unless
+    (delegator `M.member` genesisStakeholders)
+    (throwError $ SchedulingNonGenesisDelegator delegator)
+
```
dnadales, line 112: Regarding this `unless ... throwError` pattern maybe we can write:

```haskell
  (delegator `M.member` genesisStakeholders) `orThrow` SchedulingNonGenesisDelegator delegator

-- ...
orThrow :: MonadError e m => Bool -> e -> m a

```

src/Cardano/Chain/Delegation/Validation.hs
```diff
@@ -0,0 +1,269 @@
...
+  -- Check that the delegation epoch is greater than or equal to the current one
+  unless
+    (epoch <= delegationEpoch)
+    (throwError $ SchedulingPastEpoch epoch delegationEpoch)
+
+  -- Check that the delegator hasn't already delegated in 'delegationEpoch'
+  when
```
dnadales: Any reason why we're using `when` and checking the negation of the condition in the rules?

src/Cardano/Chain/Delegation/Validation.hs
```diff
@@ -0,0 +1,269 @@
...
+
+
+--------------------------------------------------------------------------------
+-- Activation
+--------------------------------------------------------------------------------
+
+-- | Maps containing, for each delegator, the active delegation and the slot it
+--   became active in.
+data ActivationState = ActivationState
+  { asDelegationMap   :: Map StakeholderId StakeholderId
```
dnadales, line 163: A small deviation here (and I think somewhere else in this PR) is that we don't distinguish between genesis and non-genesis keys. I guess making that distinction in the code will be quite difficult ...

src/Cardano/Chain/Delegation/Validation.hs
```diff
@@ -0,0 +1,269 @@
...
+  { isSchedulingState :: SchedulingState
+  , isActivationState :: ActivationState
+  } deriving (Eq, Show)
+
+
+-- | The initial state maps each genesis key to itself and overrides this using
+--   certificates from the genesis block.
+initialInterfaceState
+  :: MonadError SchedulingError m => Genesis.Config -> m InterfaceState
+initialInterfaceState config = updateDelegation config 0 0 is certificates
```
dnadales, line 202: So here we're applying the transition rules to get the initial interface state. Since the spec is not prescriptive about what the initial state should be I guess this is totally ok.

cardano-chain.cabal
ruhatch: Yeah, we could change `Txp` (Tx Processing I believe it stands for) to `UTxO`

ruhatch: Also I'd like to move them to `Cardano.Ledger` at some point ;)

crypto/src/Cardano/Crypto/Signing/Proxy/Cert.hs
ruhatch: I've discussed this with Duncan and we're going to leave it as is for now, so I will update this. We will remove this kind of cert entirely after BFT, so we can rely on this canonical encoding until then.

src/Cardano/Chain/Delegation/Validation.hs
```diff
@@ -0,0 +1,269 @@
...
+  -> m InterfaceState
+updateDelegation config slot d is certificates = do
+
+  -- Schedule new certificates
+  ss <- foldM
+    (scheduleCertificate config slot d)
+    (isSchedulingState is)
+    certificates
+
+  let SchedulingState delegations keyEpochs = ss
```
dnadales: We don't seem to need `ss` anywhere else. Would it make sense to write:

```haskell
SchedulingState delegations keyEpochs <- foldM ...
```

crypto/src/Cardano/Crypto/Signing/Proxy/Cert.hs
ruhatch: A `Delegation.Certificate` comes in the form of a `ProxySecretKey` that is then used for proxy signing. That contains a `ProxyCert` which validates the whole thing. This is potentially confusing and we should definitely think more about it. I'll ask Duncan about what crypto we'll actually need and maybe we can simplify this.

src/Cardano/Chain/Block/Block.hs
ruhatch: Yes, I will add some comments for this.

src/Cardano/Chain/Delegation/Validation.hs
```diff
@@ -0,0 +1,269 @@
...
+      activateDelegation
+      (isActivationState is)
+      (Seq.filter ((<= slot) . sdSlot) delegations)
+
+  -- Remove stale values from 'SchedulingState'
+  let
+    inWindow s = subSlotNumber d slot <= s && s <= addSlotNumber d slot
+    epoch = slotNumberEpoch (configEpochSlots config) slot
+
+    ss'   = SchedulingState
```
dnadales, line 273: Spacing?

src/Cardano/Chain/Delegation/Validation.hs
```diff
@@ -0,0 +1,269 @@
...
+-- | Update the 'InterfaceState' with a list of new 'Certificate's
+updateDelegation
+  :: MonadError SchedulingError m
+  => Genesis.Config
+  -> FlatSlotId
+  -> SlotCount
+  -> InterfaceState
+  -> [ACertificate ByteString]
+  -> m InterfaceState
+updateDelegation config slot d is certificates = do
```
dnadales, line 253: This is just amazing:

from-spec-to-haskell

dnadales (approved):
Looks great! Thanks for this Ru!

cardano-chain.cabal
dnadales: Cool. If you want make some issues under the technical debt epic, so that we don't forget :)

src/Cardano/Chain/Block/Header.hs
ruhatch: I guess the `P` is for `Proxy` and the `Heavy` refers to `Heavyweight`, but we don't make that distinction anymore. For this implementation `BlockProxySignature` might be fine? I've made an issue for this.

src/Cardano/Chain/Block/Header.hs
ruhatch: We always signed the `EpochIndex` as that is the `omega` thing. It was just hidden away behind a `ProxySigHeavy` type before.

src/Cardano/Chain/Block/Validation.hs
ruhatch: I think `InterfaceState` and then using it as `Delegation.InterfaceState` could be clear enough?

src/Cardano/Chain/Block/Validation.hs
ruhatch: Yes, I will do that!

src/Cardano/Chain/Block/Validation.hs
ruhatch: :+1:

src/Cardano/Chain/Delegation/Certificate.hs
ruhatch: It's annotated with data used for validation (in this case we need the serialized bytes of the `EpochIndex`)

src/Cardano/Chain/Block/Validation.hs
ruhatch: How about `orThrowError`, which would give something like
```
epochBlah < epochBlah'
  `orThrowError` EpochError epochBlah epochBlah'
```
which reads kind of nicely

src/Cardano/Chain/Delegation/Payload.hs
ruhatch: There's a nice pattern to use simple names and use `qualified` imports to add context. So here I could use `Delegation.Certificate`.

src/Cardano/Chain/Delegation/Validation.hs
ruhatch: Aha, you found the same thing! Haha

ruhatch: Any idea for `when ... throwError`?

src/Cardano/Chain/Delegation/Validation.hs
ruhatch: Yeah, we could stick to `unless` and then we can use `orThrow` everywhere

src/Cardano/Chain/Delegation/Validation.hs
ruhatch: Yeah, there isn't really a good way to get this at the type level. We could create a refined type that lifts the `StakeholderId`s to a type level list and values of the type have to be in the list, but really we should just start the Idris rewrite then...

src/Cardano/Chain/Delegation/Validation.hs
ruhatch: Yeah, so we imagine that the initial state is the same as the spec and then there's a special genesis update :man_shrugging:

src/Cardano/Chain/Delegation/Validation.hs
ruhatch: Yup!

src/Cardano/Chain/Delegation/Validation.hs
ruhatch: Formatting dictator likes this one for some reason ;)

src/Cardano/Chain/Delegation/Validation.hs
ruhatch: :tada:

src/Cardano/Chain/Block/Validation.hs
dnadales: Yes!

src/Cardano/Chain/Delegation/Validation.hs
dnadales: Don't use `when`? :)

src/Cardano/Chain/Delegation/Validation.hs
dnadales: Hail him/her then :P

PR #190, opened 2018-11-30
[DEVOPS-1033] include the gitrev in a dummy derivation

cleverca22:
this forces hydra to create a new eval for every rev, even if nothing has changed


ruhatch (approved):
Sneaky!

PR #199, opened 2018-12-03
Replace unless/when ... throwError with orThrowError

ruhatch:
The `unless condition (throwError Error)` pattern arises a lot in our code, and this replaces instances of it with ``condition `orThrowError` Error``. It also changes some `when inverseCondition` to use it too.

### DIFF
--- a/crypto/src/Cardano/Crypto/HD.hs
+++ b/crypto/src/Cardano/Crypto/HD.hs
@@ -192,15 +192,14 @@ decryptChaChaPoly
 decryptChaChaPoly nonce key header encDataWithTag = do
   let tagSize = 16 :: Int
   let l       = B.length encDataWithTag
-  unless (l >= tagSize)
-    $  Left
-    $  "Length of encrypted text must be at least "
-    <> show tagSize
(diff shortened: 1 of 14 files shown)


nc6:
Personally I find this _slightly_ harder to read, since the `unless` front-loads the context that this is a condition, rather than needing to work backwards from an infix. In general, I would avoid infixing things unless the entire "clause" can be read/understood in one pass.

Having said that, if others find this more readable, I'll happily go with the majority opinion.

dnadales (approved):
Unlike Nick I prefer the new operator `orThrowError` over the `unless ... throwError` patterns everywhere. But as Nick said, I'm fine with whatever the majority prefers :)

ruhatch:
I agree it might be harder to read at first. However, after reading a whole bunch of rules, and seeing the construction a few times I think it starts to pay off. To me it started to read like English, e.g. current epoch is less than delegation epoch or throw error ... Then it starts to read like a list of constraints!   How do we make a decision here?

ruhatch:
I'm going to merge and we can always revert as it didn't take very long!

PR #202, opened 2018-12-04
Delegation types for the ledger

nc6:
This PR adds the types needed for implementing the delegation rules, generators and interpreters.

I've also done some more refactoring of the ledger code, hopefully to make things simpler.


dnadales (approved):
Looks good. Thanks!

specs/ledger/hs/src/delegation/Ledger/Delegation.hs
```diff
@@ -0,0 +1,61 @@
...
+
+-- | A genesis key is a specialisation of a generic VKey.
+newtype VKeyGen = VKeyGen VKey
+
+
+data DCert = DCert
+  { -- | Body of the delegation certificate
+    _dbody :: (VKey, Epoch)
+    -- | Witness for the delegation cerfiticate
+  , _dwit :: (VKeyGen, Sig VKey)
```
mdimjasevic: Shouldn't this be of type `(VKeyGen, Sig VKeyGen)` instead?

nc6: I think given this particular encoding of `Sig`, this should just be `Sig VKeyGen`

nc6:
```suggestion
  , _dwit :: Sig VKeyGen
```

specs/ledger/hs/src/delegation/Ledger/Delegation.hs
```diff
@@ -0,0 +1,61 @@
...
+data DSState = DSState
+  { _dssScheduledDelegations :: [(Slot, VKeyGen, VKey)]
+  , _dssKeyEpochDelegations :: Set (Epoch, VKeyGen)
+  }
+
+makeLenses ''DSState
+
+-- | Delegation state
+data DState = DState
+  { _dsDelegationMap :: Map VKeyGen VKey
```
mdimjasevic, line 56: @nc6, @dnadales : can we make this to have a type `Map VKey VKeyGen` instead? It would make things much easier on the blockchain end. In particular, there I need to check for a given delegatee if it has signing rights, i.e., if somebody delegated to them. For a map of type `Mak k a` and a value of type `a`, I am not aware of a function from the `containers` library that would return a `Maybe k`. Of course, I can write such a function myself, but it will not be computationally efficient because it will have the time complexity of `O(n)` where `n` is the size of the map.

dnadales: Fine by me (but I haven't worked on the implementation, so @nc6 might have a different opinion). If we invert the key-values in the map I think we should add property-tests that express the relationship between the two abstractions.

PR #204, opened 2018-12-05
Add exported library to allow cs-blockchain to depend on cs-ledger.

nc6:


specs/ledger/hs/src/delegation/Ledger/Delegation.hs
```diff
@@ -59,3 +59,7 @@ data DState = DState
   }
 
 makeLenses ''DState
+
+--------------------------------------------------------------------------------
+-- Transition systems
+--------------------------------------------------------------------------------
```
nc6, line 65: This is here by accident, whoops. But it's only a comment.

mdimjasevic (approved):
Thanks for adding this!

PR #205, opened 2018-12-05
Incorporate additional feedback on the update-mechanism rules

### LINKED ISSUE #160: Model software version in the ledger rules
Update the ledger rules to model these aspects:

- The ledger state should contain the latest `(application-name, application-version)` pairs seen on the chain.
-  Whenever a proposal gets confirmed, we update the `(appication-name, application-version)` pairs with those of the proposal. This means that confirmations allow immediate software updates, unlike protocol updates that occur at epoch boundaries and require a confirmation depth of 2k slots.
- We need to keep track of when a software update was confirmed, so we need a map: `application-name |-> (software-version, confirmation-slot)`.
- Check all the `(application-name, application-version)` pairs that come in an update proposal:
    - The application upgrades should increase the application version (in principle we could be quite lax on how this increase looks like (any lexicographic increase might be allowed)).
- Keep a mapping from application-name to its latest version.

- Additionally: think if the protocol update could be segregated from the software update.

---
# Consequences

# Retrospective

### LINKED ISSUE #172: Determine cleanup and/or timeout mechanisms for update proposals
We should model how and when the information of update-proposals gets cleaned up from the ledger state.

The current version of the rules:
image
cleans up everything (including confirmed proposals) at a fixed timeout value `u`. This might not be desired.

Confirmed proposals could be cleaned up at an epoch boundary, when new proposals get adopted. However, this poses a new question regarding the usefulness of the timeout parameter (`u` above). Couldn't we simply cleanup all the information related with an update proposal at the end of an epoch, when we also check if a new proposal can be adopted? If at an epoch boundary, we'd keep only the confirmed proposals that have version numbers greater than the newly adopted proposal, wouldn't that be enough for bounding the ledger state size?

---
# Consequences

# Retrospective

### LINKED ISSUE #173: Make state diagrams for protocol-update proposals and software-update proposals
The life cycles of protocol-update proposals and software-update proposal differ: protocol-update proposals need to be endorsed by enough nodes to become adopted, and they can also be discarded after being confirmed. On the other hand, the software in a software-update proposal can be updated as soon as it gets confirmed.

Make state diagrams for life-cycle of the two types of proposals, and explain the different states and what triggers the transitions between them. The states include: active, rejected, confirmed, adopted. 

---
# Consequences

# Retrospective

dnadales:
- Model software version in the ledger rules. Software version is now part of the ledger state.
- Revise the cleanup mechanism for update proposals. Confirmed proposals are not cleaned when the voting period ends, they are removed from the set of confirmed proposals when a new protocol version that has higher version gets adopted.
- Make state diagrams to highlight the differences between protocol updates and software updates.

Closes #160.
Closes #172.
Closes #173.

---
# Consequences

# Retrospective

[nc6] We should think about the best way to review such things, as it's quite dense to "intuit" a whole such system at once!

### DIFF
--- a/specs/ledger/latex/blockchain-interface.tex
+++ b/specs/ledger/latex/blockchain-interface.tex
@@ -159,6 +159,16 @@
 \subsection{Update-proposals interface}
 \label{sec:update-proposals-interface}
 
+Figure~\ref{fig:ts-types:upi} defines the types of the transition systems
+related with the update-proposals interface. The acronyms in the transition
+labels have the following meaning:
+\begin{description}
+\item[UPIREG] Update-proposal registration.
+\item[UPIEND] Update-proposal vote.
+\item[UPIEND] Update-proposal endorsement.
+\item[UPIEC] Update-proposal epoch-change.
+\end{description}
+
 \begin{figure}[htb]
   \emph{Update-proposals interface environments}
   \begin{align*}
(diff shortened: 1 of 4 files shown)


nc6 (approved):
Looks pretty good to me. I haven't exhaustively run through all of the transitions (figure we'll end up doing that during translation to Haskell), but everything looks pretty sensible!

PR #215, opened 2018-12-11
STS Systems for delegation

nc6:
This can be reviewed commit by commit. The first two commits are related to a rewrite of the STS system (and I will open a separate PR for them).

Most of the work here is in the final commit, which adds delegation system types. I still need to add delegation interface types.

### DIFF
--- a/specs/ledger/hs/src/delegation/Ledger/Delegation.hs
+++ b/specs/ledger/hs/src/delegation/Ledger/Delegation.hs
@@ -1,16 +1,21 @@
 {-# LANGUAGE FlexibleInstances      #-}
 {-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeApplications          #-}
 {-# LANGUAGE TypeFamilies          #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 module Ledger.Delegation where
 
 import Control.Lens
+import Control.State.Transition
 import Ledger.Signatures
 import Ledger.Core
 import Numeric.Natural (Natural)
+import qualified Data.List as List
 import Data.Set (Set)
+import qualified Data.Set as Set
 import Data.Map.Strict (Map)
+import qualified Data.Map.Strict as Map
 
 --------------------------------------------------------------------------------
 -- Abstract types
@@ -58,9 +63,9 @@ makeFields ''DSState
 
 -- | Delegation state
 data DState = DState
-  { _dSStateDelegationMap :: Map VKeyGen VKey
+  { _dStateDelegationMap :: Map VKeyGen VKey
     -- | When was the last time each genesis key delegated.
-  , _dSStateLastDelegation :: Map VKeyGen Slot
+  , _dStateLastDelegation :: Map VKeyGen Slot
   }
 
 makeFields ''DState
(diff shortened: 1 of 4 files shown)


specs/semantics/hs/src/Control/State/Transition.hs
```diff
@@ -1,20 +1,64 @@
...
+{-# LANGUAGE TypeFamilyDependencies          #-}
 {-# LANGUAGE UndecidableInstances  #-}
 
 -- | Small step state transition systems.
 module Control.State.Transition where
 
 import Control.Lens
-import Data.Maybe (catMaybes)
+import Control.Monad (unless)
+import Control.Monad.Free.Church
```
dnadales: Maybe import these (`Lens` and `Church`) qualified, so that we know where `F` comes from :)

nc6:
#216 should be merged first.

mdimjasevic (changes requested):
Can you please rebase on top of the current `master`? It looks to me like you haven't done so yet.

nc6:
@mdimjasevic Done, although bors would have handled that.

mdimjasevic:
> @mdimjasevic Done, although bors would have handled that.

I know, but it removes things from the PR that are already in the target branch, which makes reviewing a lot easier.

specs/ledger/hs/src/delegation/Ledger/Delegation.hs
```diff
@@ -84,6 +82,202 @@ data DIState = DIState
...
+  initialRules = []
+  transitionRules =
+    [ do
+        TRC (env, st, sig) <- judgmentContext
+        case sig of
+          [] -> return st
+          (x:xs) -> do
+            dss' <- trans @SDELEGS $ TRC (env, st, xs)
+            dss'' <- trans @SDELEG $ TRC (env, dss', x)
+            return dss'
```
mdimjasevic, line 214: I haven't thought about this much, but I just wanted to check if you meant to return `dss''` (double prime) by chance. If not, I'd suggest not using a name for that last binding in this `do` block.

specs/ledger/hs/src/delegation/Ledger/Delegation.hs
```diff
@@ -84,6 +82,202 @@ data DIState = DIState
...
+  type State DELEG = DIState
+  type Signal DELEG = [DCert]
+  type Environment DELEG = DIEnv
+
+  data PredicateFailure DELEG
+    = SDelegSFailure (PredicateFailure SDELEGS)
+    | ADelegSFailure (PredicateFailure ADELEGS)
+    deriving (Eq, Show)
+
+  initialRules = []
```
mdimjasevic, line 260: Shouldn't there be at least one initial rule? There has to be an initial state. I am not completely sure about this, but I guess somewhere an initial state has to be created, based on the genesis block's verification keys. Am I right? I envisioned it would happen in a function like this: https://github.com/input-output-hk/cardano-chain/blob/mdimjasevic/193-tying-blockchain-ledger/specs/chain/hs/src/Delegation/Interface.hs#L40

mdimjasevic (approved):

mdimjasevic:
bors try

specs/ledger/hs/src/delegation/Ledger/Delegation.hs
nc6: At the moment I'm just reflecting the LaTeX spec. But yes, I wonder also whether this is missing.

PR #217, opened 2018-12-12
Fixates the number of epoch files for the CI to at most 10

mdimjasevic:
This patch tries to work around the issue of excessive memory usage during continuous integration. It fixates the number of epoch files for the CI to at most 10.


mdimjasevic:
bors try

nc6 (approved):
Great! Let's merge this at least temporarily to get CI working.

Issue #193, opened 2018-12-03
Tie blockchain layer to ledger layer (delegation)


nc6:
This should tie the blockchain and ledger layers together (limited to carrying a delegation payload) - e.g. we should have an STS implementation of the blockchain including delegation

---
# Consequences

# Retrospective

There is an inference rule or rules missing specifying initial states for the delegation interface. An issue for that is #223.

nc6:
Depends on #191

PR #221, opened 2018-12-14
[193] Ties blockchain layer to ledger layer (delegation)

mdimjasevic:
This patch ties together the blockchain layer and the ledger layer specifications via the delegation interface.

This closes issue #193.


mdimjasevic:
bors try

specs/chain/hs/src/Chain/Blockchain.hs
```diff
@@ -90,42 +65,41 @@ trimIx m (MkK k) ix = foldl (flip f) m (Map.keysSet m)
...
 
 -- | Environment for blockchain rules
 data BlockchainEnv = MkBlockChainEnv
   {
-    bcEnvPp :: ProtParams
-  , bcEnvSl :: Slot
-  , bcEnvK :: K
-  , bcEnvT :: T
+    bcEnvPp    :: ProtParams
+  , bcEnvDIEnv :: DSEnv
```
nc6: Might want to use the `DIEnv` alias here, just in case things change

specs/chain/hs/src/Delegation/Interface.hs
```diff
@@ -1,47 +1,45 @@
...
-data DSIState
-
--- | For a given delegation state, it returns a mapping from a delegatee key
--- to a delegator key. If a key is not present in the key set of the returned
+-- | For a given delegation state, it returns a mapping from a delegator key
+-- to a delegatee key. If a key is not present in the value set of the returned
 -- map, it has no right to sign a block in the current slot
-delegates :: DSIState -> Map.Map VKey VKey
-delegates = undefined
+delegates :: DIState -> Map.Map VKeyGen VKey
```
nc6: Better to remove this function, it just adds unnecessary indirection.

specs/chain/hs/src/Delegation/Interface.hs
```diff
@@ -1,47 +1,45 @@
...
-newCertsRule :: TransitionRule Interf
-newCertsRule = undefined
-
--- | Updates the delegation interface state with a set of heavyweight
--- delegation certificates that arrived in a block issued in the given
--- slot
-updateCerts :: Slot -> Set HCert -> DSIState -> DSIState
-updateCerts = undefined
+initDIState :: DIState
+initDIState = initDIStateFromKeys initVKeys
```
nc6, line 23: The keys are probably in the environment, aren't they?

specs/chain/hs/src/Delegation/Interface.hs
```diff
@@ -1,47 +1,45 @@
...
+-- | For a given delegation state, it returns a mapping from a delegator key
+-- to a delegatee key. If a key is not present in the value set of the returned
 -- map, it has no right to sign a block in the current slot
-delegates :: DSIState -> Map.Map VKey VKey
-delegates = undefined
+delegates :: DIState -> Map.Map VKeyGen VKey
+delegates s = s ^. delegationMap
+
+-- | Returns a key from a map for a given value.
+maybeMapKeyForValue :: (Eq a, Ord k) => a -> Map.Map k a -> Maybe k
```
nc6: Reviewing this on a call, so don't have the spec up in front of me. But this seems a slightly weird thing to do - can you point to the section of the spec where this is done? We should make sure that this is reasonable.

nc6 (changes requested):
Few things that need changing, and a few questions I'd like to clarify!

specs/chain/hs/src/Delegation/Interface.hs
mdimjasevic: Done.

specs/chain/hs/src/Chain/Blockchain.hs
mdimjasevic: Done

specs/chain/hs/src/Delegation/Interface.hs
mdimjasevic: This simply follows from the ledger spec that @dnadales wrote. Take a look at `DState` (as you implemented it, and you said you simply followed the spec) and its delegation map. I did ask for the keys and values to be swapped, i.e., to make it a mapping from delegatee keys to delegator keys, but no one reacted to that. Hence this weird function you're asking about, Nicholas.

specs/chain/hs/src/Delegation/Interface.hs
mdimjasevic: The initial keys are in the genesis block, which the ledger layer is not aware of. Hence it needs to fetch that piece of information from the blockchain layer. Is that the same as the environment?

specs/chain/hs/src/Delegation/Interface.hs
nc6: Fair enough

nc6:
@mdimjasevic You've marked things as done but don't seem to have pushed the corresponding changes?

nc6 (approved):

mdimjasevic:
> @mdimjasevic You've marked things as done but don't seem to have pushed the corresponding changes?

My apologies for that. I kept them local while waiting to settle down the remaining issues and to push it at once. They're available now.

nc6:
bors retry

mdimjasevic:
> bors retry

Thank you for baby-sitting this silly issue that we currently have with builds, @nc6!

Issue #223, opened 2018-12-18
Add missing initial rules for delegation


mdimjasevic:
Momentarily, the ledger spec implementation lacks any rules regarding the initial state of delegation:

https://github.com/input-output-hk/cardano-chain/blob/d4483b89a0ec8a6346faac09d13b2dae6939bab6/specs/ledger/hs/src/delegation/Ledger/Delegation.hs#L260

As far as I can tell, they're missing from the LaTeX spec too. @dnadales, can you confirm? Without such rules, the system has no state to start from, which means it cannot run, hence no actual validation can happen. Therefore, rules specifying the initial delegation states have to be provided.

I take this also needs to be part of the delegation interface between the ledger and the blockchain. Verification keys from the genesis block will need to be there.

# Retrospective

* A lot of time was lost in waiting to agree on initial states, which was amplified by the next point,
* Berlin workshop postponed this task for a week,
* Some refactoring was done that we decided to discard in the end.

mdimjasevic:
I started working on this in the following branch: [cardano-chain/mdimjasevic/223-initial-delegation-rules](https://github.com/input-output-hk/cardano-chain/tree/mdimjasevic/223-initial-delegation-rules).

PR #228, opened 2018-12-21
[223] Add initial delegation rules

mdimjasevic:
This PR adds initial delegation rules.

To enable actual running of all transition systems, several modules were pulled out into a new package `cs-core`. Then both `cs-ledger` and `cs-blockchain` depend on it.

This patch closes issue #223.


specs/chain/hs/src/Chain/Blockchain.hs
```diff
@@ -17,11 +16,11 @@ import Data.ByteString.Lazy.Char8 (pack)
 
 import Chain.GenesisBlock (genesisBlock)
 import Control.State.Transition
-import Data.Maybe (fromJust, listToMaybe)
+import Data.Maybe (isJust, fromJust, listToMaybe)
```
nc6: Not needed?

specs/ledger/hs/src/Ledger/Delegation.hs
```diff
@@ -257,7 +268,16 @@ instance STS DELEG where
...
-  initialRules = []
+  initialRules = [ do
+                     IRC env <- judgmentContext
+                     return DIState
+                       { _dIStateDelegationMap        =
+                           Map.fromSet (\(VKeyGen k) -> k) (env ^. allowedDelegators)
+                       , _dIStateLastDelegation       = Map.empty
+                       , _dIStateScheduledDelegations = []
+                       , _dIStateKeyEpochDelegations  = Set.empty
+                       }
```
nc6, line 391: Yep

specs/chain/hs/src/Delegation/Interface.hs
```diff
@@ -8,15 +8,30 @@ where
...
+  , _dIStateKeyEpochDelegations
+  )
 import Types
 
 
 -- | Computes an initial delegation interface state from a set of
 -- verification keys
 initDIStateFromKeys :: Set VKeyGen -> DIState
-initDIStateFromKeys certs = undefined
+initDIStateFromKeys keys = DIState
```
nc6: This whole module should not be needed now, I think?

nc6 (commented):
Changes look good in general, have marked a couple of areas which could be changed.

dnadales:
One question for ourselves: do we want to restrict the initial environment as well? For instance, in the case of the delegation environment, if there are no allowed delegators, then the generators will not be able to generate any trace (since a delegation certificate requires at least one allowed delegator). What to do in this case?

specs/chain/hs/src/Chain/Blockchain.hs
mdimjasevic: Are you referring to all the three imported functions? Each of them is used in the module.

specs/ledger/hs/src/Ledger/Delegation.hs
mdimjasevic: Are you saying this is a good implementation?

specs/chain/hs/src/Delegation/Interface.hs
mdimjasevic: I believe you're right. I removed the module.

mdimjasevic:
> One question for ourselves: do we want to restrict the initial environment as well? For instance, in the case of the delegation environment, if there are no allowed delegators, then the generators will not be able to generate any trace (since a delegation certificate requires at least one allowed delegator). What to do in this case?

Can you please provide some further explanation? I am not sure I understand what in specific you're talking about.

dnadales:
> Can you please provide some further explanation? I am not sure I understand what in specific you're talking about.

The initial rules could restrict the environment. I gave an example of this above. Consider if the set of valid delegators is empty. Do we want that?

mdimjasevic:
> The initial rules could restrict the environment. I gave an example of this above. Consider if the set of valid delegators is empty. Do we want that?

How about using the [Data.NonEmpty.Map](http://hackage.haskell.org/package/non-empty-0.3.0.1/docs/Data-NonEmpty-Map.html) module?

dnadales:
> How about using the Data.NonEmpty.Map module?

Sure. But I was aiming at what do we want in the specs :)

specs/ledger/hs/src/Ledger/Delegation.hs
```diff
@@ -257,7 +268,16 @@ instance STS DELEG where
     | ADelegSFailure (PredicateFailure ADELEGS)
     deriving (Eq, Show)
 
-  initialRules = []
+  initialRules = [ do
```
dnadales, line 382: It seems the initial state rule for the delegation interface doesn't use any of the other initial rules  (in `SDELEG`(`S`) or `ADELEG`(`S`)), so the question is: do we need the other initial rules? Can't we just specify the initial rules for `DELEG`?

I'm just trying to remove what we might not need.

mdimjasevic: To admit, I wasn't following closely how are all `*State` states related, mostly because, as I mentioned it earlier, I find their names rather confusing and hard to read. Anyhow, now by looking at it, am I right to say that `DIState` is actually made of a `DSState` and a `DState`? If so, I'd suggest to reflect that in the `DIState`'s data constructor and then we can update these initial states accordingly.

dnadales: Yes, that could also work.

nc6:
Why the last change? `cs-blockchain` depends on `cs-ledger`, so what's wrong with having the relevant types there?

mdimjasevic:
> Why the last change? `cs-blockchain` depends on `cs-ledger`, so what's wrong with having the relevant types there?

This is to enable actual running (e.g., testing) eventually. With the way code was organised, there would be a circular dependency because keys from the genesis block are needed for the initial state of the `ADELEG` STS, yet as you said `cs-blockchain` already depends on `cs-ledger`).

mdimjasevic:
> > Why the last change? `cs-blockchain` depends on `cs-ledger`, so what's wrong with having the relevant types there?
> 
> This is to enable actual running (e.g., testing) eventually. With the way code was organised, there would be a circular dependency because keys from the genesis block are needed for the initial state of the `ADELEG` STS, yet as you said `cs-blockchain` already depends on `cs-ledger`).

@nc6 , do you think this is not needed? On a second thought, I guess so long as no attempt to actually run `cs-ledger`'s STSes is made from within `cs-ledger`, there is no need for this separation into `cs-core`. If you think we won't ever run it from there, I'll undo this separation.

mdimjasevic:
@dnadales , @nc6: I'm done with implementing initial delegation rules. Please review again.

ruhatch (commented):
This seems sensible, but I want more eyes on it as I'm not as familiar as others. The cs-core package is still here, but you said in a comment above you wanted to remove it, is that still the case?

mdimjasevic:
> This seems sensible, but I want more eyes on it as I'm not as familiar as others. The cs-core package is still here, but you said in a comment above you wanted to remove it, is that still the case?

I'm waiting on Nicholas to comment, as he had concerns with introducing `cs-core`.

nc6 (changes requested):
I would like to revert the cs-core stuff; I don't feel it adds anything and it means one more place people have to look. Other than that, LGTM!

mdimjasevic:
> I would like to revert the cs-core stuff; I don't feel it adds anything and it means one more place people have to look. Other than that, LGTM!

Ok, I removed it!

nc6 (approved):

PR #229, opened 2018-12-27
Add a valid delegation signals generator

### LINKED ISSUE #194: Generator for (valid) delegation
Add generators for delegation transition rules. 

Remaining tasks:

- [x] Add parameters to control the size of the traces. Try to leverage on `hedgehog`'s `Gen` parameters.
- [x] Write property based tests for the `DELEG` traces.
- [x] Add RTS constraints to the tests.
- [x] Organize the code.
- [x] Factor out possible duplication.
- [ ] OPTIONAL: consider adding [coverage information](https://hackage.haskell.org/package/tasty-hedgehog-coverage-0.1.0.0/docs/Test-Tasty-Hedgehog-Coverage.html)
- [x] Apply hlint.
- [x] Remove warnings.
- [x] Update the README if necessary.
- [x] Rebase from `master`.

# Retrospective

See the comments in this issue! Also:

- A lot of time spent in trying to make `nix` build with the `capability` package. At this package was not necessary, since I did not use `ProgressiveGen`.
- Had to add ad-hoc tests for the STS rules (as a way of some basic sanity check).
- Had to add property based tests for the `DELEG`.
- Tests detected errors in the Haskell translation of the STS rules for delegation, which is nice. However this takes additional time, which we should take into consideration when estimating tasks related to coding STS in Haskell or their generators.
- I could not make `ProgressiveGen` work. Tried several approaches. Had lengthy (but quite insightful) discussions with @nc6. At the end we settle for the "generate-then-validate" approach, in which we would generate signals and then validate them using the STS rules (hoping that the signal generators won't need much knowledge of the rules).
- :clock1: Had to spend quite some time trying learn how to use `ghcid` and `nix` in a workflow that resembles the one I was used to with `stack`. I'm still unsatisfied by the results.
- I spent a lot of time hammering on the `GenM` monad: shuffling the monad transformers stack, adding auxiliary functions, etc. At the end this monad was removed, so a lot of wasted effort here as well.
- Trying to workaround the [limitations in `ghci`](https://github.com/haskell/cabal/issues/5103) (see also [this issue](https://github.com/haskell/cabal/issues/5374)) is not worth the time,  specially since a `stack` workflow could be implemented. Additionally, by explaining the workaround in the README will probably take a lot of effort.

Things that we need to address:

- Classify the test cases: how do we know that we're covering all the interesting delegation cases?

dnadales:
- Add a generator for delegation traces. In a nutshell, a trace is a sequence of signal-and-ensuing-state pairs.
- Add two property based tests which check properties on the generated traces: delegation certificates in a trace are applied as expected, and duplicated scheduled delegations are rejected.
- Add two unit tests for the delegation rules: one for delegation activation, and other for delegation scheduling. These simple unit test proved to be useful to detect errors in the coding of the delegation rules.
- Add the concept of non-trivial signal, which allows to detect when errors are introduced in the encoding of the delegation rules. Otherwise, only trivial signals would be produced, and the error would remain undetected.
- Set RTS limits for the tests.
- Remove `brittany` checks on `specs` from CI: in some cases the automatic applied formatting did not produce more readable code, so I don't think we can benefit from it.
- Introduce a simple DSL for specifying expectations on transition system traces.
- Compiler warnings will be considered errors. Existing warnings in `specs` were removed.
- Remove the `Makefile` from `specs/ledger/hs`.

Close #194.


specs/ledger/hs/src/Ledger/Core.hs
```diff
@@ -19,19 +24,42 @@ class HasHash a where
...
+
+instance BA.ByteArrayAccess VKey where
+  length        = BA.length . BS.pack . show
+  withByteArray = BA.withByteArray . BS.pack . show
+
+instance HasOwner VKey where
+  owner (VKey o) = o
+
+-- | A genesis key is a specialisation of a generic VKey.
+newtype VKeyGen = VKeyGen VKey
```
mgudemann: I would prefer making `Gen` unambiguous here, e.g. `VKeyGenesis` I initially read it as abbreviation for generator.

specs/ledger/hs/src/Ledger/Core/Generator.hs
```diff
@@ -0,0 +1,41 @@
...
+epochGen = Epoch <$> natGen
+
+slotGen :: Gen Slot
+slotGen = Slot <$> natGen
+
+slotCountGen :: Gen SlotCount
+slotCountGen = SlotCount <$> natGen
+
+natGen :: Gen Natural
+natGen = integral (linear 0 100)
```
mgudemann: As I have been bitten by this, maybe parameterize this for min/max and make it a choice between linear and exponential choice?

specs/ledger/hs/src/Ledger/Delegation.hs
```diff
@@ -138,22 +249,23 @@ instance STS SDELEG where
...
-      notAlreadyDelegated st cert = not $ Set.member (cert ^. depoch, cert ^. dwho . _1) (st ^. keyEpochDelegations)
-      -- Check that there is not already a scheduled delegation from this key
+      notAlreadyDelegated st cert =
+        Set.notMember (cert ^. depoch, cert ^. dwho . _1) (st ^. keyEpochDelegations)
+      -- Check whether there is a scheduled delegation from this key
       notAlreadyScheduled :: DSEnv -> DSState -> DCert -> Bool
       notAlreadyScheduled env st cert =
-        List.elem
-          (((env ^. slot) `addSlot` (env ^. liveness)), cert ^. dwho . _1)
+        List.notElem
```
mgudemann, line 266: where is the negation coming from, was this wrong before ? Above it was a change from `not $ Set.elem ... ` to `Set.notElem ...`

specs/ledger/hs/src/Ledger/Delegation.hs
```diff
@@ -281,3 +393,52 @@ instance Embed SDELEGS DELEG where
...
+--
+-- At the moment the size of the generated list is severely constrained since
+-- the longer the list the higher the probability that it will contain
+-- conflicting delegation certificates (that will be rejected by the
+-- transition-system-rules).
+dcertsGen :: DSEnv -> Gen [DCert]
+-- NOTE: at the moment we cannot use a linear range that depends on the
+-- generator size: the problem is that the more delegation certificates in the
+-- resulting list, the higher the probability that this list will be rejected
+-- and the generator will have to retry.
```
mgudemann, line 427: should this Note be merged with the above comment?

specs/ledger/hs/test/Ledger/Delegation/Examples.hs
```diff
@@ -0,0 +1,87 @@
...
+
+      .- (s 1, (gk 1, k 11)) .-> DState [(gk 0, k 10), (gk 1, k 11)]
+                                        [(gk 0, s 0), (gk 1, s 1)]
+
+      .- (s 2, (gk 0, k 11)) .-> DState [(gk 0, k 11), (gk 1, k 11)]
+                                        [(gk 0, s 2), (gk 1, s 1)]
+
+      .- (s 3, (gk 2, k 12)) .-> DState [(gk 0, k 11), (gk 1, k 11), (gk 2, k 12)]
+                                        [(gk 0, s 2), (gk 1, s 1), (gk 2, s 3)]
+
```
mgudemann: remove newline here ?

specs/semantics/hs/src/Control/State/Transition/Trace.hs
```diff
@@ -0,0 +1,131 @@
...
+  , STS
+  , Signal
+  , State
+  , TRC(TRC)
+  , applySTS
+  , PredicateFailure
+  )
+
+-- | A successful trace of a transition system.
+--
```
mgudemann, line 44: Maybe add in this description that the sequence is reversed, i.e., the initial state is the last.

specs/ledger/hs/cs-ledger.cabal
```diff
@@ -16,18 +16,60 @@ build-type:          Simple
...
+    manual: True
+
 library
   hs-source-dirs:      src
-  default-language:    Haskell2010
-  exposed-modules:     Ledger.Signatures
-                     , Ledger.Core
-                     , Ledger.UTxO
+  exposed-modules:     Ledger.Core
+                     , Ledger.Core.Generator
```
mdimjasevic, line 27: Why is not this generator module part of testing?

specs/ledger/hs/src/Ledger/Delegation.hs
```diff
@@ -262,14 +374,14 @@ instance STS DELEG where
     [ do
         TRC (env, st, sig) <- judgmentContext
         sds <- trans @SDELEGS $ TRC (env, st ^. dIStateDSState, sig)
-        let slots = filter ((< (env ^. slot)) . fst) $ sds ^. scheduledDelegations
+        let slots = filter ((<= (env ^. slot)) . fst) $ sds ^. scheduledDelegations
```
mdimjasevic, line 376: Was this a buggy implementation, hence the change to `<=`?

specs/ledger/hs/src/Ledger/Core/Generator.hs
```diff
@@ -0,0 +1,41 @@
...
+
+vkGen :: Gen VKey
+vkGen = VKey . Owner <$> natGen
+
+vkgGen :: Gen VKeyGen
+vkgGen = VKeyGen <$> vkGen
```
mdimjasevic: Can you please use more distinguishing names? At first I thought this was a circular definition where `vkgGen` depends on itself. It took mi minutes to spot it is defined via `vkGen` instead. Descriptions would be helpful in comments too.

specs/ledger/hs/src/Ledger/Delegation.hs
```diff
@@ -281,3 +393,52 @@ instance Embed SDELEGS DELEG where
...
+dcertGen env = do
+  -- The generated delegator must be one of the genesis keys in the
+  -- environment.
+  vkS <- Gen.element $ Set.toList (env ^. allowedDelegators)
+  vkD <- vkGen
+  epo <- epochGen
+  return DCert { _dbody = (vkD, epo)
+               , _dwit = Sig vkS (owner vkS)
+               , _dwho = (vkS, vkD)
+               , _depoch = epo
```
mdimjasevic, line 414: Shouldn't you make the generated epoch somehow related to the epoch in the environment?

mdimjasevic (dismissed):
Generators should be put under the `test` directory, and not `src`.

specs/ledger/hs/cs-ledger.cabal
nc6: This whole library really pertains to testing - we are defining the spec in order to test the real implementation. And we will want to expose these generators to downstream libraries that wish to use them - for example, we have discussed having a CLI utility to generate chains to use in testing the Rust implementation.

So having this in the normal library is definitely correct :-)

specs/ledger/hs/cs-ledger.cabal
```diff
@@ -16,18 +16,60 @@ build-type:          Simple
...
+               , tasty-hedgehog
+               -- Local deps
+               , cs-ledger
+               , small-steps
+  ghc-options: -Wall
+               -Wcompat
+               -Wincomplete-record-updates
+               -Wincomplete-uni-patterns
+               -Wredundant-constraints
+               -- The 50 megabytes heap bound was determined ad-hoc.
```
nc6, line 75: It would be useful to record here the purpose of this bound

specs/ledger/hs/test/Ledger/Delegation/Properties.hs
```diff
@@ -0,0 +1,256 @@
...
+
+--------------------------------------------------------------------------------
+-- Delegation certification triggering tests
+--------------------------------------------------------------------------------
+
+-- | Delegation blocks. Simple blockchain to test delegation.
+data DBLOCK
+
+-- | A delegation block.
+data DBlock
```
mdimjasevic, line 106: What's the motivation behind introducing a new type for blocks? Are there limitations or obstacles in using the one defined in `cs-blockchain`?

specs/ledger/hs/src/Ledger/Core/Generator.hs
nc6: I wonder whether we would be better inlining this in `epochGen`, `slotGen` etc. I feel that the idea of "generator for natural numbers" is unlikely to be semantically useful.

specs/ledger/hs/src/Ledger/Delegation.hs
```diff
@@ -1,31 +1,115 @@
...
+  , SlotCount
+  , VKey
+  , VKeyGen
+  , (⨃)
+  , addSlot
+  , minusSlot
+  , owner
+  , owner
+  )
+import Ledger.Core.Generator (epochGen, slotCountGen, slotGen, vkGen, vkgGen)
```
nc6: I must admit to being a sceptic about the idea of importing everything explicitly like this! I feel it results in a lot of boilerplate and a lot of churn when things change. But each to their own!

specs/ledger/hs/src/Ledger/Delegation.hs
nc6: I think it was wrong before, which Damian helpfully discovered when testing!

specs/semantics/hs/src/Control/State/Transition/Trace.hs
mdimjasevic, line 44: Overall, this is an enormous PR. There is an "infrastructure" part like the `Generator` and `Trace` modules, there are properties, there are examples, there are actual HedheHog generators, changes to the existing STS library and what not. Please split it and sequence it into smaller self-contained chunks.

specs/ledger/hs/src/Ledger/Delegation.hs
```diff
@@ -78,10 +181,18 @@ data DIState = DIState
   , _dIStateLastDelegation :: Map VKeyGen Slot
   , _dIStateScheduledDelegations :: [(Slot, (VKeyGen, VKey))]
   , _dIStateKeyEpochDelegations :: Set (Epoch, VKeyGen)
-  }
+  } deriving (Show, Eq)
 
 makeFields ''DIState
 
+initialDIState :: DIState
```
nc6, line 187: Would it make more sense to have this as an initial rule?

specs/semantics/hs/src/Control/State/Transition.hs
```diff
@@ -35,29 +34,35 @@ data RuleType
...
+instance RuleTypeRep 'Initial where
   rTypeRep = SInitial
 
-instance RuleTypeRep Transition where
+instance RuleTypeRep 'Transition where
   rTypeRep = STransition
 
 -- | Context available to initial rules.
-newtype IRC sts = IRC (Environment sts)
+newtype IRC sts = IRC (Environment sts, State sts)
```
nc6: I don't understand this change. What can it mean to have an existing state, given this is an initial rule? And how would it be used in any case?

specs/semantics/hs/src/Control/State/Transition.hs
```diff
@@ -35,29 +34,35 @@ data RuleType
...
+
+type InitialRule sts = Rule sts 'Initial (State sts)
+type TransitionRule sts = Rule sts 'Transition (State sts)
 
-type InitialRule sts = Rule sts Initial (State sts)
-type TransitionRule sts = Rule sts Transition (State sts)
+-- | Extract the state inside the context.
+ctxState :: SRuleType rtype -> RuleContext rtype s -> State s
+ctxState SInitial (IRC (_, st)) = st
+ctxState STransition (TRC (_, st, _)) = st
```
nc6: As above, this seems wrong.

specs/semantics/hs/src/Control/State/Transition.hs
```diff
@@ -152,14 +157,24 @@ applySTSIndifferently
...
+    :: forall s' rtype'
+     . STS s'
+    => SRuleType rtype'
+    -> RuleContext rtype' s'
+    -> [(State s', [PredicateFailure s'])]
   applySTSIndifferently' SInitial env =
     map (applyRuleIndifferently env) initialRules
   applySTSIndifferently' STransition jc =
     map (applyRuleIndifferently jc) transitionRules
+  headOrState [] = (ctxState rTypeRep ctx, [])
```
nc6: This definitely seems wrong. If this fails at the initial rule, there is no state to fall back on. So an error is really the only option. And this case is only called if there are actually no rules to apply.

specs/semantics/hs/src/Control/State/Transition/Generator.hs
```diff
@@ -0,0 +1,177 @@
...
+  , lastState
+  , mkTrace
+  , traceLength
+  , traceSignals
+  )
+
+class STS s => HasTrace s where
+  initEnvGen :: Gen (Environment s)
+
+  initStGen :: Environment s -> Gen (State s)
```
nc6: Not seeing the point of this. Why does this not simply arise from applying an initial rule against the generated environment?

specs/semantics/hs/src/Control/State/Transition/Trace.hs
```diff
@@ -0,0 +1,131 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+-- | Traces of transition systems and associated operators.
+--
+-- This module also includes a minimal DEL to specify expectations on traces.
```
nc6: What's a DEL?

nc6 (changes requested):
Done a first pass review and left a few comments in particular places. I think in general this looks great. My main concern is again around the conception of initial rules and states - happy to either chat more about this in Slack or in Berlin next week!

specs/ledger/hs/cs-ledger.cabal
dnadales: :+1: 

What about:

```
               -- We set a bound here so that we're alerted of potential space
               -- leaks in our generators (or test) code.
```

specs/ledger/hs/src/Ledger/Core.hs
dnadales: Good one, I'm gonna change that.

specs/ledger/hs/src/Ledger/Core/Generator.hs
dnadales: Note that `natGen` is only used in this module (and thus not exported). I was not sure what would be good arbitrary ranges for slots, epochs, keys, etc. So I just used the same generator, `natGen`.

I would say, as we find that `integral (linear 0 100)` is not a good generator for any of the type wappers around `Natural` (e.g. `Epoch`, we could inline the generator and tweak the parameters, or replace it by `exponential` for instance). Does that make sense? @mgudemann @nc6 If it doesn't, do you have any concrete suggestion about what would be a good approach here?

specs/ledger/hs/src/Ledger/Delegation.hs
dnadales: Yes, that is a big downside. The problem is that when reading code, I usually find myself looking for the place at which a symbol is defined, and if it is imported I don't know where it is imported from. GNU tags alleviate this problem somehow, but it doesn't work for symbols imported from external libraries.

I think the ideal solution would be to have proper tag navigation support in Haskell (maybe it exists already and I don't know), but in its absence I'd rather go for the more bureaucratic approach for the sake of future maintainers.

However I'm wondering how do you deal with this situation, maybe if you have a good approach for looking up symbols we might just document that and have implicit imports only.

specs/ledger/hs/src/Ledger/Delegation.hs
dnadales: Certainly! Let's discuss how to go about this!

From the top of my head this would correspond to an inital rule of the form:

```
dms = K \mapsto K
dws = \emptyset
sds = \epsilon
eks = \emptyset
------------------
(dms
 dws
 sds
 eks)
```

But we also have to contemplate cases with initial rules where multiple initial states are possible, e.g.:

```
0 < n
-------
( n )
```

specs/ledger/hs/src/Ledger/Delegation.hs
dnadales: Before, `List.elem` was actually computing whether a delegation was scheduled, we wanted to know whether it wasn't scheduled instead. And indeed, testing helped us to discover this :)

specs/ledger/hs/src/Ledger/Delegation.hs
dnadales: Yup, we discovered this when testing.

specs/ledger/hs/src/Ledger/Delegation.hs
dnadales: Why?

specs/ledger/hs/src/Ledger/Delegation.hs
dnadales: I was thinking the note has more to do with the implementation detail, and I didn't want it showing as part of the documentation. Does it make sense?

specs/ledger/hs/test/Ledger/Delegation/Properties.hs
dnadales: These type of blocks is used to tests the delegation rules, using the transition system defined below. I didn't need the full blown `cs-blockchain` here. Besides this would introduce a circular dependency.

specs/semantics/hs/small-steps.cabal
```diff
@@ -16,13 +16,34 @@ build-type:          Simple
...
+                     , transformers >= 0.5
   hs-source-dirs:      src
   default-language:    Haskell2010
+  ghc-options: -Wall
+               -Wcompat
+               -Wincomplete-record-updates
+               -Wincomplete-uni-patterns
+               -Wredundant-constraints
+               -- The 50 megabytes heap bound was determined ad-hoc.
+               "-with-rtsopts=-K1m -M50m"
```
dnadales: I think I did something stupid here. RTS options don't have effect on libraries :/

specs/semantics/hs/src/Control/State/Transition/Trace.hs
dnadales: The original idea was to use the progressive gen's in #127 . However, as we discussed with @nc6 I couldn't use it to get generators for delegation, so all the infrastructure was needed. The work on this was done during the christmas break, so it made no sense to split this into smaller PR's since I'd have to wait till after the break.

I'll try to make smaller PR's in the future, but for this one I guess a workaround would be to consider N files a day, and pretend they come from smaller PR's :man_shrugging: :)

dnadales: My idea was not to give any information in the documentation about the order in which the signals-state pairs are added to the list. Since this field is not exported, the users of this module don't have to worry about this detail. When the trace needs to be extracted as a list, the trace order has to be given explicitly. Does this sound reasonable?

specs/semantics/hs/src/Control/State/Transition/Trace.hs
dnadales: What results from a typo in `DSL` :P Sorry about that. Changing it now.

specs/ledger/hs/src/Ledger/Delegation.hs
mdimjasevic: At the moment, the generated epoch is completely unrelated to the epoch in the environment. Will not this lead to a lot of generated test cases being uninteresting in the sense that they will not lead to a next valid state?

specs/semantics/hs/src/Control/State/Transition/Generator.hs
dnadales: My idea was that one should be able to generate initial states for rules that specify more than one initial state. For instance:

```
0 <  n
--------
( n )
```

specs/semantics/hs/src/Control/State/Transition.hs
dnadales: Yeah, might be. This was my clumsy workaround for avoiding `error`s in cases in which the initial rules where empty. I'd like to avoid the use of `error` if possible.

specs/semantics/hs/src/Control/State/Transition.hs
```diff
@@ -152,14 +157,24 @@ applySTSIndifferently
 applySTSIndifferently ctx =
   successOrFirstFailure $ applySTSIndifferently' @s @rtype rTypeRep ctx
  where
-  successOrFirstFailure xs = fromMaybe (head xs) $ find (not . null . snd) xs
```
dnadales, line 155: TODO: replace `head` by `headWithError` which states what is the error we got.

specs/semantics/hs/src/Control/State/Transition/Generator.hs
dnadales: Now I see what you meant. So I'm gonna change.

specs/ledger/hs/src/Ledger/Delegation.hs
dnadales: Indeed, good point. I'll change the generator so that a random increment is added to the epoch in the environment.

specs/semantics/hs/src/Control/State/Transition.hs
dnadales: I reverted this back, as we discussed today.

specs/ledger/hs/src/Ledger/Delegation.hs
dnadales: Alright, I just incorporated this initial state in an initial rule.

dnadales:
Thank you @mdimjasevic @mgudemann and @nc6 for taking the time to review this. I tried to address your concerns as well as I could. Let me know if you approve this version or whether there are other things I should address.

@mdimjasevic note that I added an initial rule for `DELEG` to be able the interpret the generators in the way we discussed today, however whatever rules you're writing in your PR should take precedence, so I hope this is not a problem.

specs/ledger/hs/src/Ledger/Core/Generator.hs
mgudemann: I currently have this
```
-- | Generator for a natural number between 'lower' and 'upper'.
genNatural :: Natural -> Natural -> Gen Natural
genNatural lower upper = Gen.integral $ Range.linear lower upper
```
But I'd suggest a choice like
```
genNatural lower upper = Gen.choice [ Gen.integral $ Range.linear lower upper
                                  , Gen.exponential $ Range.linear lower upper]
```

mgudemann: that isn't perfect, the suggestion at the training was to generate lots of "low" values, lots of "high" values and some in between

mgudemann (approved):

nc6 (approved):

specs/ledger/hs/src/Ledger/Core/Generator.hs
dnadales: Oh, I see. Right now I just removed `natGen`, and inlined the `Gen.integral` whenever I needed them. 

@nc6 do you think having a generator like the one @mgudemann suggests would make sense at this stage?

mdimjasevic:
Now that you changed `VKeyGen` to `VKeyGenesis` in the code, do you want to update it in the specs too? I know I used `VKeyGen` in the blockchain spec.

mdimjasevic:
I wasn't able to decipher how Github works with reviews, but I guess that my requested changes review isn't technically blocking you from proceeding with merging to `master`. You do have two approving reviews. If it is blocking you, please let me know.

dnadales:
bors try

dnadales:
> Now that you changed VKeyGen to VKeyGenesis in the code, do you want to update it in the specs too? I know I used VKeyGen in the blockchain spec.

Sure.

dnadales:
Bors r+

PR #237, opened 2019-01-14
Add ledger utxo properties and incorporate feedback

### LINKED ISSUE #219: Incorporate this feedback into the ledger spec
- [x] section 2: "integer ranges", but we use a,b \in R, did we mean N or Z?
- [x] section 5, fig 3: txin |→ txout ∈ TxIn |→ TxOut has the wrong type / wrong cardinality
- [x] section 3, fig 2: we don't define `sign` but we use it in the constraints. We should define it along with `verify`.
- [x] section 5, fig 3: "abstract function" would be better labelled as "functions on transactions" or "functions on abstract transaction type".
- [x] section 5, fig 4: reverse the order of the let bindings.
- [x] decide on whether to use Z or N for Lovelace.
  - [x] Change N to Z. We decided we'll use integers.

### LINKED ISSUE #231: Add git revision in the ledger spec
This is useful to keep track of the document version. See https://github.com/input-output-hk/fm-ledger-rules/pull/153

### LINKED ISSUE #232: Specify no-double spend property in the ledger spec
Specify the property that states that every `UTXO` trace does not contain two transactions that have an input in common.

### LINKED ISSUE #233: Specify "utxo is outputs minus inputs" property in the ledger spec
The final state of a `UTXO` trace must equal the intial UTxO state plust all the outputs in the trace, minus the inputs of the trace.

This would ensure that every input and output in the trace transactions are accounted for.

### LINKED ISSUE #234: In the UTXO rule require that the set of inputs is not empty
An non-empty set of inputs is needed in order to guarantee replay protection. This also enables us to prove the absence of double spending.

dnadales:
- Add two ledger UTxO properties: no double spending and utxo is outputs minus inputs.
- Incorporate feedback received from @dcoutts and @jcmincke, and part of the feedback received during the Berlin workshop.
- Add injectivity constraint to `txid`.
- Replace `N` by `Z` as an alias for `Lovelace` and remove `Lovelace` cap and explicit fees.
- Make protocol parameters consistent with Shelley spec.
- Add author field, list of contributors, and git revision.

Close #219, close #233, close #234, close #232, close #231.

### DIFF
--- a/specs/ledger/latex/properties.tex
+++ b/specs/ledger/latex/properties.tex
@@ -1,0 +1,158 @@
+\section{Transition System Properties}
+\label{sec:ts-properties}
+
+
+\subsection{Transition-system traces}
+\label{sec:ts-traces}
+
+This section introduces the notion of traces induced by the transition systems
+described in \cite{small_step_semantics}.
+
+\begin{definition}[Traces]
+  Given a state transition system $L=(S,T,\Sigma, R, \Gamma)$, the set of
+  traces of $L$, denoted as $\fun{traces}_L$ is defined as:
+  $$
+  \{ (e, s, \txs, s') \mid e \in \Gamma,~ s \in S,~ \txs \in \seqof{\Sigma},~ s' \in S\}
+  $$
+
+\end{definition}
+
+
+\begin{definition}[Valid traces]
+  Given a state transition system $L=(S,T,\Sigma, R, \Gamma)$, we define the
+  notion of valid traces inductively:
+
+  \begin{itemize}
+  \item For all $e \in \Gamma$, $s \in S$, $(e, s, \epsilon, s)$ is a valid
+    trace of $L$.
+
+  \item If $(e, s, \txs s')$ is a valid trace, and
+    $e \vdash s' \trans{}{\var{t}} s''$ is a valid transition according to the
+    rules of $L$, then $(e, s, \txs; t, s'')$ is also a valid trace.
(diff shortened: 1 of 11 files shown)


specs/ledger/latex/properties.tex
```diff
@@ -0,0 +1,158 @@
+\section{Transition System Properties}
+\label{sec:ts-properties}
+
+
+\subsection{Transition-system traces}
+\label{sec:ts-traces}
+
+This section introduces the notion of traces induced by the transition systems
+described in \cite{small_step_semantics}.
```
nc6, line 9: I wonder if we should move this into that document?

nc6 (approved):
LGTM. Is the intention to include the proofs of these properties as well?

specs/ledger/latex/properties.tex
dnadales: If it makes sense we should. My idea was to use this document to fine tune the notion of traces and associated notations, and then move it to the small steps doc.

JaredCorduan (approved):
Looks good to me!

dnadales:
> LGTM. Is the intention to include the proofs of these properties as well?

Good question. I don't know if anybody would care about them. The proof of the two UTxO properties would take about 5 pages, requiring a couple of lemmas. If we also want to include the proofs of the lemmas I guess we're looking at another 10 pages. And all this assuming the proofs are correct :)

Issue #196, opened 2018-12-03
Elaborator from executable spec to real chain types (delegation)


nc6:
This should create a real chain from an executable spec chain (only including delegation payload). This should carry information on the validity of said chains through to the real types, to allow these to be used as test cases.

PR #238, opened 2019-01-14
[#196] Interpret and validate delegation certs

ruhatch:


specs/ledger/hs/src/Ledger/Delegation.hs
```diff
@@ -27,7 +27,7 @@ module Ledger.Delegation
   , ADELEG
   , ADELEGS
   , initialDState
-  , DSEnv (DSEnv)
+  , DSEnv (..)
```
dnadales: Why do we prefer to export things explicitly? My argument for using explicit exports is that we don't export things we don't want inadvertently.

specs/ledger/hs/test/Main.hs
```diff
@@ -3,23 +3,24 @@ module Main
...
-            , testProperty "One delegation per-slot per-key" rejectDupSchedDelegs
-            ]
-        ]
+  tests = localOption Auto $ testGroup
+    "Ledger"
+    [ testGroup "Ledger" [testGroup "Delegation Examples" deleg]
+    , testGroup
+      "Properties"
+      [ testProperty "Activation"                      dcertsAreTriggered
+      , testProperty "One delegation per-slot per-key" rejectDupSchedDelegs
```
dnadales, line 24: One issue I see with this alignment is that we'll likely have to change two more lines if we add a third one :)

dnadales (approved):
LGTM.

PR #239, opened 2019-01-14
Updates to chain specification

nc6:
This branch will contain some updates to the chain specification, and should be reviewed commit-by-commit. I'll keep adding to this branch with a few additional updates I am planning to make over the next few days.

### DIFF
--- a/specs/chain/latex/Makefile
+++ b/specs/chain/latex/Makefile
@@ -36,7 +36,7 @@ $(DOCNAME).pdf: $(DOCNAME).tex
 	latexmk -pdf -pdflatex="pdflatex -interaction=nonstopmode" -use-make $(DOCNAME).tex
 
 watch: $(DOCNAME).tex
-	latexmk -pvc -pdf -pdflatex="pdflatex -interaction=nonstopmode" -use-make $(DOCNAME).tex
+	latexmk -pvc -pdf -pdflatex="pdflatex -interaction=nonstopmode -synctex=1" -use-make $(DOCNAME).tex
 
 clean:
 	latexmk -CA
--- a/specs/chain/latex/blockchain-spec.tex
+++ b/specs/chain/latex/blockchain-spec.tex
@@ -67,7 +67,6 @@
 \newcommand{\Slot}{\type{Slot}}
 \newcommand{\BlockIx}{\type{BlockIx}}
 \newcommand{\GBlock}{\type{GBlock}}
-\newcommand{\RBlock}{\type{RBlock}}
 \newcommand{\Block}{\type{Block}}
 \newcommand{\DCert}{\type{DCert}}
 \newcommand{\Queue}{\type{Q}}
@@ -100,20 +99,20 @@
 \newcommand{\blocksizelimitname}{blockSizeLimit}
 \newcommand{\maxblocksizename}{maxBlockSize}
 \newcommand{\maxheadersizename}{maxHeaderSize}
-\newcommand{\isebbname}{rbIsEBB}
-\newcommand{\rbdataname}{rbData}
-\newcommand{\rbcertsname}{rbCerts}
-\newcommand{\rbsigname}{rbSig}
-\newcommand{\rbixname}{rbIx}
-\newcommand{\rbslname}{rbSl}
-\newcommand{\rbsignername}{rbSigner}
+\newcommand{\isebbname}{bIsEBB}
+\newcommand{\bdataname}{bData}
+\newcommand{\bcertsname}{bCerts}
+\newcommand{\bsigname}{bSig}
+\newcommand{\bixname}{bIx}
+\newcommand{\bslname}{bSl}
+\newcommand{\bsignername}{bSigner}
 
 \newcommand{\qrestrname}{qRestrict}
 \newcommand{\qpopname}{qPop}
 \newcommand{\qheadname}{qHead}
 \newcommand{\qpushname}{qPush}
 
-%% 
+%%
 %% Functions and relations
 %%
 \newcommand{\sign}[4]{\fun{\signname}\ #1 ~ #2 ~ #3 ~ #4}
@@ -132,12 +131,12 @@
 \newcommand{\maxblocksize}[1]{\fun{\maxblocksizename} ~ #1}
 \newcommand{\maxheadersize}[1]{\fun{\maxheadersizename} ~ #1}
 \newcommand{\isebb}[1]{\fun{\isebbname} ~ #1}
-\newcommand{\rbdata}[1]{\fun{\rbdataname} ~ #1}
-\newcommand{\rbcerts}[1]{\fun{\rbcertsname} ~ #1}
-\newcommand{\rbsig}[1]{\fun{\rbsigname} ~ #1}
-\newcommand{\rbix}[1]{\fun{\rbixname} ~ #1}
-\newcommand{\rbsl}[1]{\fun{\rbslname} ~ #1}
-\newcommand{\rbsigner}[1]{\fun{\rbsignername} ~ #1}
+\newcommand{\bdata}[1]{\fun{\bdataname} ~ #1}
+\newcommand{\bcerts}[1]{\fun{\bcertsname} ~ #1}
+\newcommand{\bsig}[1]{\fun{\bsigname} ~ #1}
+\newcommand{\bix}[1]{\fun{\bixname} ~ #1}
+\newcommand{\bsl}[1]{\fun{\bslname} ~ #1}
+\newcommand{\bsigner}[1]{\fun{\bsignername} ~ #1}
 
 \newcommand{\qpop}[1]{\fun{\qpopname} ~ #1}
 \newcommand{\qhead}[1]{\fun{\qheadname} ~ #1}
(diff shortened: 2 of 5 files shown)


mdimjasevic (commented):
Can I ask what is the reason for not talking about the genesis block any more?

dnadales:
The LaTeX file has some [compilation errors](https://hydra.iohk.io/build/499470/nixlog/1), I guess they have to do with the bibliography.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -194,6 +193,9 @@ \section{Introduction}
 In particular, this document gives a formalisation that should be
 straightforward to implement in a programming language, e.g., in Haskell.
 
+This document is intended to be read in conjunction with \cite{ledger-spec},
+which covers the payload carried around in the blockchain. Certain of the
```
dnadales, line 177: Certain parts?

specs/chain/latex/blockchain-spec.tex
```diff
@@ -695,17 +657,17 @@ \subsection{Blockchain Extension}
...
       \exists \var{\var{vk_s}}.~\var{vk_s} = \delegation{\var{ds}}{\var{vk_d}} \\
-      \verify{\var{vk_d}}{(\rbdata{b})}{(\rbsig{b})} \\
+      \verify{\var{vk_d}}{(\bdata{b})}{(\bsig{b})} \\
       \size{\signmap{\var{vk_s}}} \leq K \cdot t \\
       \signmapname' = \incixmap{\var{ix}}{\var{vk_s}}{(\trimix{\signmapname}{\var{ix}})} \\
-      p' = b 
+      p' = b
      \end{array}}
-   & (\var{DSEnv}) \vdash \var{ds} \trans{\text{DELEG}}{\rbcerts{b}} \var{ds'}
+   & (\var{DSEnv}) \vdash \var{ds} \trans{\text{DELEG}}{\bcerts{b}} \var{ds'}
```
dnadales: BTW, Jared and I use lowercase letters for environment and state variables. I don't know if we want to do the same here.

nc6: Yes, will update this

specs/chain/latex/blockchain-spec.tex
nc6: I don't think so... this says "certain of the xs and ys in ..."

nc6:
@mdimjasevic The reason for not discussing the genesis block is that there is no such thing. Cardano doesn't have a distinguished genesis block. It certainly doesn't have one that would contain the information we want (because such a block would predate the BFT way of doing things).

specs/chain/latex/blockchain-spec.tex
```diff
@@ -738,6 +700,12 @@ \subsection{Blockchain Extension}
   \label{fig:blockchain-extension}
 \end{figure}
 
+\begin{thebibliography}{1}
```
dnadales: `natbib` is complaining:

```
! Package natbib Error: Bibliography not compatible with author-year citations.
```

Do we need to use `natbib`? If so the only solution I can think of is to add a `.bib` file.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -423,80 +392,133 @@ \subsection{Blocks}
...
+  \label{fig:ts-types:sigcnt}
+\end{figure}
 
+\begin{figure}[ht]
+  \begin{equation*}
+    \inference
+    {
+      \var{K} = \fun{blockSignatureCountWindow} ~ pp & \var{t} = \fun{blockSignatureCountThreshold} ~ pp \\
+      \forall \var{vk_s} \in \var{dms^{-1}} ~ \var{vk_d} . \size{\signmap{\var{vk_s}}} \leq K \cdot t \\
+      \signmapname' = \incixmap{\var{bIx}}{\left( \var{dms^{-1}} ~ \var{vk_d} \right)}{(\trimix{\signmapname}{\var{bIx}})}
```
dnadales: Shouldn't the second argument of `incixmap` by `VKey_G` instead of a set?

dnadales: BTW, I also realized that `trimx` uses `qRestrict` which requires a `K` parameter, which is not passed to it.

mdimjasevic:
> @mdimjasevic The reason for not discussing the genesis block is that there is no such thing. Cardano doesn't have a distinguished genesis block. It certainly doesn't have one that would contain the information we want (because such a block would predate the BFT way of doing things).

I wasn't aware of this. Where can I look to check for myself that there really is no genesis block?

specs/chain/latex/blockchain-spec.tex
nc6: Yeah, updating this

specs/chain/latex/blockchain-spec.tex
nc6: I modified it to take a set.

nc6: Realised I missed updating the type signature, done now

mdimjasevic (dismissed):
As removing the genesis block from the specification seems quite a change to me and I am not aware where it comes from, I'm asking for references witnessing the need to this change.

dnadales:
> As removing the genesis block from the specification seems quite a change to me and I am not aware where it comes from, I'm asking for references witnessing the need to this change.

I guess the confusion might arise from the fact that epoch boundary blocks are called [genesis blocks](https://github.com/input-output-hk/cardano-sl/blob/34337f53764bc558396be8bc146699ba9d6e6f4e/chain/src/Pos/Chain/Block/Genesis.hs#L4) in `cardano-sl`.

We do have data like the "genesis hash" which is needed to make [the very first epoch boundary block](https://github.com/input-output-hk/cardano-sl/blob/34337f53764bc558396be8bc146699ba9d6e6f4e/chain/src/Pos/Chain/Block/Block.hs#L244).

So I think what you called genesis block is actually an epoch boundary block.

dnadales:
About the genesis hash, it seems to be stored in a [`Config` data structure](https://github.com/input-output-hk/cardano-sl/blob/bcacaa9b5a78f48c7db347f73fec2fe5421e0dfd/chain/src/Pos/Chain/Genesis/Config.hs#L165).

The genesis config seems to be loaded [here](https://github.com/input-output-hk/cardano-sl/blob/a9a091f324d42c5034c2ed1cde1e9342f8182306/lib/src/Pos/Launcher/Launcher.hs#L59), which calls [`withConfigurations`](https://github.com/input-output-hk/cardano-sl/blob/ccca8b346413f12e9fad8cab4c855e69669a2151/lib/src/Pos/Launcher/Configuration.hs#L192).

dnadales (approved):

specs/chain/latex/Makefile
```diff
@@ -36,7 +36,7 @@ $(DOCNAME).pdf: $(DOCNAME).tex
 	latexmk -pdf -pdflatex="pdflatex -interaction=nonstopmode" -use-make $(DOCNAME).tex
 
 watch: $(DOCNAME).tex
-	latexmk -pvc -pdf -pdflatex="pdflatex -interaction=nonstopmode" -use-make $(DOCNAME).tex
+	latexmk -pvc -pdf -pdflatex="pdflatex -interaction=nonstopmode -synctex=1" -use-make $(DOCNAME).tex
```
dnadales, line 39: Do you have any pointers on how to setup emacs to support this? :D

dnadales (approved):

specs/ledger/latex/update-mechanism.tex
```diff
@@ -148,6 +148,8 @@ \subsection{Update proposals}
 \item Script version: $\var{scriptVersion}$
 \item Update adoption threshold: $\var{upAdptThd}$. This would correspond with
   `srInitThd` in `SoftForkRule`.
+\item Chain stability parameter: $\var{chainStabilityParameter}$.
```
dnadales, line 151: Maybe we need to clarify whether these parameters are part of `BlockVersionData`:

```haskell
-- | Data which is associated with 'BlockVersion'
data BlockVersionData = BlockVersionData
  { bvdScriptVersion     :: !ScriptVersion
  , bvdSlotDuration      :: !NominalDiffTime
  , bvdMaxBlockSize      :: !Natural
  , bvdMaxHeaderSize     :: !Natural
  , bvdMaxTxSize         :: !Natural
  , bvdMaxProposalSize   :: !Natural
  , bvdMpcThd            :: !LovelacePortion
  , bvdHeavyDelThd       :: !LovelacePortion
  , bvdUpdateVoteThd     :: !LovelacePortion
  , bvdUpdateProposalThd :: !LovelacePortion
  , bvdUpdateImplicit    :: !FlatSlotId
  , bvdSoftforkRule      :: !SoftforkRule
  , bvdTxFeePolicy       :: !TxFeePolicy
  , bvdUnlockStakeEpoch  :: !EpochIndex
  } deriving (Show, Eq, Ord, Generic)
```

For instance `bvdUpdateImplicit` could correspond with `upropTTL`.

We could also say that these new fields will be modeled as protocol parameters, even if they do not exist in the old code base.

PR #242, opened 2019-01-16
Dnadales/update update mechanism

### LINKED ISSUE #241: Typo in in "Domain and range operations on sequences"?
I was just going through old notes from Berlin...

Should `$A \restrictdom R$` instead be `$S \restrictdom R$` [here](https://github.com/input-output-hk/cardano-chain/blob/3bf95f74d459226df4c72db0b0a693eebe6739aa/specs/ledger/latex/notation.tex#L41)?

dnadales:
-  Allow updates in either the protocol version or software version, but requiring at least an update proposal in any of these two aspects.
- Remove the limits in the number of update proposals can submit in an epoch.
- Use slot numbers instead of block numbers.
- Be more lenient on the endorsements that are accepted: now it is possible for a node to endorse an unconfirmed proposal, however this will not affect the update process.

Closes #241.

### DIFF
--- a/specs/ledger/latex/blockchain-interface.tex
+++ b/specs/ledger/latex/blockchain-interface.tex
@@ -178,7 +178,7 @@ labels have the following meaning:
         \var{k} & \mathbb{N} & \text{chain stability parameter}\\
         \var{u} & \mathbb{N} & \text{update proposals time-to-live}\\
         \var{e_c} & \Epoch & \text{current epoch}\\
-        \var{b_n} & \BkNr & \text{current block number}\\
+        \var{s_n} & \Slot & \text{current slot number}\\
         \var{dms} & \VKeyGen \mapsto \VKey & \text{delegation map}\\
       \end{array}\right)
   \end{align*}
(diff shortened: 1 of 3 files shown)


ruhatch (commented):
LGTM!

ruhatch:
Does this include discussions from Slack over the last couple of days?

dnadales:
> Does this include discussions from Slack over the last couple of days?

@mhuesch and @cleverca22: If I recall correctly there where two remarks:

0. Would skipping software update or protocol updates in a proposal be a problem?
1. Do we need to limit the number of proposals per-key per-epoch?

For 0 my conclusion was that the only inconvenient this could bring is that a node could see a block rejected that it doesn't expect. However this should not affect the soundness of the update mechanism. 

As for 1, after discussing with @kantp I just removed this restriction.

mhuesch:
@dnadales yes, I think your comment about point (0) is correct. It would only be an issue, for example, with block size increases - if an intermediate update doubling the max block size was lost, then the next one (intended to double it again) would actually be 4x'ing, and would be rejected. But the update system is still sound, and we can observe a UP being rejected by checking logs, so it is fine.

nc6 (approved):
LGTM

PR #243, opened 2019-01-16
Shared bibliography

nc6:
This moves the bibliography and IOHK style file into a shared directory, for use
by other things as well.


dnadales:
This would probably break the nix builds, since all the sources need to be in one directory. Maybe we want to unify our `default.nix`'s and put them inside `spec`, and have the build output the three documents.

nc6:
Ah, good point :-(

PR #244, opened 2019-01-16
Implement HeapWords type class for UTxO types

ruhatch:


src/Cardano/Chain/Txp/Tx.hs
```diff
@@ -184,7 +188,10 @@ instance Bi TxIn where
   encodedSizeExpr size _ = 2 + knownCborDataItemSizeExpr
     (szCases [Case "TxInUtxo" $ size $ Proxy @(TxId, Word32)])
 
-instance NFData TxIn
+instance MemSize TxIn where
+  memSize = \case
+    TxInUtxo txid w32 -> memSize2 txid w32
+    TxInUnknown _ bs -> 3 + memSize bs
```
redxaxder: I think the`!Word8` is unpacked here as well. If so, this would be `2 + memSize bs`.

src/Cardano/Chain/Common/AddrSpendingData.hs
```diff
@@ -126,6 +126,13 @@ instance Bi AddrType where
             tag -> ATUnknown tag
     encodedSizeExpr size _ = encodedSizeExpr size (Proxy @Word8)
 
+instance MemSize AddrType where
+  memSize = \case
+    ATPubKey -> 0
+    ATScript -> 0
+    ATRedeem -> 0
+    ATUnknown tag -> memSize1 tag
```
redxaxder: I believe the `!Word8` in `ATUnknown` is unpacked, which would make this 2.

src/Cardano/Chain/Txp/Tx.hs
```diff
@@ -248,6 +254,9 @@ instance Bi TxOut where
   encodedSizeExpr size pxy =
     1 + size (txOutAddress <$> pxy) + size (txOutValue <$> pxy)
 
+instance MemSize TxOut where
+  memSize (TxOut address _) = 3 + memSize address
```
redxaxder: Since `Lovelace` is a single `Word64`, and `TxOut` has a `!Lovelace`, I think this one is unpacked as well. If so, we'd have `2 + memSize address`

src/Cardano/Chain/Txp/Tx.hs
redxaxder: Err, I take that back. I think you were right.

It's 1 for the tag, 1 for the word8, 1 for the pointer to the bytestring, and then the bytestring.

src/Cardano/Chain/Txp/Tx.hs
redxaxder: I think you were right here, too.

src/Cardano/Chain/Txp/Tx.hs
ruhatch: 1 (header) + 1 (Word8) + 1 (bs pointer) + memSize bs?

src/Cardano/Chain/Common/AddrSpendingData.hs
ruhatch: Ah, does the `Word8` get unpacked even when this type has more than one constructor? That does make sense actually...

src/Cardano/Chain/Txp/Tx.hs
ruhatch: Same as the first one, this is with it unpacked I believe.

1 header + 1 word64 + 1 pointer to address + address?

src/Cardano/Chain/Common/AddrSpendingData.hs
redxaxder: I think the number of constructors matters for the value that is potentially being unpacked.

ruhatch: :+1:

redxaxder (approved):

crypto/src/Cardano/Crypto/Hashing.hs
```diff
@@ -114,6 +114,9 @@ instance (Typeable algo, Typeable a, HashAlgorithm algo) => Bi (AbstractHash alg
     let realSz = hashDigestSize (panic "unused, I hope!" :: algo)
     in fromInteger (toInteger (withWordSize realSz + realSz))
 
+instance HeapWords (AbstractHash algo a) where
+  heapWords _ = 12 -- Hardcoded ass 8 word overhead of Digest + 4 words for hash
```
dnadales: Harcoded ass?

ruhatch: Hahaha

mdimjasevic:
Unless I'm missing something, this isn't compiling and it was merged to `master`. When I run `stack build` on commit 4e46b44, I get this:

```
    [ 1 of 28] Compiling Cardano.Crypto.Hashing ( src/Cardano/Crypto/Hashing.hs, .stack-work/dist/x86_64-linux/Cabal-2.2.0.1/build/Cardano/Crypto/Hashing.o )
    
    ~/projects/iohk/chain-spec/265-crpyto-ghc863/crypto/src/Cardano/Crypto/Hashing.hs:117:10: error:
        Not in scope: type constructor or class ‘HeapWords’
        |
    117 | instance HeapWords (AbstractHash algo a) where
        |          ^^^^^^^^^
```

PR #246, opened 2019-01-18
[DEVOPS-1214] Nix-toolification of cardano-chain

angerman:
This adds the nix-tools infra to cardano-chain.
Cross compilation is blocked on additional support
in iohk-nix, which will come in due course.


cleverca22 (approved):

nc6:
This looks reasonable, but also makes the nix infrastructure in this repo look a lot more complicated. Can we include with this PR appropriate instructions for what one needs to do for such things as:

- Adding a new Haskell specification
- Adding a new LaTeX specificiation
- Changing the dependencies of one of the packages

nc6 (changes requested):
Documentation as per previous comment?

angerman:
@nc6 We've moved the default and release nix templates into iohk-nix, which should make this easier to read (I hope).

In general you will be able to use `stack-to-nix` (instructions will follow) to turn your regular `stack.yaml` file into the necessary `nix/pkgs.nix` file.  That is also where you could do custom
configurations of haskell packages should they need any configuration beyond what stack
already sets.

You should also see the custom packages section where the LaTeX specs are in.  Let me know any other thoughts you have.  I'd be happy make this work for you as good as possible.

nix/pkgs.nix
```diff
@@ -0,0 +1,48 @@
+{ args ? { config = import ./config.nix; }
+, nixpkgs ? import <nixpkgs>
+}:
+let
+  pkgs = nixpkgs args;
+in
+let
+
+  inherit (import ./haskell-hackage-stackage.nix { inherit pkgs; }) haskell hackage stackage;
```
angerman: @rvl, I think these should become arguments, and be feed in from iohk-nix. What do you think? That would remove the need to have a `haskell-hackage-stackage.nix` expression per repo. And we rally don't want to have to bump them across projects do we? Rather bump `iohk-nix` as needed.

nix/pkgs.nix
```diff
@@ -0,0 +1,48 @@
...
+      (hackage: {
+          hsc2hs = hackage.hsc2hs."0.68.4".revisions.default;
+          # stackage beautifully omitts the Win32 pkg
+          Win32 = hackage.Win32."2.6.2.0".revisions.default;
+      })
+    ];
+    # package customizations
+    modules = [
+      haskell.ghcHackagePatches.${(stack-pkgs.overlay hackage).compiler.nix-name}
+      {
```
angerman: @rvl, @jbgi: I wonder if we could feed these in as cross patches from iohk-nix as well. it seems stupid if we have to replicate all patches in each and every repo.  Ideally we'd even apply the patches transparently with the project having to worry about it :(

angerman:
This currently depends on input-output-hk/iohk-nix#30. Once merged, this should be good to go.

nc6 (approved):

angerman:
bors try

angerman:
@nc6 the stack-rebuild build kite job failed. All I did was make a small change to the stack.yaml file to contain all targets:
https://github.com/input-output-hk/cardano-chain/pull/246/files#diff-fafd0cdcd559a7b124cc61c29413fb54

Any idea?

nc6:
@angerman Looks like we're trying to pass an option that doesn't exist to the chain tests. But we're about to prune the specs from the repo (and move them to fm-ledger-rules), so I think we should just remove this change from the stack.yaml file and go ahead with this anyway.

nc6:
bors try

angerman:
@nc6 Alright. I'll take care of it and merge tomorrow SGT then.

angerman:
@nc6 one more thing. Do we need to keep the stack build job if we have hydra?

nc6:
You mean the buildkite job? Quite possibly not, but let's deal with that in a separate PR.

PR #249, opened 2019-01-21
Introduce epoch transition system.

nc6:
Would be good to validate the order that the updates are applied in in particular!

### DIFF
--- a/specs/chain/latex/blockchain-spec.tex
+++ b/specs/chain/latex/blockchain-spec.tex
@@ -55,6 +55,7 @@
 \newcommand{\var}[1]{\mathit{#1}}
 \newcommand{\fun}[1]{\mathsf{#1}}
 \newcommand{\type}[1]{\mathsf{#1}}
+\newcommand{\pp}[1]{\mathsf{#1}}
 \newcommand{\size}[1]{\left| #1 \right|}
 \newcommand{\trans}[2]{\xlongrightarrow[\textsc{#1}]{#2}}
 \newcommand{\seqof}[1]{#1^{*}}
@@ -237,6 +238,12 @@ There are several standard sets used in the document:
 
 \newcommand{\Epoch}{\type{Epoch}}
 
+\newcommand{\ETState}{\type{ETState}}
+\newcommand{\ETEnv}{\type{ETEnv}}
+
+% Imported definitions
+\newcommand{\UPIState}{\type{UPIState}}
+
 This section should cover:
 \begin{itemize}
 \item How we determine the end of an epoch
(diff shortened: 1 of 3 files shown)


specs/chain/latex/blockchain-spec.tex
```diff
@@ -245,7 +252,148 @@ \section{Epoch transitions}
 
 The epoch transition rules should compute whether a given slot is at the end of
 an epoch
+\begin{figure}[ht]
+  \emph{Protocol parameters}
+  \begin{equation*}
+    \begin{array}{r@{~\partialf~}l!{~\in~\ProtParams~}r}
+    \pp{slotsPerEpoch} & \mathbb{N} & \text{Number of slots per epoch} \\
```
dnadales: I guess this is a "new" protocol parameter. We'd have to see how to reconcile this with the current block structure. It seems that [a block contains a `SlotId`](https://github.com/input-output-hk/cardano-sl/blob/develop/docs/block-processing/types.md#main-block-header), and this `SlotId` is defined to be an epoch, plus a "local slot" (or relative slot):

```haskell
-- file: cardano-sl/core/src/Pos/Core/Slotting/SlotId.hs

-- | Slot is identified by index of epoch and index of slot in
-- this epoch. This is a global index, an index to a global
-- slot position.
data SlotId = SlotId
    { siEpoch :: !EpochIndex
    , siSlot  :: !LocalSlotIndex
    } deriving (Show, Eq, Ord, Generic, Typeable)

```

So maybe we don't need this parameter, since it seems we have the epoch information in a block.

On the other hand, it seems we would need to calculate the slot from the epoch and the current slot, so this might mean adding an extra state variable.

dnadales:
I've made some comments [here](https://drive.google.com/file/d/1AaMtO2ieg8k4FfqHGSCncLAw1UXLJBIB/view?usp=sharing).

I think we need to check which information comes in the block, since we might not need to calculate the current epoch (but if we only have the relative slot we need to calculate the absolute slot from this).

specs/chain/latex/blockchain-spec.tex
nc6: I've been basing this on the assumption that our "new" slot structure will contain only the absolute slot, rather than a slot/epoch pair. So I'm assuming that in such a case we would drop the epoch information directly in a block. But it's a fair point that in the current system this seems the wrong way around.

I think it would be good to discuss the approach we take with regards to this and the Shelley spec.

Also this raises another point - I'm using the slot from the environment, but this is completely wrong! This is the "slot at this point in time", not "slot in the block"! So this just needs to be used to check that we're not issuing something for a future slot.

dnadales (approved):
LGTM.

PR #252, opened 2019-01-22
[#185] Rename some update mechanism types

ruhatch:
This renames some of the main types in the update mechanism:
- `BlockVersion` -> `ProtocolVersion`
- `BlockVersionData` -> `ProtocolParameters`
- `BlockVersionModifier` -> `ProtocolParameterUpdate`
- `BlockVersionState` -> `ProtocolVersionState` (this one might change later)


src/Cardano/Chain/Block/ExtraHeaderData.hs
```diff
@@ -21,15 +21,15 @@ import qualified Formatting.Buildable as B
...
 import Cardano.Chain.Update.SoftwareVersion
   (SoftwareVersion, SoftwareVersionError, checkSoftwareVersion)
 import Cardano.Crypto (Hash)
 
 
 -- | Represents block header extra data
 data ExtraHeaderData = ExtraHeaderData
-  { ehdBlockVersion    :: !BlockVersion
+  { ehdProtocolVersion :: !ProtocolVersion
   -- ^ Version of block
```
dnadales: Do we need to adapt the comment as well?

src/Cardano/Chain/Genesis/Generate.hs
```diff
@@ -214,14 +214,14 @@ generateGenesisData startTime genesisSpec = do
...
     genesisData = GenesisData
-      { gdBootStakeholders = bootStakeholders
-      , gdHeavyDelegation  = genesisDlg
-      , gdStartTime        = startTime
-      , gdNonAvvmBalances  = nonAvvmDistr
-      , gdBlockVersionData = gsBlockVersionData genesisSpec
-      , gdK                = gsK genesisSpec
-      , gdProtocolMagic    = pm
-      , gdAvvmDistr        = fakeAvvmDistr <> realAvvmMultiplied
+      { gdBootStakeholders   = bootStakeholders
```
dnadales, line 217: Here's an example in which aligning is a bit annoying for reviewing purposes. Actually nothing is changed here besides the space. Just a side note. Sorry for the noise.

dnadales (approved):
:+1:

src/Cardano/Chain/Genesis/Generate.hs
ruhatch: No worries, I believe if you review with no whitespace changes you won't see these lines. That way we get nice alignment for readers and a good time for reviewers. I think you can set no whitespace as a default as well somehow.

Issue #259, opened 2019-01-23
Switch to new notation for assignment/equality


nc6:
Per the results of [this poll](https://input-output-rnd.slack.com/archives/CCZSDE005/p1548167327100200) we should switch to using `:=` for assignment (let-binding) in rules, and `=` for equality testing.

The following can be used to get a nicely-formatted `:=` in LaTeX

```
\newcommand{\leteq}{\ensuremath{\mathrel{\mathop:}=}}
```

nc6:
Addressed in the chain spec

dnadales:
Moved to: https://github.com/input-output-hk/fm-ledger-rules/issues/296

PR #260, opened 2019-01-23
Switch to using new let-binding syntax.

nc6:
Addresses #259 for the chain spec. Also makes a couple of tidy-up changes along
the way.

### DIFF
--- a/specs/chain/latex/blockchain-spec.tex
+++ b/specs/chain/latex/blockchain-spec.tex
@@ -62,6 +62,8 @@
 \newcommand{\nextdef}{\\[1em]}
 \newcommand{\serialised}[1]{\llbracket \var{#1} \rrbracket}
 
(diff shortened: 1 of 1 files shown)


specs/chain/latex/blockchain-spec.tex
```diff
@@ -725,8 +726,7 @@ \section{Blockchain Extension}
   %
   \begin{equation*}
   \begin{array}{r@{~\in~}lr}
-    \var{data} & \Data    & \text{data}\\
-    \var{ds} & \DelegState & \text{ledger layer delegation state}
```
nc6, line 729: This is from the ledger spec. Other imported definitions are left undefined, so removing this for consistency

dnadales (approved):
So long overloaded `=` operator :cry:

Issue #258, opened 2019-01-23
Group the premises of the chain rules into header transition and body transitions


dnadales:
The current rule for chain extension is getting large:

![image](https://user-images.githubusercontent.com/175315/51602961-a631ed00-1f08-11e9-80f1-e44499253673.png)

And the consensus layer will need to use a header transition to validate just the header.

In view of the above considerations, we should define two transitions, say `BHEAD` and `BBODY`, and group the `CHAIN` transition premises into the former two.

PR #261, opened 2019-01-24
Factor out BHEAD and BBODY rules

nc6:
Addresses #258 

This commit splits the block header and body processing into separate rules.

### DIFF
--- a/specs/chain/latex/blockchain-spec.tex
+++ b/specs/chain/latex/blockchain-spec.tex
@@ -495,8 +495,6 @@ $1/5 \leq t \leq 1/4$.
 
 \newcommand{\bheadname}{bHead}
 \newcommand{\bhead}[1]{\fun{\bheadname}\ #1}
-\newcommand{\bbodyname}{bBody}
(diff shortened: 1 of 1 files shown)


specs/chain/latex/blockchain-spec.tex
```diff
@@ -824,120 +929,102 @@ \section{Blockchain Extension}
...
+\end{equation*}
+\vspace{20pt}
+\begin{equation*}
   \inference
   {
     \neg\isebb{b} \\
     \fun{dms} ~ \var{ds} \vdash
     {\left(
         \begin{array}{l}
-          \sepoch{s_{last}} \\
```
dnadales, line 838: We have one environment too many now

image

dnadales (approved):
LGTM modulo the extra environment. 

Thanks Nick!

specs/chain/latex/blockchain-spec.tex
nc6: Whoops...

PR #266, opened 2019-01-24
Add: Block body validation.

nc6:
TODO: Split aside the update rules, handle the case where there is no update
proposal.

### DIFF
--- a/specs/chain/latex/blockchain-spec.tex
+++ b/specs/chain/latex/blockchain-spec.tex
@@ -503,6 +503,8 @@ $1/5 \leq t \leq 1/4$.
 \newcommand{\bhtosign}[1]{\fun{\bhtosignname}\ #1}
 \newcommand{\bhslotname}{bhSlot}
 \newcommand{\bhslot}[1]{\fun{\bhslotname}\ #1}
+\newcommand{\bupdpayloadname}{bUpdPayload}
+\newcommand{\bupdpayload}[1]{\fun{\bupdpayloadname}\ #1}
 
 \newcommand{\bslotname}{bSlot}
 \newcommand{\bslot}[1]{\fun{\bslotname}\ #1}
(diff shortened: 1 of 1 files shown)


specs/chain/latex/blockchain-spec.tex
```diff
@@ -772,6 +780,10 @@ \subsection{Block body processing}
   \begin{equation*}
     \inference
     { \maxblocksize \mapsto \var{b_{max}} \in \var{pps} && \bsize{b} \leq \var{b_{max}} \\
+      \var{bh} \leteq \bhead{b} & \var{vk_d} \leteq \bissuer{\var{bh}} \\
```
dnadales, line 783: Now that we have a bunch of verification here we might want to factor that out into a separate rule.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -717,9 +720,12 @@ \subsection{Block body processing}
       \fun{bUtxo} & \Block \totalf \UTxO & \text{Block UTxO payload} \\
       \fun{\bcertsname} & \Block \totalf \seqof{\DCert}
       & \text{block certificates} \\
-      \fun{bUpdProp} & \Block \totalf \UProp & \text{Block update proposal payload} \\
+      \fun{bUpdProp} & \Block \totalf \UProp^{?} & \text{Block update proposal payload} \\
```
dnadales, line 723: We need to update the UPIREG rule to work with this kind of maybe type.

dnadales (approved):

specs/chain/latex/blockchain-spec.tex
nc6: Well, either that or have a wrapper. I was thinking of factoring the three update rules out and having that taken care of there.

PR #271, opened 2019-01-25
Add a generator valid for valid blockchains containing only delegation payload

### LINKED ISSUE #195: Generator for (valid) blockchain (delegation)
Generator at the blockchain layer (when only carrying delegation payload).

Requires:

- [x] Compute the size of a block and block header.
- [x] Adapt the STS rules to match the modification in the spec in `master`.
- [x] Define the `HasTrace` instance for the `CHAIN` transition system.

Minor tasks before PR:

- [x] Add compiler warnings, treat warnings as errors, and `development` flag.

Retrospective
==========

- Nix build:
  - :disappointed: The addition of `cardano-prelude`, which was really easy with `stack`, broke the `nix` build, and I'm spending quite a lot of time trying to figure out how to fix this.
  - :disappointed: It is still not clear why [PR-246](https://github.com/input-output-hk/cardano-chain/pull/246#issuecomment-456783393), which will automate this painful process, hasn't been merged.
  - :pray: devops (@cleverca22) was really helpful.
  - :scream: editing the nix configuration feels like black magic to me.

dnadales:
Only the relevant subset of the [current blockchain rules](https://hydra.iohk.io/build/550660/download/1/blockchain-spec.pdf) are implemented.

I've noticed some problems with the generators, which might become more serious in the future:

- The size of the traces is quite limited (about 25 blocks with a size parameter of 100). It seems that as more rules are composed (e.g. `CHAIN` + `DELEGATION`) the trace size gets smaller.
- When I wrote the traces generator I was not able use `Gen.discard` when a rule failed on a generated signal: there were too many failures, and `hedgehog` will simply give up. This seems to have a negative impact on shrinking. When counterexamples are found these are huge, even though it is clear that simpler counter examples could have been found by a manually encoded shrinks (so the "shrinking for free" killer feature of hedgehog might be backfiring here.
- The memory usage seems quite large.
- The rules need to be encoded in the STS and also in the signal generators. By doing this we are not able to detect errors in the STS encoding of the rules. As an example, comment out [this line](https://github.com/input-output-hk/cardano-chain/compare/dnadales/195-generator-valid-blockchain-delegation?expand=1#diff-b95b25e0432df58ccaf29a5e5fd8db33R63) and see how the only property we have in the tests still passes. People might be fine with this, since we want to produce valid traces only in the very short term, but I have an uneasy feeling about it.

Take the above remarks with a pinch of salt, since there are the observations of a single person who might have made quite some mistakes.

Closes #195


dnadales:
@cleverca22 is helping me to fix the broken nix build.

specs/chain/hs/src/Cardano/Spec/Chain/STS/Block.hs
```diff
@@ -0,0 +1,104 @@
...
+  = Block
+  { _bHeader :: BlockHeader
+  , _bBody :: BlockBody
+  } deriving (Eq, Show)
+
+makeLenses ''Block
+
+-- | Compute the size (in words) that a block takes.
+bSize :: Block -> Natural
+bSize = fromInteger . toInteger . heapWords
```
ruhatch, line 62: Couldn't you just use `fromIntegral` here?

specs/chain/hs/src/Cardano/Spec/Chain/STS/Block.hs
```diff
@@ -0,0 +1,104 @@
...
+  heapWords b = heapWords2 (b ^. bHeader) (b ^. bBody)
+
+instance HeapWords BlockHeader where
+  heapWords header
+    -- The constant 12 is was taken from:
+    --
+    -- https://github.com/input-output-hk/cardano-chain/pull/244/files#diff-2955aa8b04471dc586f90cb5f22948beR118
+    --
+    -- 12 = 8 words of digest + 4 words for hash
+    = 12
```
ruhatch, line 74: You'll need an extra `1` for the pointer to the hash I think

specs/chain/hs/src/Cardano/Spec/Chain/STS/Block.hs
```diff
@@ -0,0 +1,104 @@
...
+    + heapWords3 (header ^. bSlot)
+                 (header ^. bIssuer)
+                 (header ^. bSig)
+
+instance HeapWords BlockBody where
+  heapWords body = heapWords1 (body ^. bDCerts)
+
+-- | Compute the size (in words) that a block header.
+bHeaderSize :: BlockHeader -> Natural
+bHeaderSize = fromInteger . toInteger . heapWords
```
ruhatch, line 84: `fromIntegral`?

specs/chain/hs/src/Cardano/Spec/Chain/STS/Block.hs
```diff
@@ -0,0 +1,104 @@
...
+hashHeader = hashlazy . pack . show
+-- TODO: we might want to serialize this properly, without using show...
+
+
+-- | Compute the epoch for the given _absolute_ slot
+sEpoch :: Slot -> Epoch
+sEpoch (Slot s) = Epoch $ s `div` slotsPerEpoch
+  where
+    -- Hardcoded number of slots per epoch, as per Byron.
+    slotsPerEpoch = 21600
```
ruhatch, line 97: Hmm, we probably shouldn't hardcode this., but we might have to for now. In any case I think we should bubble it up, i.e. move it as far up the call stack as possible, and then pass it in there hardcoded. That way if we replace it with some sort of test configuration later on we've already done the plumbing.

specs/chain/hs/src/Cardano/Spec/Chain/STS/Rule/Chain.hs
```diff
@@ -0,0 +1,170 @@
...
+          , _maxBkSz = mBSz
+          , _dLiveness = d
+          , _bkSgnCntW = w
+          , _bkSgnCntT = t
+          , _bkSlotsPerEpoch = spe
+          }
+    initGKeys <- Gen.set (Range.constant 1 70) vkgenesisGen
+    -- If we want to generate large traces, we need to set up the value of the
+    -- "clock-slot" to a sufficiently large value.
+    initSlot <- Slot <$>
```
ruhatch, line 143: I'm confused by this one - why isn't the inital slot 0?

specs/ledger/hs/src/Ledger/Core.hs
```diff
@@ -63,6 +68,9 @@ keyPair o = KeyPair (SKey o) (VKey o)
 -- |A digital signature.
 data Sig a = Sig a Owner deriving (Show, Eq, Ord)
 
+instance HeapWords a => HeapWords (Sig a) where
+  heapWords (Sig a o) = 1 + heapWords2 a o
```
ruhatch, line 72: `heapWords2` takes care of the additional header field I believe, but do check that

specs/stack.yaml
```diff
@@ -9,3 +9,8 @@ extra-deps:
 - hedgehog-0.6.1
```
ruhatch, line 9: I think you should be able to prune some of these extra deps, as they're provided by the snapshot :)

ruhatch (commented):
A few comments and questions, but looks good in general. I'll leave up to someone else to approve as I'm not sure how to address your concerns about potential STS mistakes.

specs/chain/hs/src/Cardano/Spec/Chain/STS/Block.hs
nc6: There's a plan (and issue: https://github.com/input-output-hk/cardano-chain/issues/255) for moving to explicit slot/epoch calculations and moving this to the protocol parameters in the update system. Since that part isn't modelled yet, I think it makes sense to leave this hardcoded for now and then grab it from the protocol parameters later.

specs/chain/hs/src/Cardano/Spec/Chain/STS/Rule/BHead.hs
```diff
@@ -0,0 +1,81 @@
...
+    = ( Slot
+      , Map VKeyGenesis VKey
+      )
+  type State BHEAD
+    = ( Epoch
+      , Slot
+      , Hash
+      , Seq VKeyGenesis
+      , PParams
+      )
```
nc6, line 33: I guess this is one case where big tuples really do mesh well with the way we're writing our rules

specs/chain/hs/src/Cardano/Spec/Chain/STS/Rule/Chain.hs
```diff
@@ -0,0 +1,170 @@
...
+
+instance STS CHAIN where
+  type Environment CHAIN =
+    ( Slot            -- Current slot
+    , Set VKeyGenesis -- Genesis keys
+    , PParams         -- Initial protocol parameters
+    )
+    -- TODO: it can be confusing to have this in the environment. It will be
+    -- used by the initial rule only, and then we'll have to drag it for
+    -- eternity. The state contains the up to date protocol parameters.
```
nc6, line 40: One solution could be to wrap this in an `Initial` wrapper. This wouldn't do much, but would make it clear that this is not a value to be used later.

specs/chain/hs/src/Cardano/Spec/Chain/STS/Rule/Chain.hs
```diff
@@ -0,0 +1,170 @@
...
+    , DIState
+    , PParams
+    )
+
+  type Signal CHAIN = Block
+
+  data PredicateFailure CHAIN
+    = BHeadFailure (PredicateFailure BHEAD)
+    | BBodyFailure (PredicateFailure BBODY)
+    | LedgerFailure (PredicateFailure DELEG)
```
nc6, line 56: Maybe slightly confusing name?

specs/chain/hs/src/Cardano/Spec/Chain/STS/Rule/Chain.hs
```diff
@@ -0,0 +1,170 @@
...
+
+  transitionRules =
+    [ do
+        TRC
+          ( (sNow, gks, _)
+          , (eLast, sLast, hLast, sgs, ds, us)
+          , b ) <- judgmentContext
+
+        (eNext, sNext, h', sgs', us') <- trans @BHEAD $ TRC ( (sNow, ds ^. dms)
+                                                            , (eLast, sLast, hLast, sgs, us)
```
nc6, line 87: Why do we need `eLast` and `sLast` in here? This doesn't seem to match the current spec

specs/chain/hs/src/Cardano/Spec/Chain/STS/Rule/BHead.hs
```diff
@@ -0,0 +1,81 @@
...
+data BHEAD
+
+instance STS BHEAD where
+  type Environment BHEAD
+    = ( Slot
+      , Map VKeyGenesis VKey
+      )
+  type State BHEAD
+    = ( Epoch
+      , Slot
```
nc6, line 29: Slightly different from the spec, which has the "last slot" in the environment, and calculates the epoch from the slot.

specs/chain/hs/src/Cardano/Spec/Chain/STS/Rule/Chain.hs
nc6: This is the "current" slot rather than the initial one. Maybe confusingly named here.

specs/chain/hs/src/Cardano/Spec/Chain/STS/Rule/SigCnt.hs
```diff
@@ -0,0 +1,65 @@
...
+          [vkG] -> do
+            let sgs' = S.drop (S.length sgs + 1 - w) (sgs |> vkG)
+                nrSignedBks = fromIntegral (S.length (S.filter (==vkG) sgs'))
+            nrSignedBks <= fromIntegral w * t ?! TooManyIssuedBlocks vkG
+            return $! sgs'
+          [] -> do
+            failBecause NotADelegate
+            return sgs -- TODO: this is a quite inconvenient encoding for this transition system!
+          _ -> do
+            failBecause NonInjectiveDelegationMap
```
nc6, line 63: This is *not* the same as the current rules, which don't forbid such a situation.

specs/ledger/hs/src/Ledger/Delegation.hs
```diff
@@ -54,8 +56,10 @@ module Ledger.Delegation
   -- * State lens type classes
   , HasScheduledDelegations
   , scheduledDelegations
+  , dms
```
nc6, line 59: Why does github make these things red?

nc6 (approved):
Some differences from the rules. But since this is only a partial implementation, and the rules are in flux, maybe we should leave this and just make sure everything is in sync at a later point.

dnadales:
Superseded by https://github.com/input-output-hk/cardano-chain/pull/302

specs/stack.yaml
dnadales: Thanks for the tip! Done in #301

PR #273, opened 2019-01-28
Add: Update interface section.

nc6:
This covers the case in which there is no update proposal in the block.

Although I'm not so sure the syntax looks that nice :-(

Asking bors to try this in order to get the built PDF...

### DIFF
--- a/specs/chain/latex/blockchain-spec.tex
+++ b/specs/chain/latex/blockchain-spec.tex
@@ -331,51 +331,51 @@ There are several standard sets used in the document:
       }
       {\var{us'''}}
     }
-\end{equation*}
-\vspace{20pt}
-\begin{equation*}
-  \inference
-  { {\left(
-        \begin{array}{l}
-          \var{slot} \\
-          \var{e_n} \\
-          \var{dms}
-        \end{array}
-      \right)}
-    \vdash \var{us} \trans{upivotes}{\var{votes}} \var{us'}
-    &
-    {\left(
-        \begin{array}{l}
-          \var{slot} \\
-          \var{e_n} \\
-          \var{dms}
-        \end{array}
-      \right)}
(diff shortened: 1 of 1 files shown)


nc6:
bors try

specs/chain/latex/blockchain-spec.tex
```diff
@@ -236,6 +236,184 @@ \subsection{Sets}
...
+  {
+    \left(
+      {\begin{array}{c}
+         \var{utxo} \\
+         \var{ds} \\
+         \var{us}
+       \end{array}}
+   \right)
+ }
+ \trans{upi}
```
dnadales: We're missing the signal here. And this also causes the LaTeX compilation to fail with a rather confusing error message, since the next brace will be used as the second argument of `\trans`.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -236,6 +236,184 @@ \subsection{Sets}
...
+ {
+   \left(
+     \begin{array}{l}
+       1 \\
```
dnadales: `1`?

nc6:
bors try

specs/chain/latex/blockchain-spec.tex
```diff
@@ -236,6 +236,150 @@ \subsection{Sets}
...
+    {
+      {\left(
+          \begin{array}{l}
+            \var{slot} \\
+            \var{e_n} \\
+            \var{dms}
+          \end{array}
+        \right)}
+      {\var{us}}
+      \trans{upi}
```
dnadales: `BUPI`?

dnadales (approved):

specs/chain/latex/blockchain-spec.tex
nc6:
```suggestion
      \trans{bupi}
```

PR #275, opened 2019-01-28
Update cardano-mainnet-mirror to lfs version

ruhatch:
This updates the version of `cardano-mainnet-mirror` to a version using git-lfs. This way we don't always have to pull in all the epoch files when doing a recursive clone.


nc6:
I'm confused. So now we're using `git-lfs` _inside_ the submodule? Why not just revert to the previous case where we just used `git-lfs` in this repo. And does this get around whatever problems caused us to switch away from `git-lfs` in the first place?

nc6:
Sorry, didn't mean to close, hit the wrong button!

redxaxder (approved):

PR #276, opened 2019-01-29
Replace git-lfs with cardano-mainnet-mirror Haskell package

ruhatch:
This allows us to make changes to the cardano-binary or cardano-crypto repos without propagating the changes all the way up the codebase. That allows us to have smaller PRs and is a stepping stone to splitting those packages into their own repos.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -11,9 +11,6 @@ steps:
       - "rm -rf /build/cardano-chain"
       - "cp -R . /build/cardano-chain"
       - "cd /build/cardano-chain"
-      - "git lfs install"
-      - "git lfs pull"
-      - "ls -lh test/resources/epochs/"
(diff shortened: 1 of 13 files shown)


.buildkite/pipeline.yml
```diff
@@ -11,9 +11,6 @@ steps:
       - "rm -rf /build/cardano-chain"
       - "cp -R . /build/cardano-chain"
       - "cd /build/cardano-chain"
-      - "git lfs install"
-      - "git lfs pull"
-      - "ls -lh test/resources/epochs/"
```
nc6, line 16: This is just general (unrelated) tidying?

nc6 (approved):

.buildkite/pipeline.yml
ruhatch: Yeah, I've now removed all references to lfs and the submodules

PR #283, opened 2019-01-30
Add: Prose explanations.

nc6:


specs/chain/latex/blockchain-spec.tex
```diff
@@ -839,6 +881,21 @@ \subsection{Block header processing}
...
 
 \subsection{Block body processing}
+
+During processing of the block body, we perform two main functions: verification
+of the body integrity using the proofs contained in the block header, and update
+of the various state components. These rules are given in figure
+\ref{fig:rules:bbody}.
+
+Verification is done independently for the three components of the body payload:
+UTxO, delegation and update. Each of these have three a signature in the block
```
mdimjasevic: "Each of these three have a signature"

specs/chain/latex/blockchain-spec.tex
```diff
@@ -839,6 +881,21 @@ \subsection{Block header processing}
...
+header. Note that Byron-era block payload also has an additional component: the
+VSS payload. This part of the block is unnecessary during the BFT era, and hence
+we do not bother to verify it.
```
mdimjasevic: I find it strange to use a phrase like "we do not bother" in a communication document like this. Can you rephrase this?

mdimjasevic (approved):
The prose looks OK.

I have one LaTeX remark: to avoid a line break in the PDF right in the middle of a figure name (e.g., "Figure\n6"), I suggest to update figure references like this:

```latex
figure~\ref{fig:rules:bbody}
```

This will ensure that e.g. "Figure 6" is always rendered on the same line.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -702,6 +717,33 @@ \section{Block processing}
...
+    recognition that this is a block in a new epoch.
+  \item We check that the current block is being issued for a slot later than
+    the last block issued.
+  \item We check that the current block is being issued for a slot no later than
+    the current slot (as determined by the system clock).
+  \item We verify that the block signer is the current delegate of at least one
+    of the authorised genesis keys.
+  \item We check that the previous block hash contained in the block header
+    corresponds to the known hash of the previous block.
+  \item We check that the header signature correctly verifies the ``signable''
```
dnadales: what about "signed content"?

dnadales (approved):

specs/chain/latex/blockchain-spec.tex
nc6:
```suggestion
UTxO, delegation and update. Each of these three have a signature in the block
```

specs/chain/latex/blockchain-spec.tex
nc6:
```suggestion
we do not verify it.
```

specs/chain/latex/blockchain-spec.tex
nc6: Yeah, maybe? I don't like either, to be honest, but I'll go with that for now.
```suggestion
  \item We check that the header signature correctly verifies the ``signed''
```

dnadales:
bors self destroy.

:weary:

nc6:
bors retry

PR #284, opened 2019-01-31
[#185] Implement update registration rules

### LINKED ISSUE #185: Implement update registration rules

ruhatch:
These are currently not tested against the mainnet chain because we don't have the code for actually adopting proposals yet.

Closed #185

### DIFF
--- a/src/Cardano/Chain/Update/ProtocolParameterUpdate.hs
+++ b/src/Cardano/Chain/Update/ProtocolParameterUpdate.hs
@@ -7,11 +7,12 @@
 
 module Cardano.Chain.Update.ProtocolParameterUpdate
   ( ProtocolParameterUpdate(..)
-  , applyPPU
+  , isEmpty
+  , apply
   )
 where
 
-import Cardano.Prelude
+import Cardano.Prelude hiding (empty)
 
 import Data.Text.Lazy.Builder (Builder)
 import Data.Time (NominalDiffTime)
@@ -119,27 +120,48 @@ instance Bi ProtocolParameterUpdate where
       <*> decode
       <*> decode
 
+empty :: ProtocolParameterUpdate
+empty = ProtocolParameterUpdate
+  { ppuScriptVersion    = Nothing
+  , ppuSlotDuration     = Nothing
+  , ppuMaxBlockSize     = Nothing
+  , ppuMaxHeaderSize    = Nothing
+  , ppuMaxTxSize        = Nothing
+  , ppuMaxProposalSize  = Nothing
+  , ppuMpcThd           = Nothing
+  , ppuHeavyDelThd      = Nothing
+  , ppuUpdateVoteThd    = Nothing
+  , ppuUpdateProposalThd = Nothing
+  , ppuUpdateImplicit   = Nothing
+  , ppuSoftforkRule     = Nothing
+  , ppuTxFeePolicy      = Nothing
+  , ppuUnlockStakeEpoch = Nothing
+  }
+
+isEmpty :: ProtocolParameterUpdate -> Bool
+isEmpty = (== empty)
+
 -- | Apply 'ProtocolParameterUpdate' to 'ProtocolParameters'
-applyPPU :: ProtocolParameterUpdate -> ProtocolParameters -> ProtocolParameters
-applyPPU ppu pp = ProtocolParameters
-  { ppScriptVersion     = fromMaybe (ppScriptVersion pp) (ppuScriptVersion ppu)
-  , ppSlotDuration      = fromMaybe (ppSlotDuration pp) (ppuSlotDuration ppu)
-  , ppMaxBlockSize      = fromMaybe (ppMaxBlockSize pp) (ppuMaxBlockSize ppu)
-  , ppMaxHeaderSize     = fromMaybe (ppMaxHeaderSize pp) (ppuMaxHeaderSize ppu)
-  , ppMaxTxSize         = fromMaybe (ppMaxTxSize pp) (ppuMaxTxSize ppu)
-  , ppMaxProposalSize   = fromMaybe
+apply :: ProtocolParameterUpdate -> ProtocolParameters -> ProtocolParameters
+apply ppu pp = ProtocolParameters
+  { ppScriptVersion    = fromMaybe (ppScriptVersion pp) (ppuScriptVersion ppu)
+  , ppSlotDuration     = fromMaybe (ppSlotDuration pp) (ppuSlotDuration ppu)
+  , ppMaxBlockSize     = fromMaybe (ppMaxBlockSize pp) (ppuMaxBlockSize ppu)
+  , ppMaxHeaderSize    = fromMaybe (ppMaxHeaderSize pp) (ppuMaxHeaderSize ppu)
+  , ppMaxTxSize        = fromMaybe (ppMaxTxSize pp) (ppuMaxTxSize ppu)
+  , ppMaxProposalSize  = fromMaybe
     (ppMaxProposalSize pp)
     (ppuMaxProposalSize ppu)
-  , ppMpcThd            = fromMaybe (ppMpcThd pp) (ppuMpcThd ppu)
-  , ppHeavyDelThd       = fromMaybe (ppHeavyDelThd pp) (ppuHeavyDelThd ppu)
-  , ppUpdateVoteThd     = fromMaybe (ppUpdateVoteThd pp) (ppuUpdateVoteThd ppu)
+  , ppMpcThd           = fromMaybe (ppMpcThd pp) (ppuMpcThd ppu)
+  , ppHeavyDelThd      = fromMaybe (ppHeavyDelThd pp) (ppuHeavyDelThd ppu)
+  , ppUpdateVoteThd    = fromMaybe (ppUpdateVoteThd pp) (ppuUpdateVoteThd ppu)
   , ppUpdateProposalThd = fromMaybe
     (ppUpdateProposalThd pp)
     (ppuUpdateProposalThd ppu)
-  , ppUpdateImplicit = fromMaybe (ppUpdateImplicit pp) (ppuUpdateImplicit ppu)
-  , ppSoftforkRule      = fromMaybe (ppSoftforkRule pp) (ppuSoftforkRule ppu)
-  , ppTxFeePolicy       = fromMaybe (ppTxFeePolicy pp) (ppuTxFeePolicy ppu)
-  , ppUnlockStakeEpoch  = fromMaybe
+  , ppUpdateImplicit   = fromMaybe (ppUpdateImplicit pp) (ppuUpdateImplicit ppu)
+  , ppSoftforkRule     = fromMaybe (ppSoftforkRule pp) (ppuSoftforkRule ppu)
+  , ppTxFeePolicy      = fromMaybe (ppTxFeePolicy pp) (ppuTxFeePolicy ppu)
+  , ppUnlockStakeEpoch = fromMaybe
     (ppUnlockStakeEpoch pp)
     (ppuUnlockStakeEpoch ppu)
   }
--- a/src/Cardano/Chain/Update/Validation/Registration.hs
+++ b/src/Cardano/Chain/Update/Validation/Registration.hs
@@ -1,0 +1,264 @@
+{-# LANGUAGE FlexibleContexts #-}
+
+-- | Validation rules for registering updates
+--
+--   This is an implementation of the rules defined in the Byron ledger
+--   specification
+module Cardano.Chain.Update.Validation.Registration
+  ( RegistrationError
+  , RegistrationState(..)
+  , registerProposal
+  )
+where
+
+import Cardano.Prelude
+
+import qualified Data.ByteString as BS
+import qualified Data.Map.Strict as M
+
+import Cardano.Chain.Common
+import Cardano.Chain.Slotting
+import Cardano.Chain.Update.ApplicationName
+import Cardano.Chain.Update.ProtocolParameters
+import qualified Cardano.Chain.Update.ProtocolParameterUpdate as PPU
+import Cardano.Chain.Update.ProtocolVersion
+import Cardano.Chain.Update.SoftwareVersion
+import Cardano.Chain.Update.Vote
+import Cardano.Crypto
+
+
+-- | RegistrationState keeps track of registered protocol and software update
+--   proposals
+data RegistrationState = RegistrationState
+  { rsProtocolUpdateProposals :: !ProtocolUpdateProposals
+  , rsSoftwareUpdateProposals :: !SoftwareUpdateProposals
+  }
+
+type ProtocolUpdateProposals = Map UpId (ProtocolVersion, ProtocolParameters)
+type SoftwareUpdateProposals = Map UpId SoftwareVersion
+
+type ApplicationVersions = Map ApplicationName (NumSoftwareVersion, SlotId)
+
+-- | RegistrationError captures the ways in which registration could fail
+data RegistrationError
+  = RegistrationDuplicateProtocolVersion ProtocolVersion
+  | RegistrationDuplicateSoftwareVersion SoftwareVersion
+  | RegistrationInvalidProposer StakeholderId
+  | RegistrationInvalidProtocolVersion ProtocolVersion
+  | RegistrationInvalidScriptVersion ScriptVersion ScriptVersion
+  | RegistrationInvalidSignature
+  | RegistrationInvalidSoftwareVersion SoftwareVersion
+  | RegistrationMaxBlockSizeTooLarge (TooLarge Natural)
+  | RegistrationMaxTxSizeTooLarge (TooLarge Natural)
+  | RegistrationProposalEmpty
+  | RegistrationProposalTooLarge (TooLarge Natural)
+  deriving (Eq, Show)
+
+data TooLarge n = TooLarge
+  { tlExpected :: n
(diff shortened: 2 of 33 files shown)


src/Cardano/Chain/Update/Validation/Registration.hs
```diff
@@ -0,0 +1,261 @@
...
+
+-- | RegistrationState keeps track of registered protocol and software update
+--   proposals
+data RegistrationState = RegistrationState
+  { rsProtocolUpdateProposals :: Map UpId (ProtocolVersion, ProtocolParameters)
+  , rsSoftwareUpdateProposals :: Map UpId SoftwareVersion
+  }
+
+-- | RegistrationError captures the ways in which registration could fail
+data RegistrationError
```
dnadales, line 43: Shouldn't we mention that it is an *update* registration in the names of the data type and constructors. This will result in a very long name, but maybe we can abbreviate it to `UpdReg...`.

src/Cardano/Chain/Update/Validation/Registration.hs
```diff
@@ -0,0 +1,261 @@
...
+  = RegistrationDuplicateProtocolVersion ProtocolVersion
+  | RegistrationDuplicateSoftwareVersion SoftwareVersion
+  | RegistrationInvalidProposer StakeholderId
+  | RegistrationInvalidProtocolVersion ProtocolVersion
+  | RegistrationInvalidScriptVersion ScriptVersion ScriptVersion
+  | RegistrationInvalidSignature
+  | RegistrationInvalidSoftwareVersion SoftwareVersion
+  | RegistrationMaxBlockSizeTooLarge Natural Natural
```
dnadales: Would it make sense to add accessors to specify which number is the maximum size, and which number was the actual size?

Or maybe define:

```haskell
data TooLarge n = TooLarge { actual :: n, maximum :: n }

data UpdError = UpdErrorBlockSize TooLarge
                        | UpdErrorProposalSize TooLarge
```

src/Cardano/Chain/Update/Validation/Registration.hs
```diff
@@ -0,0 +1,261 @@
...
+  | RegistrationMaxTxSizeTooLarge Natural Natural
+  | RegistrationProposalTooLarge Natural Natural
```
dnadales: Don't we need to be consistent with naming? We have `RegistrationMaxBlahTooLarge` and `RegistrationBlahTooLarge`

src/Cardano/Chain/Update/Validation/Registration.hs
```diff
@@ -0,0 +1,261 @@
...
+  , rsSoftwareUpdateProposals :: Map UpId SoftwareVersion
+  }
+
+-- | RegistrationError captures the ways in which registration could fail
+data RegistrationError
+  = RegistrationDuplicateProtocolVersion ProtocolVersion
+  | RegistrationDuplicateSoftwareVersion SoftwareVersion
+  | RegistrationInvalidProposer StakeholderId
+  | RegistrationInvalidProtocolVersion ProtocolVersion
+  | RegistrationInvalidScriptVersion ScriptVersion ScriptVersion
```
dnadales, line 48: What does each `ScriptVersion` represent?

src/Cardano/Chain/Update/Validation/Registration.hs
```diff
@@ -0,0 +1,261 @@
...
+  :: MonadError RegistrationError m
+  => ProtocolVersion
+  -> ProtocolParameters
+  -> ApplicationVersions
+  -> RegistrationState
+  -> AProposal ByteString
+  -> m RegistrationState
+validateProposal adoptedPV adoptedPP appVersions rs proposal
+  | -- Only validate the protocol update
+    softwareVersionUnchanged = validateProtocolUpdate
```
dnadales: Maybe it's just me, but I find this a bit hard to read. It took me a while (yeah I'm a bit slow) to realize that the paramters of `validateProtocolUpdate` where below and to the left of this symbol.

src/Cardano/Chain/Update/Validation/Registration.hs
```diff
@@ -0,0 +1,261 @@
...
+    adoptedPV
+    adoptedPP
+    rs
+    proposal
+  | -- Only validate the software update
+    protocolVersionUnchanged = validateSoftwareVersion appVersions rs proposal
+  | -- Validate both updates
+    otherwise = do
```
dnadales: I'm curious as to why we couldn't have an implementation that resembles the `UPV` rule a bit more: we call `validateProtocolUpdate`, then `validateSoftwareVersion`, and then return the new state which will be an (orthogonal) composition of the two sub-states that these two functions return.

src/Cardano/Chain/Update/Validation/Registration.hs
```diff
@@ -0,0 +1,261 @@
...
+    rs' <- validateProtocolUpdate adoptedPV adoptedPP rs proposal
+    validateSoftwareVersion appVersions rs' proposal
+ where
+  ProposalBody protocolVersion ppu softwareVersion _ _ =
+    proposalBody proposal
+
+  SoftwareVersion appName appVersion = softwareVersion
+
+  softwareVersionUnchanged           = fromMaybe False $ do
```
dnadales: Is `(== appVersion) <$> M.lookup appName appVersions` too cryptic?

src/Cardano/Chain/Update/Validation/Registration.hs
```diff
@@ -0,0 +1,261 @@
...
+  null (M.filter ((== newPV) . fst) registeredPUPs)
+    `orThrowError` RegistrationDuplicateProtocolVersion newPV
+
+  -- Check that this protocol version is a valid next version
+  pvCanFollow newPV adoptedPV
+    `orThrowError` RegistrationInvalidProtocolVersion newPV
+
+  canUpdate adoptedPP newPP proposal
+
+  pure $ rs
```
dnadales: Should we start using `pure $! ...`? Also I realized I never asked you why do we use `pure` instead of `return` :grin:  I must say I like `pure` sounds less operational :D

src/Cardano/Chain/Update/Validation/Registration.hs
dnadales: Aw, forget it, I was looking at the old version of the specs.

src/Cardano/Chain/Update/Validation/Registration.hs
```diff
@@ -0,0 +1,261 @@
...
+
+-- | Validate a protocol update
+--
+--   We check that:
+--
+--   1) The protocol update hasn't already been registered
+--   2) The protocol version is a valid next version
+--   3) The new 'ProtocolParameters' represent a valid update
+--
+validateProtocolUpdate
```
dnadales: Here I realized I choose a bad name for the rule: `validate` suggests me that the function won't change the state, which is not true. Could we rename this to something like `registerProtocolUpdate`? And if so fee free to either rename the labels in the spec, or assign an issue to me to change this.

src/Cardano/Chain/Update/Validation/Registration.hs
```diff
@@ -0,0 +1,261 @@
...
+  :: MonadError RegistrationError m
+  => ProtocolVersion
+  -> ProtocolParameters
+  -> RegistrationState
+  -> AProposal ByteString
+  -> m RegistrationState
```
dnadales: In the rules, the registration of a protocol update can only change the part of the state that contain the protocol update registrations. Could we do the same here, so that they types clearly reflect that this function won't mess with the software update part?

src/Cardano/Chain/Update/ProtocolParameterUpdate.hs
```diff
@@ -119,6 +120,27 @@ instance Bi ProtocolParameterUpdate where
       <*> decode
       <*> decode
 
+emptyProtocolParameterUpdate :: ProtocolParameterUpdate
```
nc6: More name bikeshedding here, but we have three naming conventions in this module - using the full name, as in this function, using nothing, as in `isEmpty`, and abbreviating to `PPU`. I suggest designing for qualified importing, and going with simply `isEmpty`, `empty` and `apply`. Everything else is in the module name!

src/Cardano/Chain/Update/Validation/Registration.hs
nc6: I'd go the other way, and design for qualified imports. We're in a module called `Update.Validation.Registration` - so we shouldn't need to specify that it's an update registration.

src/Cardano/Chain/Update/Validation/Registration.hs
```diff
@@ -0,0 +1,261 @@
...
+  | RegistrationMaxTxSizeTooLarge Natural Natural
+  | RegistrationProposalTooLarge Natural Natural
+  deriving (Eq, Show)
+
+type ApplicationVersions = Map ApplicationName (NumSoftwareVersion, SlotId)
+
+
+-- | Register an update proposal after verifying its signature and validating
+--   its contents
+registerProposal
```
nc6, line 65:
```suggestion
--   This corresponds to the `UPREG` rule in the ledger specification.
registerProposal
```

src/Cardano/Chain/Update/Validation/Registration.hs
nc6: You'd need a composition with `fst` in there as well...

src/Cardano/Chain/Update/Validation/Registration.hs
```diff
@@ -0,0 +1,261 @@
...
+  newPP          = PPU.applyPPU ppu adoptedPP
+
+
+-- | Check that the new 'ProtocolVersion' is a valid next version
+pvCanFollow :: ProtocolVersion -> ProtocolVersion -> Bool
+pvCanFollow newPV adoptedPV = adoptedPV < newPV && isNextVersion
+ where
+  ProtocolVersion newMajor     newMinor     _ = newPV
+  ProtocolVersion adoptedMajor adoptedMinor _ = adoptedPV
+  isNextVersion                               = case newMajor - adoptedMajor of
```
nc6: Alignment is really bad here. I spent a while trying to work out what crazy syntax
```
isNextVersion
   0 -> newMinor == adoptedMinor + 1
``` 
corresponded to, before noticing that extra stuff way off to the right on the first line.

```suggestion
  isNextVersion = case newMajor - adoptedMajor of
```

src/Cardano/Chain/Update/Validation/Registration.hs
```diff
@@ -0,0 +1,261 @@
...
+    0 -> newMinor == adoptedMinor + 1
+    1 -> newMinor == 0
+    _ -> False
+
+
+-- | Check that the new 'ProtocolParameters' represent a valid update
+--
+--   This is where we enforce constrains on how the 'ProtocolParameters' change
```
nc6:
```suggestion
--   This is where we enforce constraints on how the 'ProtocolParameters' change
```

src/Cardano/Chain/Update/Validation/Registration.hs
```diff
@@ -0,0 +1,261 @@
...
+validateSoftwareVersion
+  :: MonadError RegistrationError m
+  => ApplicationVersions
+  -> RegistrationState
+  -> AProposal ByteString
+  -> m RegistrationState
+validateSoftwareVersion appVersions rs proposal = do
+
+  -- Check that this software version isn't already registered
+  null (M.filter (== softwareVersion) registeredSUPs)
```
nc6, line 244: I was about to comment that this doesn't exactly follow the spec, and then realised that this is because the spec looks slightly wrong!

nc6 (dismissed):
Most of the changes I've suggested are minor, but I think some of the formatting is definitely difficult to understand here.

src/Cardano/Chain/Update/Validation/Registration.hs
dnadales: It is very wrong! 

@ruhatch mind to change `\vee`  by `\wedge` in `svCanFollow`?

src/Cardano/Chain/Update/Validation/Registration.hs
dnadales: Oh, that makes more sense. So do you mean changing `RegistrationError` To `Error`?

src/Cardano/Chain/Update/Validation/Registration.hs
dnadales: Is `(== appVersion) . fst <$> M.lookup appName appVersions` too cryptic?  :stuck_out_tongue:

src/Cardano/Chain/Update/Validation/Registration.hs
```diff
@@ -0,0 +1,261 @@
...
+    `orThrowError` RegistrationDuplicateSoftwareVersion softwareVersion
+
+  -- Check that this software version is a valid next version
+  svCanFollow appVersions softwareVersion
+    `orThrowError` RegistrationInvalidSoftwareVersion softwareVersion
+
+  -- Add to the list of registered software update proposals
+  pure $ rs
+    { rsSoftwareUpdateProposals = M.insert
```
dnadales: Are we ok with this idea of putting the function arguments to the left of the function? If so I'll have to get used to it :scream:

src/Cardano/Chain/Update/Validation/Registration.hs
```diff
@@ -0,0 +1,261 @@
...
+  (newMaxTxSize <= newMaxBlockSize)
+    `orThrowError` RegistrationMaxTxSizeTooLarge newMaxBlockSize newMaxTxSize
+
+  -- Check that the new script version is either the same or incremented
+  (0 <= scriptVersionDiff && scriptVersionDiff <= 1)
+    `orThrowError` RegistrationInvalidScriptVersion
+                    adoptedScriptVersion
+                    newScriptVersion
+ where
+  proposalSize :: Natural
```
dnadales, line 213: Minor thing here, but if we're aligning the `=`'s, don't we want to align the `::`?

dnadales (approved):
This looks really good.

My biggest worry is the fact that we're not reflecting in the types that `validateProtocolUpdate` can only update the protocol update part of `RegistrationState`, and `validateSoftwareUpdate` can only update the software update part of  `RegistrationState`.

Other than that I only have some minor comments regarding the formatting, and being explicit about what the values mean when we have a constructor like `XTooLarge N N`.

Oh, and don't forget to fix my error in the spec :man_facepalming: :pray:

src/Cardano/Chain/Update/ProtocolParameterUpdate.hs
dnadales: Yup. I like this idea.

src/Cardano/Chain/Update/Validation/Registration.hs
ruhatch: I think `Update.RegistrationError` is good!

ruhatch: Qualification was indeed my intention :)

src/Cardano/Chain/Update/Validation/Registration.hs
ruhatch: I think that is consistent. In the first case it is the value for the new `maxBlockSize` that is too large, whereas it is the `proposal` that is too large in the last case. Does that make sense?

src/Cardano/Chain/Update/Validation/Registration.hs
ruhatch: This is a good idea, I'll figure something out for this

src/Cardano/Chain/Update/Validation/Registration.hs
ruhatch: This should use some `TooLarge` as well

src/Cardano/Chain/Update/Validation/Registration.hs
ruhatch: Yeah, this one is a little hard - it might be better without the comments...I'll take another stab at it

src/Cardano/Chain/Update/Validation/Registration.hs
ruhatch: I did do this, but then this actually looked more complicated!

src/Cardano/Chain/Update/Validation/Registration.hs
ruhatch: I think that we can instead add strictness annotations to the datatype!

I use `pure` because at my last workplace we switched to using it instead of `return` and now it looks weird to me :P I think it's a better name and they're the same thing!

src/Cardano/Chain/Update/Validation/Registration.hs
ruhatch: Good point - naming is hard! I'll update both :)

src/Cardano/Chain/Update/Validation/Registration.hs
ruhatch: Sure, I can change it back and try and figure out a better way to do the multiplexing rule!

src/Cardano/Chain/Update/ProtocolParameterUpdate.hs
ruhatch: :+1: thanks for the suggestion!

src/Cardano/Chain/Update/Validation/Registration.hs
ruhatch: Yeah I'll see what I can do with this one

PR #287, opened 2019-01-31
Add a abstractSize function to compute the abstract size of values

### LINKED ISSUE #278: Define a function in the specs to compute the abstract size of STS signals
@nc6 Suggested that we add the following type class:

```haskell
{-# LANGUAGE DefaultSignatures #-}
{-# LANGUAGE FlexibleContexts #-}
{-# LANGUAGE FlexibleInstances #-}
{-# LANGUAGE MultiParamTypeClasses #-}
module Data.Type.Accountant where

import GHC.Generics
import Data.Reflection (Given(..))
import qualified Data.Map.Strict as Map

type AccountingMap = Map.Map String Int

class Accountant a where
  beanCount :: a -> Int
  default beanCount
    :: (Generic a, GAccountant (Rep a))
    => a -> Int
  beanCount a = gBeanCount (GHC.Generics.from a)

class GAccountant f where
  gBeanCount :: f a -> Int

instance GAccountant V1 where
  gBeanCount = const 0

instance GAccountant U1 where
  gBeanCount = const 0

instance (Given AccountingMap) => GAccountant (M1 D t f) where
  gBeanCount (M1 x) = let m = given in Map.findWithDefault 0 (moduleName ++ datatypeName) m
```

To compute the size of signals like transactions and block we'd have to provide a map where we specify the costs.

Part of #272.

dnadales:
This PR defines a function:

```haskell
abstractSize :: HasTypeReps a => AccountingMap -> a -> Size
```

where

```haskell
type Size = Int
type AccountingMap = Map TypeRep Size
```

By providing the size associated to different `TypeRep`s we can compute the total size of a value.

The typeclass `HasTypeReps` ensures that we can extract the list of all `TypeReps` found inside a given value.

Close #278.

### DIFF
--- a/specs/semantics/hs/src/Data/AbstractSize.hs
+++ b/specs/semantics/hs/src/Data/AbstractSize.hs
@@ -1,0 +1,145 @@
+{-# LANGUAGE DefaultSignatures #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE OverloadedLists #-}
+{-# LANGUAGE TypeOperators #-}
+
+-- | An approach to computing the abstract size of data using 'TypeRep'.
+--
+module Data.AbstractSize
+  ( HasTypeReps
+  , typeReps
+  , abstractSize
+  , AccountingMap
+  , Size
+  ) where
+
+import Data.Map.Strict (Map)
+import qualified Data.Map.Strict as Map
+import Data.Sequence (Seq, (<|), (><), empty)
+import Data.Typeable (TypeRep, Typeable, typeOf)
+import GHC.Generics
+  ( (:*:)((:*:))
+  , (:+:)(L1, R1)
+  , Generic
+  , K1(K1)
+  , M1(M1)
+  , Rep
+  , U1(U1)
+  , from
+  )
+
+-- | @abstractSize m a@ computes the abstract size of @a@, using the accounting
+-- map @m@. The map @m@ determines the abstract size of each 'TypeRep'
+-- contained in @a@, and this function simply adds all the individual abstract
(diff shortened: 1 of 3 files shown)


specs/semantics/hs/src/Data/AbstractSize.hs
```diff
@@ -0,0 +1,145 @@
...
+-- | @abstractSize m a@ computes the abstract size of @a@, using the accounting
+-- map @m@. The map @m@ determines the abstract size of each 'TypeRep'
+-- contained in @a@, and this function simply adds all the individual abstract
+-- sizes. To be able to extract the type representations ('TypeRep's) inside
+-- @a@, we require it to be an instance of 'HasTypeReps'.
+--
+-- Examples:
+--
+-- >>> :set -XOverloadedLists
+-- >>> abstractSize [(typeOf (undefined:: Char), 10)] 'a'
```
nc6, line 40: I really like the doctest approach, we should do this more elsewhere!

specs/semantics/hs/src/Data/AbstractSize.hs
```diff
@@ -0,0 +1,145 @@
...
+type AccountingMap = Map TypeRep Size
+
+--------------------------------------------------------------------------------
+-- HasTypeReps class
+--------------------------------------------------------------------------------
+
+-- | The 'typeReps' function retrieves all the type representations found while
+-- traversing the data given as parameter.
+--
+-- If you custom data type in an instance of 'Generics', a default
```
nc6, line 71:
```suggestion
-- If your custom data type has a 'Generic' instance, a default
```

nc6 (approved):
LGTM

Issue #255, opened 2019-01-22
Model explicit absolute/relative slot calculations


nc6:
As per https://input-output-rnd.slack.com/archives/CCZSDE005/p1548153284081900:

- When we switch from BFT to Praos, we will need to change the number of slots per epoch. Currently this is hard-coded. We also intend to switch from using relative "epoch, slot" pairs to using absolute slots. As such, we should explicitly model the calculations involved in translating between these two concepts under the assumption of a configurable number of slots per epoch.

nc6:
For the consensus layer, it would also be helpful if the ledger layer kept track of the length of historical epochs.

dnadales:
More requirements gathered from slack: I don't know whether you want to add them to the issue description @nc6 or add them as a separate issue:

- The size of an epoch should be in a map of a type like `:: Map EpochIndex EpochSize`.
- When we store blocks in the immutable store (the part of the chain that we are sure will not be rolled back), we store them per epoch. when opening the DB we want a map of precisely this form, for all epochs up to including the current one, and each time that a new epoch is started, we must be told its size

PR #290, opened 2019-02-01
 Add: Explicit epoch/slot calculations.

nc6:
Built atop #283 

- Epoch length is now a protocol parameter.
- Ledger maintains a map of epoch to epoch length both to slot/epoch
  calculations, and for the database layer.

Addresses #255

### DIFF
--- a/specs/chain/latex/blockchain-spec.tex
+++ b/specs/chain/latex/blockchain-spec.tex
@@ -73,6 +73,7 @@
 %%
 \newcommand{\Hash}{\type{Hash}}  % hashes of various things, including blocks
 \newcommand{\Slot}{\type{Slot}}
+\newcommand{\SlotCount}{\type{SlotCount}}
 \newcommand{\BlockIx}{\type{BlockIx}}
 \newcommand{\Block}{\type{Block}}
 \newcommand{\DCert}{\type{DCert}}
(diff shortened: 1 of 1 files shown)


specs/chain/latex/blockchain-spec.tex
```diff
@@ -399,24 +400,49 @@ \section{Epoch transitions}
...
-      \fun{\sepochname} & \Slot \totalf \Epoch & \text{Epoch containing this slot} \\
+    \begin{array}{rlr}
+      \fun{\sepochname} & \in \Slot \totalf (\Epoch \partialf \SlotCount) \totalf \Epoch & \text{Epoch containing this slot} \\
+    \end{array}
+  \end{equation*}
+  \begin{equation*}
+    \begin{array}{rlr}
+      \fun{\sepochname}~\var{slot}~\fun{elens} & = \begin{cases}
+        \max{\left\{ \dom{elens} \right\}} + 1 & \text{if } \sum{\left\{ \range{elens} \right\}} > \var{slot} \\
+        \max_{m}\left\{m \middle| \sum_{i < m}{\fun{elens}~i \leq \var{slot}} \right\} & \text{otherwise}
```
dnadales: Why the suffix `m` in the second `max`. Can't we just use `max`?

dnadales: I usually put the range of the summation below the sigma, and use big sigma. This is what I usually find in literature, but like every other comment of mine, feel free to ignore :)

dnadales: Is this another of my brain farts or it seems like we could have just defined:

```latex
\fun{\sepochname}~\var{slot}~\fun{elens} = \min  \{ e | slot < \sum_{e_i \leq e} \fun{elens}~i \} 
```

specs/chain/latex/blockchain-spec.tex
```diff
@@ -428,6 +454,7 @@ \section{Epoch transitions}
     \ETState =
     \left(
       \begin{array}{r@{~\in~}lr}
+        \var{elens} & \Epoch \partialf \SlotCount & \text{Historical epoch lengths} \\
```
dnadales, line 452: Do we want to start everything with capitals in the comments? I use lowercase, and I cannot recall what @JaredCorduan used for his ledger spec. We might want some consistency here.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -1233,47 +1273,47 @@ \section{Blockchain Extension}
...
-
-\begin{figure}[ht]
-  \begin{center}
-    \includegraphics[scale=0.5]{calculating-t.png}
-  \end{center}
-  \caption{Probability of generating an invalid chain for values of $t\in\left[
-      \frac{1}{4}, \frac{1}{5} \right]$.}
-  \label{fig:calculating-t}
-\end{figure}
+  \section{Calculating the $t$ parameter}
```
dnadales, line 1271: Do we need the indent here?

specs/chain/latex/blockchain-spec.tex
```diff
@@ -817,13 +850,15 @@ \subsection{Block header processing}
     \inference
     { {\left(
           \begin{array}{l}
+            \var{elens} \\
             \sepoch{s_{last}} \\
```
dnadales: I think we need to add the extra parameter to `sepoch`

specs/chain/latex/blockchain-spec.tex
```diff
@@ -1162,6 +1199,7 @@ \section{Blockchain Extension}
      {\begin{array}{c}
         \var{h'} \\
         \var{us'} \\
+        \var{elens'} \\
         \signmapname' \\
         bIx' \\
       \end{array}}
```
dnadales, line 1200: Github won't let me comment on line 1211, but I think we miss the `elens` parameter in `\sepoch` as well.

specs/chain/latex/blockchain-spec.tex
nc6: I just used auto-indent, which indented the appendices.

specs/chain/latex/blockchain-spec.tex
nc6: Yes, good catch

specs/chain/latex/blockchain-spec.tex
nc6: Looks like Jared also uses lower case, so I should adjust these. But I'll do this in a separate PR, since I've been using upper case in various places.

specs/chain/latex/blockchain-spec.tex
nc6: Yes, that would be much simpler! Thanks!

dnadales (approved):
I forgot to hit the green button.

PR #292, opened 2019-02-01
Update the delegation rules so that the delegation map is injective.

### LINKED ISSUE #257: Restrict delegation rules so that there is an isomorphism between genesis keys and delegators
Having multiple keys delegating to the same key can open up the possibility for an attack in the OBFT era: if a malicious node, with key `vk_m` gets control of a key that issued the maximum number of allowed blocks in a given window `w`, then it can prevent a node holding a key `vk` from issuing a block by simply delegating to `vk`. In such case, the current rules will count the block as being issued by all the delegators `vk`, and since `vk_m` cannot issue any more blocks, `vk` will be prevented from issuing a block.

To prevent this situation, we could strengthen the pre-conditions of the delegation rules, so that there is an isomorphism between the genesis keys and their delegates. In other words, each genesis key most delegate to exactly one key (which might be the genesis key itself).

As an added advantage of this restriction, counting the keys that signed blocks become much simpler.

dnadales:
The delegation rules have changed slightly: now the delegation relation from genesis keys to keys must be injective:

image

:arrow_up: before => now

And the corresponding prose:

image 1

Close #257.

### DIFF
--- a/specs/ledger/latex/delegation.tex
+++ b/specs/ledger/latex/delegation.tex
@@ -108,7 +108,7 @@ $\trans{sdeleg}{\wcard}$ are presented in
 
 \begin{figure}[htb]
   \begin{equation}
-    \inference[Initial-SDELEG]
+    \inference
     {
     }
     {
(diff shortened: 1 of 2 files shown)


specs/ledger/latex/delegation.tex
```diff
@@ -169,9 +169,44 @@ \section{Delegation}
...
+\item The delegating key ($\var{vk_s}$) did not activate a delegation
+  certificate in a slot greater or equal than the certificate slot ($s$). This
+  check is performed to avoid having the constraint that the delegation
+  certificates have to be activated in slot order.
+\item The key being delegated to ($\var{vk_d}$) has not been delegated by
+  another key (injectivity constraint).
+\end{itemize}
+The reason why we check that the delegation map is injective is to avoid a
+potential risk (during the OBFT era) in which a malicious node gets control of
+genesis a key $\var{vk_m}$ that issued the maximum number of blocks in a given
```
nc6:
```suggestion
a genesis key $\var{vk_m}$ that issued the maximum number of blocks in a given
```

specs/ledger/latex/delegation.tex
```diff
@@ -169,9 +169,44 @@ \section{Delegation}
...
+window. By delegating to another key $\var{vk_d}$, which was already delegated
```
nc6:
```suggestion
window. By delegating to another key $\var{vk_d}$, which was already delegated to
```

specs/ledger/latex/delegation.tex
```diff
@@ -169,9 +169,44 @@ \section{Delegation}
...
+(delegates of) genesis keys.
+
+Note also, that we could not impose the injectivity constraint in
+Rule~\ref{eq:rule:delegation-scheduling} since we do not have information about
+the delegations that will become effective. We could of course detect a
+violation in the injectivity constraint when scheduling a delegation
+certificate, but this will lead to a complex computation and larger state in
+said rule.
+
+Finally, note that we do not want to reject a scheduled delegation what would
```
nc6:
```suggestion
Finally, note that we do not want to reject a scheduled delegation that would
```

nc6 (approved):

dnadales:
Thanks for the edits @nc6 !

ruhatch (approved):
Nice!

disassembler:
bors ping

PR #293, opened 2019-02-01
Simplify block sign count rules

### LINKED ISSUE #291: Simplify the chain block signature count rules
After #257 is implemented, we can simplify the block-signing-keys counting-rules since we will have only one (genesis) signer per block. This will allow us to remove the block index from the state, and just use a sequence to keep track of the last k signers. E.g.:

image

image

Depends on #257.

Note that this is already efficiently implemented using the [`SigningHistory`](https://github.com/input-output-hk/cardano-chain/blob/3b2282983c993c450a8a57db3a136015a3003931/src/Cardano/Chain/Block/Validation.hs#L76) data structure.

dnadales:
With the changes in the delegation rules (see #292 ) we can introduce some simplifications in the counting rules. Basically I replaced these operations:

image

And this rule:

image

With this other rule:

image

I also renamed the size of the moving window from `K` to `w` to avoid confusion with the other `k`'s we have around.

Please check whether such simplification is not missing any important detail w.r.t. the current counting rules.

I should also point out that @ruhatch already made [an efficient implementation of this](https://github.com/input-output-hk/cardano-chain/blob/3b2282983c993c450a8a57db3a136015a3003931/src/Cardano/Chain/Block/Validation.hs#L76)

Closes #291 

Depends on #292

### DIFF
--- a/specs/chain/latex/blockchain-spec.tex
+++ b/specs/chain/latex/blockchain-spec.tex
@@ -93,19 +93,11 @@
 \newcommand{\bsizename}{bSize}
 \newcommand{\bhdrsizename}{bHeaderSize}
 \newcommand{\verifyname}{verify}
-\newcommand{\signmapname}{\mathcal{M}}
-\newcommand{\trimixname}{trimIx}
-\newcommand{\incixmapname}{incIxMap}
 
 \newcommand{\isebbname}{bIsEBB}
 \newcommand{\bcertsname}{bCerts}
 \newcommand{\bsigname}{bSig}
 \newcommand{\bissuername}{bIssuer}
-
-\newcommand{\qrestrname}{qRestrict}
-\newcommand{\qpopname}{qPop}
-\newcommand{\qheadname}{qHead}
-\newcommand{\qpushname}{qPush}
 
 %%
 %% Functions and relations
@@ -201,6 +193,16 @@ underlying systems and types defined will rely on definitions in that document.
   %
   Furthermore, $\epsilon$ is an identity element for sequence joining:
   $\epsilon; x = x; \epsilon = x$.
+\item[Dropping on sequences] Given a sequence $\Lambda$,
+  $\Lambda \shortdownarrow n$ is the sequence that is obtained after removing
+  (dropping) the first $n$ elements from $\Lambda$. If $n \leq 0$ then
+  $\Lambda \shortdownarrow n = \Lambda$.
+\item[Appending with a moving window] Given a sequence $\Lambda$, we define
+  $$\Lambda ;_w x \leteq (\Lambda; x) \shortdownarrow (\size{\Lambda} + 1 - w)$$
+\item[Filtering on sequences] Given a sequence $\Lambda$, and a predicate $p$
+  on the elements of $\Lambda$, $\fun{filter}~p~\Lambda$ is the sequence that
+  contains all the elements of $\Lambda$ that satisfy $p$, in the same order
+  they appear on $\Lambda$.
 \item[Option type] An option type in type $A$ is denoted as $A^? = A + 1$. The
   $A$ case corresponds to a case when there is a value of type $A$ and the $1$
   case corresponds to a case when there is no value.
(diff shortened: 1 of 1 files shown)


mdimjasevic:
Just to bring it to your attention @dnadales: you have a conflict with the `master` branch. See the status details on this page.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -615,22 +572,22 @@ \section{Counting signed blocks}
   \begin{equation*}
     \inference
     {
-      \pp{blockSignatureCountWindow} \partialf \var{K} \in pps & \pp{blockSignatureCountThreshold} \partialf \var{t} \in pps \\
-      \var{D} \leteq (\fun{dms} ~ \var{ds})^{-1} ~ \var{vk_d} & \forall \var{vk_s} \in \var{D} . \size{\signmap{\var{vk_s}}} \leq K \cdot t \\
-      \signmapname' \leteq \incixmap{\var{bIx}}{\var{D}}{(\trimix{\signmapname}{\var{bIx}})}
+      \pp{blockSignatureCountWindow} \partialf \var{w} \in pps & \pp{blockSignatureCountThreshold} \partialf \var{t} \in pps \\
+      \{\var{vk_g}\} \leteq (\fun{dms} ~ \var{ds}) \restrictrange \{\var{vk_d}\}
+      & \var{sgs'} \leteq (\var{sgs}; {vk_g}) \shortdownarrow (\size{sgs'} - w) &
```
mdimjasevic: Is this `sgs'` defined in terms of `|sgs'|`, i.e., itself?

dnadales: Oh, good catch! I just wanted to avoid the `|sgs| + 1`, but I guess this is the right thing to use. I'll change that.

specs/chain/latex/blockchain-spec.tex
```diff
@@ -642,22 +599,22 @@ \section{Counting signed blocks}
   \begin{equation*}
     \inference
     {
-      \pp{blockSignatureCountWindow} \partialf \var{K} \in pps & \pp{blockSignatureCountThreshold} \partialf \var{t} \in pps \\
-      \var{D} \leteq (\fun{dms} ~ \var{ds})^{-1} ~ \var{vk_d} & \forall \var{vk_s} \in \var{D} . \size{\signmap{\var{vk_s}}} \leq K \cdot t \\
-      \signmapname' \leteq \incixmap{\var{bIx}}{\var{D}}{(\trimix{\signmapname}{\var{bIx}})}
+      \pp{blockSignatureCountWindow} \partialf \var{w} \in pps & \pp{blockSignatureCountThreshold} \partialf \var{t} \in pps \\
+      \{\var{vk_g}\} \leteq (\fun{dms} ~ \var{ds}) \restrictrange \{\var{vk_d}\}
+      & \var{sgs'} \leteq (\var{sgs}; {vk_g}) \shortdownarrow (\size{sgs} + 1 - w)\\
```
nc6: Given we've introduced new notation for this anyway, why not make that notation instead indicate "retain only the first n elements", in which case this gets even simpler?

nc6 (approved):
One idea to make this even simpler, otherwise LGTM

nc6:
Actually, if you're renaming `K` to `w`, this should also be renamed in Appendix A for consistency

dnadales:
> Actually, if you're renaming K to w, this should also be renamed in Appendix A for consistency

Oh cr*p! I did this, but then when I rebased to master I had a conflict due to the indentation, and I simply discarded my changes :man_facepalming: I'll change it back (but keeping the indentation ;) )

specs/chain/latex/blockchain-spec.tex
dnadales: Sure, I thought of introducing the `;_k` notation for this, but then I thought it was not worth it, since we use it at one point only. But since most people might not be familiar with the drop operator anyway, I think this is a good idea. I will change it. Thanks!

PR #300, opened 2019-02-05
Change protocol-version confirmation and adoption criteria

### LINKED ISSUE #297: Change the protocol-version-candidate acceptance criteria
In the current spec, a proposal is considered as a candidate for adoption if a majority of the genesis keys issue blocks that endorse that version. This is incorrect, specially in the decentralized era, since we want to  count the proportion of nodes that are ready to adopt a given protocol version. 

So the rule below should be changed so that the number of nodes is counted in `bvs`, instead of the number of genesis keys:

image

### LINKED ISSUE #298: Update proposals should only be adopted when confirmed 2k slots before the end of the epoch
In the current spec, a proposal can be adopted once it gets enough endorsements, regardless of when these endorsements occur. However, we should make sure that a proposal is adopted if it gets confirmed `2k` slots before the end of an epoch.

dnadales:
Change `k` by `2k` slots when talking about chain stability:

image

Introduce a sequence of candidate versions to be adopted, so that at the end of an epoch we can determine whether we can adopt a new protocol-version:

image

`fads` has the following meaning:

image

The endorsement rules where changed, so that the `fads` sequence is updated when new endorsements are registered (and satisfy certain conditions):

image

This had also a minor impact on the chain interface rules for endorsement:

image

And the protocol-adoption rule on epoch change now looks at the last element of `fads` restricted to slots not newer that `2k`:

image

The timeline that exemplifies an update-proposal confirmed too late was adapted accordingly:

image

And as required by #297, we now count the endorsements per-node, instead of per-genesis key:

image

# Other changes:

Clarify what `upAdptThd` should relate to existing protocol parameters (@ruhatch we have to check this):

image

Closes #297.
Closes #298.

### DIFF
--- a/specs/ledger/latex/blockchain-interface.tex
+++ b/specs/ledger/latex/blockchain-interface.tex
@@ -190,8 +190,8 @@ labels have the following meaning:
         \var{e_p} & \Epoch & \text{previously seen epoch}\\
         (\var{pv}, \var{pps}) & \ProtVer \times \PPMMap
         & \text{current protocol information}\\
-        (\var{pv_c}, \var{pps_c}) & \ProtVer \times \PPMMap
-        & \text{candidate protocol information}\\
+        \var{fads} & \seqof{(\Slot \times (\ProtVer \times \PPMMap))}
+        & \text{future protocol version adoptions}\\
         \var{avs} & \ApName \mapsto (\ApVer \times \Slot)
         & \text{application versions}\\
         \var{rpus} & \UPropId \mapsto (\ProtVer \times \PPMMap)
@@ -272,7 +272,7 @@ labels have the following meaning:
           \begin{array}{l}
             \var{e_p}\\
             (\var{pv}, \var{pps})\\
-            (\var{pv_c}, \var{pps_c})\\
+            \var{fads}\\
             \var{avs}\\
             \var{rpus}\\
             \var{raus}\\
@@ -289,7 +289,7 @@ labels have the following meaning:
           \begin{array}{l}
             \var{e_p}\\
             (\var{pv}, \var{pps})\\
-            (\var{pv_c}, \var{pps_c})\\
+            \var{fads}\\
             \var{avs}\\
             \var{rpus'}\\
             \var{raus'}\\
@@ -369,7 +369,7 @@ updated.
           \begin{array}{l}
             \var{e_p}\\
             (\var{pv}, \var{pps})\\
-            (\var{pv_c}, \var{pps_c})\\
+            \var{fads}\\
             \var{avs}\\
             \var{rpus}\\
             \var{raus}\\
@@ -386,7 +386,7 @@ updated.
           \begin{array}{l}
             \var{e_p}\\
             (\var{pv}, \var{pps})\\
-            (\var{pv_c}, \var{pps_c})\\
+            \var{fads}\\
             \var{avs} \unionoverride \var{avs_{new}}\\
             \var{rpus}\\
             \dom~ \var{cps} \subtractdom \var{raus}\\
@@ -484,7 +484,6 @@ the end on an epoch.
           k\\
           s_n\\
           (\var{pv}, \var{pps})\\
-          \var{dms}\\
           \var{cps}\\
           \var{rpus}
         \end{array}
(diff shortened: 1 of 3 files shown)


specs/ledger/latex/blockchain-interface.tex
```diff
@@ -612,6 +627,41 @@ \subsection{Update-proposals interface}
...
+        \right)
+      }
+    }
+  \end{equation}
+  \nextdef
+  \begin{equation}
+    \label{eq:rule:pvbump-change-epoch-only}
+    \inference
+    {
+      \epsilon = [.., s_n - 2 \cdot k] \restrictdom \var{fads} &  e_p < e_n
```
nc6: I guess we do have assignment now, but even so I wonder whether it would be more understandable if this equality were the other way around.

specs/ledger/latex/blockchain-interface.tex
```diff
@@ -622,16 +672,21 @@ \subsection{Update-proposals interface}
     \label{eq:rule:pvbump-change}
     \inference
     {
-      e_p < e_n
+      \wcard ; (s_c, (\var{pv_c}, \var{pps_c})) \leteq [.., s_n - 2 \cdot k] \restrictdom \var{fads}
```
nc6: The binding `s_c` isn't used here, maybe use another wildcard?

specs/ledger/latex/update-mechanism.tex
```diff
@@ -1016,29 +1016,33 @@ \subsection{Update-proposal endorsement}
 rule by $\var{bv}$:
 \begin{itemize}
 \item The set $\var{bvs}$, containing which genesis keys are ready to adopt a
```
nc6:
```suggestion
\item The set $\var{bvs}$, containing which block issuers are ready to adopt a
```

specs/ledger/latex/update-mechanism.tex
```diff
@@ -1016,29 +1016,33 @@ \subsection{Update-proposal endorsement}
...
-  next epoch. Figure~\ref{fig:up-confirmed-too-late} shows an example of a
-  proposal being confirmed too late in an epoch, where it is not possible to
-  get the enough endorsements (4 out of 7) in the remaining window. In this
-  Figure we take $k = 4$.
+  given protocol version, is updated to reflect that the block issuer
+  (identified by its verifying key $\var{vk}$) is ready to upgrade to
+  $\var{bv}$. Given a pair $(\var{pv}, ~\var{vk}) \in \var{bvs}$, we say that
+  (the owner of) key $\var{vk}$ endorses the (proposed) protocol version
+  $\var{pv}$.
+\item If there is a significant number of blocks signed with $\var{bv}$
```
nc6:
```suggestion
\item If there are a significant number of blocks signed with $\var{bv}$
```

specs/ledger/latex/update-mechanism.tex
```diff
@@ -1056,74 +1060,70 @@ \subsection{Update-proposal endorsement}
...
     {
-      (\var{bv} \leq \var{pv_c} \vee \var{pid} \notin \dom~(\var{cps} \restrictrange [.., s_n - k]))
-      &
-      {
-        \begin{array}{l}
-          \var{dms}
-        \end{array}
-      }
-      \vdash
+      \wcard ; (s_c, (\var{pv_c}, \wcard)) \leteq \var{fads} & \var{bv} \leq \var{pv_c}
```
nc6: Again, `s_c` not used here.

nc6 (approved):
Looks good, few minor comments.

specs/ledger/latex/blockchain-interface.tex
dnadales: What's wrong with my [Yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions). I was afraid of this. I'll change it.

specs/ledger/latex/update-mechanism.tex
dnadales: Missed that one. Thanks!

specs/ledger/latex/update-mechanism.tex
dnadales: Ugh! Thanks to this I realized I had a copy+paste error in this rule: I shouldn't concatenate nothing to `fads`, otherwise the invariant is not maintained!

dnadales:
Thanks for the review @nc6 I incorporated your suggestions!

PR #301, opened 2019-02-05
Rebased delegation generators

nc6:
This is a rebased version of #271 which also moves to using `AbstractSize`


pkgs.nix
```diff
@@ -9,6 +9,17 @@ import nixpkgs {
           ghc861 = pkgs.haskell.packages.ghc861.override {
             overrides = self: super: {
               basement = self.callHackage "basement" "0.0.8" {};
+              aeson = self.callHackage "aeson" "1.3.1.1" {};
+              nonempty-containers = self.callCabal2nix "nonempty-containers"
+                              (pkgs.fetchFromGitHub { owner = "mstksg";
+                                                      repo = "nonempty-containers";
+                                                      rev = "27fb9d78269be1b31351fff8ac24268f6a107964";
+                                                      sha256 = "1bj95siah7fv4wapzk4p28qx5yvadznc78cnfqcr3whmi1g03vna"; }) {};
+              cardano-prelude = self.callCabal2nix "cardano-prelude"
```
dnadales: I guess we don't need `cardano-prelude` anymore (since we don't use `heap-words`). And I'm not sure we need `nonempty-containers`.

specs/stack.yaml
```diff
@@ -1,4 +1,4 @@
-resolver: lts-13.0
+resolver: https://raw.githubusercontent.com/input-output-hk/cardano-prelude/f2802079ba2c07c4d408e3c0aa000fc363792398/snapshot.yaml
```
dnadales, line 1: Funny, I thought we already pinned the resolver.

specs/chain/hs/src/Cardano/Spec/Chain/STS/Block.hs
```diff
@@ -0,0 +1,94 @@
...
+  } deriving (Eq, Show, Generic)
+
+instance HasTypeReps Block
+makeLenses ''Block
+
+-- | Compute the size (in words) that a block takes.
+bSize :: Block -> Natural
+bSize = fromIntegral . abstractSize acctMap
+  where
+    acctMap = []
```
dnadales, line 71: Oh, don't we want to tally up something here?

dnadales (approved):
Thanks for rebasing and incorporating `AbstractSize` Nick!

specs/chain/hs/src/Cardano/Spec/Chain/STS/Block.hs
nc6: Yes, but I don't know what it is, and I think this is the wrong PR to decide this in

nc6:
bors retry

PR #302, opened 2019-02-05
Add `cardano-prelude` as a submodule

mhuesch:
This is an alternative to `stack.yaml` commit pinning.  I've found
it to be smoother for development. The pain point this addresses is
making and testing local changes (e.g. changes in `cardano-prelude`
which are used/tested from `cardano-chain`) - with the `stack.yaml`
-pinning style, changes must be pushed to a remote or done through
an include of a local git repo. The submodules style lets you make
local modifications, test that they work, then commit and push to
the submodule remote (`cardano-prelude` in this case) before bumping
the submodule tracking hash from the parent (`cardano-chain` in this
case), committing that, and pushing it to the parent's remote.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "cardano-mainnet-mirror"]
 	path = cardano-mainnet-mirror
 	url = https://github.com/input-output-hk/cardano-mainnet-mirror.git
+[submodule "submodules/cardano-prelude"]
+	path = submodules/cardano-prelude
+	url = git@github.com:input-output-hk/cardano-prelude.git
--- a/stack.yaml
+++ b/stack.yaml
@@ -7,12 +7,6 @@ packages:
   - specs/semantics/hs
 
 extra-deps:
-  - git: https://github.com/input-output-hk/cardano-prelude
-    commit: 7f8544ae9e162b6664f07093227edb148f1c8a5b
-    subdirs:
-      - .
-      - test
-
   - git: https://github.com/input-output-hk/cardano-chain
     commit: 4d9eec080374179bf15bf9c4fca09cc7d73e5f53
     subdirs:


erikd:
> The pain point this addresses is making and testing local changes (e.g. changes in cardano-prelude
which are used/tested from cardano-chain)

Is this manual testing to make sure the changes to `cardano-prelude` work in `cardano-chain`?

The reason I ask is because, i would all automated tests for `cardano-prelude` to be in  `cardano-prelude`.

nc6:
I would prefer not to merge this, having seen too many annoying failures because submodules get out of sync, or are pointing to the wrong place, or similar.

mhuesch:
> > The pain point this addresses is making and testing local changes (e.g. changes in cardano-prelude
> > which are used/tested from cardano-chain)
> 
> Is this manual testing to make sure the changes to `cardano-prelude` work in `cardano-chain`?
> 
> The reason I ask is because, i would all automated tests for `cardano-prelude` to be in `cardano-prelude`.

@erikd: ah - my use of "testing" is perhaps confusing. I meant that I found an issue while modifying code in `cardano-chain` and realized it was due to `cardano-prelude`, but it wasn't obvious to me how to make local changes to `cardano-prelude` and test their effects from `cardano-chain` since we've pinned a commit version on the remote. The change in question is https://github.com/input-output-hk/cardano-prelude/pull/29.

mhuesch:
> I would prefer not to merge this, having seen too many annoying failures because submodules get out of sync, or are pointing to the wrong place, or similar.

@nc6: definitely reasonable - feel free to close this. Since I've used submodules in the past I found it the most natural way to test the effects of the changes of https://github.com/input-output-hk/cardano-prelude/pull/29 in `cardano-chain` and then PR'd in case others would find it useful. Sorry if I was overzealous 🙂

erikd:
> I would prefer not to merge this, having seen too many annoying failures because submodules get out of sync, or are pointing to the wrong place, or similar.

I think this is due to a lack of familiarity with submodules. Once people become familiar with submodules, handling issues like these simply become part of the normal workflow.

@mhuesch I think the way you would test this the stack way is to fix `cardano-prelude` and push the branch to github (so the repo URL is the same) and then edit the `stack.yaml` file and change the hash for `cardano-prelude` to point to your branch. Obviously this is not something that can be PRed until the `cardano-prelude` patch is merged, and then you will probably need to update the hash to point to the merged commit.

mhuesch:
Closing this in favor of documenting the `stack-local.yaml` approach described in https://github.com/input-output-hk/cardano-chain/issues/309.

PR #307, opened 2019-02-06
[WIP] Make the Block type polymorphic

### LINKED ISSUE #304: Make the Block type polymorphic
Based on a discussion with @edsko, @dnadales and @nc6, the `Block` type should be made polymorphic.

To integrate the `cs-blockchain` specification implementation into `ouroboros-consensus` (see https://github.com/input-output-hk/ouroboros-network/issues/168), it would be best if the `Types.Block` type was made polymorphic. In particular, to match its LaTeX counterpart, there should be a Haskell type class that defines projection functions from a block type to fields in it. It should also be parameterised over a cryptographic parameter that determines the signing algorithm.

mdimjasevic:
@edsko, did you have something like this in mind? In particular, compare:

1. The existing `Block` type: https://github.com/input-output-hk/cardano-chain/blob/mdimjasevic/304-polymorphic-block/specs/chain/hs/src/Cardano/Spec/Chain/STS/Block.hs, and
2. The type class for blocks that this pull request adds: https://github.com/input-output-hk/cardano-chain/blob/mdimjasevic/304-polymorphic-block/specs/chain/hs/src/Cardano/Spec/Chain/STS/BlockC.hs

Then the way I see it, the `Block` type would implement the `BlockC` class (I'll rename the class later).

Closes #304.


edsko:
Yes, something like that. A few comments:

* From the perspective of the consensus layer, your `BlockHeaderC` class 
  corresponds to our `HasPayload` class (remember that we use "payload" to 
  mean "the stuff that the protocol requires in a block" -- suggestions for
  better naming welcome), with a payload of a `VKey` and `Sig VKey`, which 
  might indeed be a reasonable payload for PermBFT. `HasPayload` inherits from
  `HasHeader` from the network layer which gives you the hash and the slot.

  Of course that doesn't necessarily help you too too much since you can't
  introduce a dependency on the consensus layer; but just saying, it fits.
  
* `BlockBodyC` I think is relevant for the application of a block to the ledger
  state, but should not be relevant I think for the protocol itself right?
  There should be no need for the protocol to know anything about delegation
  certificates (maybe I'm wrong here).

* `BlockC b` doesn't make too much sense: those `h` and `bb` are universally
  quantified, not existentially, which is I think what you had in mind. But
  I suspect there might not actually a need for this at all. Maybe just

  `class (BlockHeaderC b, BlockBodyC b) => BlockC b`

  If so, then indeed just not defining `BlockC` at all and just sticking with
  `BlockHeaderC` and `BlockBodyC`  might suffice.

mdimjasevic:
Thank you for your comments, @edsko!

> Yes, something like that. A few comments:
> 
> * `BlockBodyC` I think is relevant for the application of a block to the ledger
>   state, but should not be relevant I think for the protocol itself right?
>   There should be no need for the protocol to know anything about delegation
>   certificates (maybe I'm wrong here).

Looking from the blockchain layer perspective, how should it allow the ledger layer to process delegation certificates if there is no requirement that the block body has the certificates?


> * `BlockC b` doesn't make too much sense: those `h` and `bb` are universally
>   quantified, not existentially, which is I think what you had in mind. But
>   I suspect there might not actually a need for this at all. Maybe just
>   `class (BlockHeaderC b, BlockBodyC b) => BlockC b`
>   If so, then indeed just not defining `BlockC` at all and just sticking with
>   `BlockHeaderC` and `BlockBodyC`  might suffice.

I'm still trying to understand this, but from how I see it so far, we need something that says "this is a block". If we drop the `BlockC` class, what will tell that?

Next, if I change the `BlockC` class to `class (BlockHeaderC b, BlockBodyC b) => BlockC b`, how do I extract the block's body, assuming a class has accessor functions for smaller components only?

edsko:
> Looking from the blockchain layer perspective, how should it allow the ledger layer to process delegation certificates if there is no requirement that the block body has the certificates?

The _ledger layer_ needs to know about delegation certificates, but the _consensus layer_ does not: it just needs to know the set of keys that can sign blocks (i.e., with delegation certifiates _applied_ if you will).

> I'm still trying to understand this, but from how I see it so far, we need something that says "this is a block". If we drop the `BlockC` class, what will tell that? How do I extract the block's body, assuming a class has accessor functions for smaller components only?

If you really need to extract the block body, you'll need to introduce some kind of type parameter to describe what the block body looks like. But it might not be necessary: you just pass the whole block along, and whether stuff comes from the header or the body may not matter.

mdimjasevic:
> The _ledger layer_ needs to know about delegation certificates, but the _consensus layer_ does not: it just needs to know the set of keys that can sign blocks (i.e., with delegation certifiates _applied_ if you will).

But here we're in the blockchain layer's implementation, not the consensus layer's, so I don't see how to implement the blockchain layer without having a constraint on the block body.

edsko:
Sure sure. The code that updates the ledger state will certainly need it. But you shouldn't need it in your PermBFT defintiion itself.

dnadales:
BTW, we can start by defining only the type classes that are needed for the consensus-layer, and define the other type classes only if they're needed. I prefer to minimize work, if that's possible/sensible ;)

edsko:
> BTW, we can start by defining only the type classes that are needed for the consensus-layer, and define the other type classes only if they're needed. I prefer to minimize work, if that's possible/sensible ;)

That's not really possible though is it? Unless you at the same time also want to decouple the ledger completely, running PermBFT with our mock ledger. But if you don't want to do that, you need to use your ledger, but then that would mean we need to b able to _update_ your ledger for incoming blocks, and if those incoming blocks aren't your `Block` type anymore, the update fgunction itself will need to be generalized also.

mdimjasevic:
This is still WIP, but @nc6  and @dnadales I'd like to ask for your input before proceeding.

I looked at the `PBFT` and `BHEAD` rules in the chain spec and this is what I came up with. In one of them only the block slot is used, while in the other one the rest of the header information plus the slot are used. So I restructured the `BlockHeader` type accordingly.

Does that look OK?

specs/chain/hs/src/Cardano/Spec/Chain/STS/Block.hs
```diff
@@ -17,18 +18,39 @@ import Ledger.Core
...
-  , _bIssuer :: !VKey
-    -- ^ Block issuer.
-  , _bSig :: !(Sig VKey)
-    -- ^ Signature of the block by its issuer.
+-- TODO(md): Define this type by seeing what is needed here in the
+-- blockchain LaTeX document
+data BHToSign deriving (Eq, Show, Generic)
+
+data BlockProtocolHeader
+  = MkBlockProtocolHeader
```
dnadales, line 26: Small thing: I'm fine with using `Mk` for constructors, but maybe we need to agree on that (or add an issue to write this somewhere) so that everyone (within our team at least) uses the same conventions.

specs/chain/hs/src/Cardano/Spec/Chain/STS/Block.hs
```diff
@@ -17,18 +18,39 @@ import Ledger.Core
...
+  {
+    -- | Hash of the previous block header, or 'genesisHash' in case of
+    -- the first block in a chain.
+    _bhPrevHash :: Hash
+    -- | Hash of the block.
+  , _bhHash     :: Hash
+    -- | Block issuer.
+  , _bhIssuer   :: VKey
+    -- | Part of the block header which must be signed.
+  , _bhToSign   :: BHToSign
```
dnadales, line 36: Shouldn't `bhToSign` be a function from `BlockProtocolHeader`? I'd assume that it will give a subset of the fields of the header...

specs/chain/hs/src/Cardano/Spec/Chain/STS/Block.hs
```diff
@@ -17,18 +18,39 @@ import Ledger.Core
...
+    -- | Signature of the block by its issuer.
+  , _bhSig      :: !(Sig VKey)
```
dnadales: Any reason why some fields are strict and some aren't?

dnadales (approved):
LGTM.

specs/chain/hs/src/Cardano/Spec/Chain/STS/Block.hs
mdimjasevic: That's something I wasn't sure of. It's referenced in only one place in the whole LaTeX specification (Figure 13) and that's it.

specs/chain/hs/src/Cardano/Spec/Chain/STS/Block.hs
mdimjasevic: Oh, that's my mistake. I was copy-pasting code from different sources: some had it strict and other not. What should be the case with fields here?

dnadales: Maybe @nc6 has a better answer, but I guess in most cases in our spec it won't hurt having the fields being strict by default.

mdimjasevic:
Closing in favor of https://github.com/input-output-hk/fm-ledger-rules/pull/239.

PR #308, opened 2019-02-06
[DO NOT MERGE] Add some scaffolding for testing the validator with the generated chains

dnadales:
A preliminary PR to get some early feedback.


test/Test/Cardano/Chain/Block/Validation/Spec.hs
```diff
@@ -0,0 +1,68 @@
...
+          Concrete.updateChain config cst (interpret cst ast)
+
+        initSt = either (panic . show) identity $
+          Concrete.initialChainValidationState config
+
+    -- TODO: this will be replaced with the actual interpreter.
+    -- TODO: the interpreter might need some other parameters, like the
+    -- previous block.
+    interpret
+      :: Concrete.ChainValidationState
```
ruhatch: I guess you could just take the previous hash from the `ChainValidationState` :man_shrugging:

test/Test/Cardano/Chain/Block/Validation/Spec.hs
```diff
@@ -0,0 +1,68 @@
...
+prop_generatedChainsAreValidated = property $ do
+  -- TODO: remove this duplication - see test/Test/Cardano/Chain/Block/Validation.hs
+  -- Get the 'Genesis.Config' from the mainnet genesis JSON
+  config <- readMainetCfg
+  forAll trace >>= passConcreteValidation config
+  where
+    passConcreteValidation :: MonadTest m => Genesis.Config -> Trace CHAIN -> m ()
+    passConcreteValidation config tr = do
+      let
+        res = foldM interpretAndUpdate initSt $ traceSignals OldestFirst tr
```
ruhatch: :+1:

test/Test/Cardano/Chain/Block/Validation/Spec.hs
```diff
@@ -0,0 +1,68 @@
...
+import Test.Cardano.Chain.Common.Utils (readMainetCfg)
+
+tests :: IO Bool
+tests = checkParallel $$discover
+
+-- | Every abstract chain that was generated according to the inference rules,
+-- after being interpreted must be validated by the concrete block validator.
+prop_generatedChainsAreValidated :: Property
+prop_generatedChainsAreValidated = property $ do
+  -- TODO: remove this duplication - see test/Test/Cardano/Chain/Block/Validation.hs
```
ruhatch: Did you remove this duplication already with `Utils`?

test/Test/Cardano/Chain/Block/Validation/Spec.hs
```diff
@@ -0,0 +1,68 @@
...
+  -- Get the 'Genesis.Config' from the mainnet genesis JSON
+  config <- readMainetCfg
+  forAll trace >>= passConcreteValidation config
+  where
+    passConcreteValidation :: MonadTest m => Genesis.Config -> Trace CHAIN -> m ()
+    passConcreteValidation config tr = do
+      let
+        res = foldM interpretAndUpdate initSt $ traceSignals OldestFirst tr
+      either (panic . show) (const $ pure ()) res
```
ruhatch:
```suggestion
      void $ evalEither res
```

ruhatch: I think that's more idiomatic for `hedgehog`

test/Test/Cardano/Chain/Block/Validation/Spec.hs
```diff
@@ -0,0 +1,68 @@
...
+      where
+
+        interpretAndUpdate
+          :: Concrete.ChainValidationState
+          -> Abstract.Block
+          -> Either Concrete.ChainValidationError Concrete.ChainValidationState
+        interpretAndUpdate cst ast =
+          Concrete.updateChain config cst (interpret cst ast)
+
+        initSt = either (panic . show) identity $
```
ruhatch: I guess you could use `evalEither` here too and then put this in the `do` block. Not sure if that would cause this code to run more times though...

test/Test/Cardano/Chain/Common/Utils.hs
```diff
@@ -0,0 +1,20 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Test.Cardano.Chain.Common.Utils
```
ruhatch: I think `Test.Cardano.Chain.Config` might be a better location

ruhatch (changes requested):
Looking good so far!

test/Test/Cardano/Chain/Block/Validation/Spec.hs
dnadales: Sure, maybe we need to relax the pre-conditions on this function (now called `elaborate` :)).

test/Test/Cardano/Chain/Block/Validation/Spec.hs
dnadales: Yes. I just forgot to remove the comment :man_facepalming:

test/Test/Cardano/Chain/Block/Validation/Spec.hs
dnadales: Yes. I was planing on getting read of this "panic show" pattern, so thanks for pointing me to `evalEither`.

dnadales:
Thanks for the first batch of reviews @ruhatch !

ruhatch (commented):
Maybe we should add `stack-local.yaml` to the `.gitignore` too :)

dnadales:
> Maybe we should add `stack-local.yaml` to the `.gitignore` too :)

Nice catch. Thanks @ruhatch !

dnadales:
I'm closing this one and making a new one.

PR #310, opened 2019-02-07
[#119] Add JSON canonical tests from Byron release

### LINKED ISSUE #119: Implement canonical JSON tests from Byron release
The Byron release had canonical JSON tests that need to be brought over to `cardano-chain` to ensure coherency of the canonical JSON format.

Jimbo4350:
Closes #119

### DIFF
--- a/test/Test/Cardano/Chain/Slotting/Gen.hs
+++ b/test/Test/Cardano/Chain/Slotting/Gen.hs
@@ -42,6 +42,7 @@ import Cardano.Crypto (ProtocolMagicId)
 import Test.Cardano.Crypto.Gen (genProtocolMagicId)
 
 
+
 genEpochIndex :: Gen EpochIndex
 genEpochIndex = EpochIndex <$> Gen.word64 Range.constantBounded
 
--- a/cardano-chain.cabal
+++ b/cardano-chain.cabal
@@ -199,6 +199,7 @@ test-suite cardano-chain-test
   build-depends:       base
                      , base16-bytestring
                      , bytestring
+                     , canonical-json
                      , cardano-binary
                      , cardano-binary-test
                      , cardano-chain
--- a/stack.yaml
+++ b/stack.yaml
@@ -8,7 +8,7 @@ packages:
 
 extra-deps:
   - git: https://github.com/input-output-hk/cardano-prelude
-    commit: 7f8544ae9e162b6664f07093227edb148f1c8a5b
+    commit: 7161436bd31195b2ddfa7d359cd99f0fe53c881c
     subdirs:
       - .
       - test
--- a/test/Test/Cardano/Chain/Common/Gen.hs
+++ b/test/Test/Cardano/Chain/Common/Gen.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE LambdaCase        #-}
+{-# LANGUAGE NumDecimals       #-}
 {-# LANGUAGE OverloadedStrings #-}
 
 module Test.Cardano.Chain.Common.Gen
@@ -8,6 +9,7 @@ module Test.Cardano.Chain.Common.Gen
   , genAddrSpendingData
   , genAddrStakeDistribution
   , genBlockCount
+  , genCanonicalTxFeePolicy
   , genChainDifficulty
   , genLovelace
   , genLovelacePortion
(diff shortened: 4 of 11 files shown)


test/Test/Cardano/Chain/Slotting/Gen.hs
```diff
@@ -42,6 +44,10 @@ import Cardano.Crypto (ProtocolMagicId)
 import Test.Cardano.Crypto.Gen (genProtocolMagicId)
 
 
+genCanonicalEpochIndex :: Gen EpochIndex
+genCanonicalEpochIndex =
+  EpochIndex <$> Gen.word64 (Range.constant 0 maxCanonicalLovelaceVal)
```
mhuesch: What is the relation between `maxCanonicalLovelaceVal` and `EpochIndex`? Is it just being used as `maxCanonicalJsonIntegerGiga` or something (because it is ```int54Max `div` 1e9```)? If so maybe we could rename for clarity.

mhuesch (changes requested):
Looks good! Minor question before I approve.

mhuesch:
@Jimbo4350 Also, will other JSON instances be brought over from `cardano-sl` to `cardano-chain` later, or does this represent all of them? I made up two cursory lists of types with CanonicalJson instances by grepping in the two repo's - are some of those types missing or do they represent refactoring / renaming where the old types are no longer needed?

[cardano-chain_canonical_json.txt](https://github.com/input-output-hk/cardano-chain/files/2842192/cardano-chain_canonical_json.txt)
[cardano-sl_canonical_json.txt](https://github.com/input-output-hk/cardano-chain/files/2842193/cardano-sl_canonical_json.txt)

test/Test/Cardano/Chain/Slotting/Gen.hs
Jimbo4350: We are actually removing this. Canonical JSON uses 54-bit integers. My understanding is @ruhatch wrote instances to avoid overflow errors (e.g https://github.com/input-output-hk/cardano-prelude/blob/master/src/Cardano/Prelude/Json/Canonical.hs#L104)

However a couple of the types' Canonical JSON instances are converting them to `Int54`s. In my debugging I restricted `genCanonicalEpochIndex` to only produce indicies within the bounds of `Int54`. We decided instead of having a `genCanonicalEpochIndex`, we would modify the types that have Canonical instances that demand an `Int54`. Is this clear?

mhuesch: > We are actually removing this. Canonical JSON uses 54-bit integers. My understanding is @ruhatch wrote instances to avoid overflow errors (e.g https://github.com/input-output-hk/cardano-prelude/blob/master/src/Cardano/Prelude/Json/Canonical.hs#L104)

Ok, right. Those instances should either wrap the JSNum's into the smaller types (Word32/16, Int32) or promote them (no wrapping since the # of bits is greater than 54) into Word64, Integer, Natural.

> However a couple of the types' Canonical JSON instances are converting them to `Int54`s. In my debugging I restricted `genCanonicalEpochIndex` to only produce indicies within the bounds of `Int54`. We decided instead of having a `genCanonicalEpochIndex`, we would modify the types that have Canonical instances that demand an `Int54`. Is this clear?

I'm a little confused by the wording here - "We decided instead of having a `genCanonicalEpochIndex`" - but we have that, don't we?

mhuesch: I've been able to run `stack test cardano-chain` successfully with this patch applied:
```
diff --git a/test/Test/Cardano/Chain/Slotting/Gen.hs b/test/Test/Cardano/Chain/Slotting/Gen.hs
index c2e2efc..3b538a5 100644
--- a/test/Test/Cardano/Chain/Slotting/Gen.hs
+++ b/test/Test/Cardano/Chain/Slotting/Gen.hs
@@ -46,7 +46,7 @@ import Test.Cardano.Crypto.Gen (genProtocolMagicId)
 
 genCanonicalEpochIndex :: Gen EpochIndex
 genCanonicalEpochIndex =
-  EpochIndex <$> Gen.word64 (Range.constant 0 maxCanonicalLovelaceVal)
+  EpochIndex <$> Gen.word64 (Range.constant 0 (100000000 * maxCanonicalLovelaceVal))
 
 genEpochIndex :: Gen EpochIndex
 genEpochIndex = EpochIndex <$> Gen.word64 Range.constantBounded
```
that is, multiplying the supposed max value by 10^8. Oddly, if I multiply by 10^9 I observe failures. So something odd is happening here with the Int54 max value, but I'm not sure what. It seems we have more room at the top end than your original generator was assuming, though. Unless I'm misunderstanding...

Jimbo4350:
> @Jimbo4350 Also, will other JSON instances be brought over from `cardano-sl` to `cardano-chain` later, or does this represent all of them? I made up two cursory lists of types with CanonicalJson instances by grepping in the two repo's - are some of those types missing or do they represent refactoring / renaming where the old types are no longer needed?
> 
> [cardano-chain_canonical_json.txt](https://github.com/input-output-hk/cardano-chain/files/2842192/cardano-chain_canonical_json.txt)
> [cardano-sl_canonical_json.txt](https://github.com/input-output-hk/cardano-chain/files/2842193/cardano-sl_canonical_json.txt)

These are the canonical JSON tests in Byron:
https://github.com/input-output-hk/cardano-sl/blob/ccca8b346413f12e9fad8cab4c855e69669a2151/lib/test/Test/Pos/Genesis/CanonicalSpec.hs

Golden canonical JSON test for `GenesisData`:
https://github.com/input-output-hk/cardano-sl/blob/577c82944042ca46a916ba86683b5db149acd49e/chain/test/Test/Pos/Chain/Genesis/Json.hs#L141

I implemented the tests above. The other tests I implemented were mainly for debugging purposes. Most of those types in your grep search are subtypes of `GenesisData` so we should be ok. If you think I missed anything please let me know.

mhuesch:
@Jimbo4350 Thanks! Sorry for being hard-nosed but wanted to make sure I'm not missing something. Seems like the list of types I see in `cardano-sl` which aren't reflected verbatim in `cardano-chain` is
```
GenesisProtocolConstants
SharedSeed
Timestamp
VssCertificate
VssCertificatesMap
VssMaxTTL
VssMinTTL
ProxySKHeavy
```
I assume those just have different names in `cardano-chain` or have been revised into different types. Looks good to me then 👍

mhuesch (approved):
LGTM 🚢 :shipit: 🔥

PR #318, opened 2019-02-12
[#186/#187] Implement update voting and confirmation rules

### LINKED ISSUE #186: Implement update voting rules

### LINKED ISSUE #187: Implement update confirmation rules

ruhatch:
Closes #186 
Closes #187

### DIFF
--- a/src/Cardano/Chain/Update/Validation/Voting.hs
+++ b/src/Cardano/Chain/Update/Validation/Voting.hs
@@ -1,0 +1,141 @@
+{-# LANGUAGE FlexibleContexts #-}
+
+-- | Validation rules for registering votes and confirming proposals
+--
+--   This is an implementation of the rules defined in the Byron ledger
+--   specification
+module Cardano.Chain.Update.Validation.Voting
+  ( VotingEnvironment(..)
+  , VoteRegistrationEnvironment(..)
+  , VotingState(..)
+  , VotingError(..)
+  , registerVoteWithConfirmation
+  )
+where
+
+import Cardano.Prelude
+
+import qualified Data.Map as M
+import qualified Data.Set as Set
+
+import Cardano.Chain.Common
+import Cardano.Chain.Slotting
+import Cardano.Chain.Update.ProtocolParameters
+import Cardano.Chain.Update.Vote
+import Cardano.Crypto
+
+
+-- | Environment used to register votes and confirm proposals
+data VotingEnvironment = VotingEnvironment
+  { veCurrentSlot                   :: SlotId
+  , veProtocolParameters            :: ProtocolParameters
+  , veVotingRegistrationEnvironment :: VoteRegistrationEnvironment
+  }
+
+-- | Environment required to validate and register a vote
+data VoteRegistrationEnvironment = VoteRegistrationEnvironment
+ { vreRegisteredUpdateProposal :: Set UpId
+ , vreDelegationMap            :: Map StakeholderId StakeholderId
+ }
+
+-- | VotingState keeps track of registered votes and confirmed proposals
+data VotingState = VotingState
+  { vsVotes              :: RegisteredVotes
+  , vsConfirmedProposals :: Map UpId SlotId
+  }
+
+type RegisteredVotes = Map UpId (Set StakeholderId)
+
+-- | VotingError captures the ways in which vote registration could fail
+data VotingError
+  = VotingInvalidSignature
+  | VotingProposalNotRegistered UpId
+  | VotingVoterNotDelegate StakeholderId
+
+
+-- | Register a vote and confirm the corresponding proposal if it passes the
+--   voting threshold. This corresponds to the `UPVOTE` rules in the spec.
+registerVoteWithConfirmation
+  :: MonadError VotingError m
+  => ProtocolMagicId
+  -> VotingEnvironment
+  -> VotingState
+  -> AVote ByteString
+  -> m VotingState
+registerVoteWithConfirmation pm votingEnv vs vote = do
+
+  -- Register the vote ignoring proposal confirmation
+  votes' <- registerVote pm voteRegEnv votes vote
+
+  -- Confirm the proposal if it passes the threshold and isn't confirmed
+  let
+    confirmedProposals' = if pastThreshold votes' && not (isConfirmed upId)
+      then M.insert upId slot confirmedProposals
+      else confirmedProposals
+
+  -- Return the new state with additional vote and maybe confirmation
+  pure $ VotingState
+    { vsVotes = votes'
+    , vsConfirmedProposals = confirmedProposals'
+    }
+ where
+  VotingEnvironment slot _ voteRegEnv  = votingEnv
+  VotingState votes confirmedProposals = vs
+
+  -- | This is the number of genesis keys that need to support a proposal
+  threshold :: Int
+  threshold = 5
+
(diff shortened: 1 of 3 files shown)


src/Cardano/Chain/Update/Validation/Voting.hs
```diff
@@ -0,0 +1,126 @@
...
+import Cardano.Chain.Slotting
+import Cardano.Chain.Update.ProtocolParameters
+import Cardano.Chain.Update.Vote
+import Cardano.Crypto
+
+
+data VotingState = VotingState
+  { vsVotes              :: RegisteredVotes
+  , vsConfirmedProposals :: Map UpId SlotId
+  }
```
dnadales, line 45: As our states become nested, lenses become quite handy. Is there any reason why we're not making lenses out of these fields?

src/Cardano/Chain/Update/Validation/Voting.hs
```diff
@@ -0,0 +1,126 @@
...
+
+type RegisteredVotes = Map UpId (Set StakeholderId)
```
dnadales, line 47: I don't want to be like they say in the NL, an "ant ~fuc~ lover", but there is some asymmetry between the types in the `VotingState`: in one case we're using a type alias, and in the other case we're using the type directly. Do we want to unify the style? (I personally prefer using the map directly)

:ant: :running_man:

src/Cardano/Chain/Update/Validation/Voting.hs
```diff
@@ -0,0 +1,126 @@
...
+
+-- | Register a vote and confirm the corresponding proposal if it pass the
+--   voting threshold. This corresponds to the `UPVOTE` rules in the spec.
+registerVoteWithConfirmation
+  :: MonadError VotingError m
+  => ProtocolMagicId
+  -> ProtocolParameters
+  -> SlotId
+  -> Set UpId
+  -> Map StakeholderId StakeholderId
```
dnadales: :ant: :heart: :running_man: : Maybe add some comments on what these parameters are, so that people don't need to read the specs. Also I don't know whether it is wise to place the environment parameters in the same order they appear in the spec.

Also related, is there a reason why we have a `VotingState` but no `VotingEnvironment`?

src/Cardano/Chain/Update/Validation/Voting.hs
```diff
@@ -0,0 +1,126 @@
...
+
+    -- Return the new state with additional vote and maybe confirmation
+    pure $ VotingState
+      { vsVotes = votes'
+      , vsConfirmedProposals = confirmedProposals'
+      }
+ where
+  VotingState votes confirmedProposals = vs
+
+  -- This is the number of genesis keys that need to support a proposal
```
dnadales: `-- | This ..`?

src/Cardano/Chain/Update/Validation/Voting.hs
```diff
@@ -0,0 +1,126 @@
...
+  threshold :: Int
+  threshold = 5
```
dnadales, line 87: Any ideas on whether we want to make this parameter configurable?

src/Cardano/Chain/Update/Validation/Voting.hs
```diff
@@ -0,0 +1,126 @@
...
+-- | Validate and register a vote
+--
+--   We check that
+--
+--   1) The vote is for a registered proposal
+--   2) There is at least one genesis key delegating to the voter
+--   3) The signature is valid
+--
+--   This corresponds to the `ADDVOTE` rule in the spec.
+registerVote
```
dnadales, line 107: The same questions as with `registerVoteWithConfirmation` apply here :)

specs/ledger/latex/update-mechanism.tex
```diff
@@ -747,10 +749,10 @@ \subsection{Voting on update proposals}
     \label{eq:rule:voting}
     \inference
     {
-      \var{pid} = \vPropId{v} & \var{vk} = \vCaster{v} &
+      \var{pid} = \vPropId{v} & \var{vk} = \vCaster{v} & \var{pid} \in \var{rups}\\
       \var{vts}_{\var{pid}} =
-      \{ (\var{pid}, \var{vk_s}) \mid \var{vk_s} \mapsto \var{vk} \in \var{dms} \}\\
-      & \var{pid} \in \var{rups} &
+      \{ (\var{pid}, \var{vk_s}) \mid \var{vk_s} \mapsto \var{vk} \in \var{dms} \} &
+      \var{vts}_{\var{pid}} \neq \emptyset &
```
dnadales: Thanks Ru! :pray:

dnadales (approved):
Looking so good and clean :sparkles: Thanks!

src/Cardano/Chain/Update/Validation/Voting.hs
ruhatch: Haha! We use the type alias in the first case because that map is then passed as the argument to another function. At that stage it becomes clearer in the type signature what this thing is representing. As we don't use the second one in any type signatures I don't think we need an alias for it right now.

src/Cardano/Chain/Update/Validation/Voting.hs
ruhatch: I don't think this is complicated enough that we actually gain anything from using lenses

dnadales: It is not complicated now, but if we later decide to use lenses we'd have to do some renamings here and there. But nothing serious anyway ... It's just that for the elaborator I wished I had lenses available.

src/Cardano/Chain/Update/Validation/Voting.hs
ruhatch: Hmm, yeah, it's hard to know when to clarify with a comment and when to clarify with a well named `type` alias. In this case if I put things into the `VotingEnvironment` then all the references will be by name, so that would clarify. I'll do that here!

src/Cardano/Chain/Update/Validation/Voting.hs
ruhatch: Is that valid syntax in `where` clause?

src/Cardano/Chain/Update/Validation/Voting.hs
ruhatch: We could hardcode it as a proportion of the number of genesis keys?

ruhatch: That way it takes into account how many keys there are, but we don't have to add extra configuration for this

src/Cardano/Chain/Update/Validation/Voting.hs
dnadales: Oh, I missed that fact.

src/Cardano/Chain/Update/Validation/Voting.hs
dnadales: :+1:

src/Cardano/Chain/Update/Validation/Voting.hs
erikd: Shouldn't those fields have strict annotations?

dnadales: @ruhatch, should I create an issue to review the strictness annotations in `cardano-chain`?

PR #322, opened 2019-02-12
[#319] Add `Lovelace` property tests

### LINKED ISSUE #319: Implement property tests for `Cardano.Chain.Common.Lovelace` module
We currently have no property tests for the functions related to `Lovelace`.

Jimbo4350:
Closes #319

### DIFF
--- a/.stack-to-nix.cache
+++ b/.stack-to-nix.cache
@@ -13,3 +13,5 @@ https://github.com/input-output-hk/cardano-crypto 3c707936ba0a665375acf5bd240dc4
 https://github.com/avieth/plutus-prototype d094be301195fcd8ab864d793f114970426a4478 . 1s932rghn4zn441waansp408b5bwk20rc1wsa5693a2nwnp4dijw plutus-prototype .stack.nix/plutus-prototype.nix
 https://github.com/input-output-hk/haskell-hedgehog.git 2e741bb53afb085741807018948ae17d956c53af hedgehog 0l0d1n2b68m0628j4yi214q5fy6pz777qfj1bc1lrra8scs5gcxh hedgehog .stack.nix/hedgehog.nix
 http://github.com/nc6/canonical-json a9dc9b893649bc2e2a770ab22d278a780f7e3a3c . 0alwbi9xqaj6fmwzs6lr2drqrnhlnp13d9k2qkl5ga7h4grz9zcr canonical-json .stack.nix/canonical-json.nix
+https://github.com/input-output-hk/cardano-prelude e1fb84b1a955b6e3e3e53d9022e8bc0f927417a2 . 0ikxxv3b5rhq56fj2vbfsdnvc3g5lj5yyfxdph05n8wccjx0bd2m cardano-prelude .stack.nix/cardano-prelude.nix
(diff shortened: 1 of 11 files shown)


test/Test/Cardano/Chain/Common/LovelaceSpec.hs
```diff
@@ -0,0 +1,114 @@
...
+  a <- forAll genLovelace
+  let newRange = maxLovelaceVal - unsafeGetLovelace a
+  b <- forAll $ genCustomLovelace newRange
+  assertEitherIsRight (addLovelace a) b
+
+prop_addLovelaceOverflow :: Property
+prop_addLovelaceOverflow = eachOf
+  1
+  (pure lovelaceMaxval)
+  (assertEitherIsLeft (addLovelace (mkKnownLovelace @1)))
```
mhuesch: I think this could be written as `property $ assertEitherIsLeft (addLovelace (mkKnownLovelace @1)) lovelaceMaxval`, no?

[`property`](https://hackage.haskell.org/package/hedgehog-0.6.1/docs/Hedgehog.html#v:property) takes a PropertyT and produces a Property. `assertEitherIsLeft` has a MonadTest instance, so should be able to get instantiated at PropertyT type.

And same goes for the other `eachOf 1 ...` properties below.

test/Test/Cardano/Chain/Common/LovelaceSpec.hs
```diff
@@ -0,0 +1,114 @@
...
+  b <- forAll $ genCustomLovelace (unsafeGetLovelace a)
+  assertEitherIsRight (subLovelace a) b
+
+prop_subLovelaceUnderflow :: Property
+prop_subLovelaceUnderflow =
+  withTests 1000
+    . property
+    $ do
+  -- (maxLovelaveVal - 1) to avoid an overflow error in `addLovelace`
+  -- in case expression
```
mhuesch: These could be indented over in line with `a <- ...` below.

mhuesch (changes requested):
Looks good! Minor misuse of `eachOf` which could be rewritten.

test/Test/Cardano/Chain/Common/LovelaceSpec.hs
Jimbo4350: Yep this can def work, I was just trying to reuse functions from `cardano-prelude`.

mhuesch (approved):

test/Test/Cardano/Chain/Common/LovelaceSpec.hs
```diff
@@ -0,0 +1,112 @@
+{-# LANGUAGE DataKinds         #-}
+{-# LANGUAGE NumDecimals       #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TemplateHaskell   #-}
+{-# LANGUAGE TypeApplications  #-}
+
+module Test.Cardano.Chain.Common.LovelaceSpec
```
ruhatch: Could you remove `Spec` from this? I think the module name is enough without it :)

test/Test/Cardano/Chain/Common/LovelaceSpec.hs
```diff
@@ -0,0 +1,112 @@
...
+          Right added -> assertEitherIsLeft (subLovelace a) added
+          Left  err   -> panic $ sformat
+            ("The impossible happened in subLovelaceUnderflow: " . build)
+            err
+
+--------------------------------------------------------------------------------
+-- Helpers
+--------------------------------------------------------------------------------
+
+lovelaceMaxval :: Lovelace
```
ruhatch: You don't need this, you can just use `maxBound`

ruhatch (approved):
Just a couple of small changes!

PR #328, opened 2019-02-15
Document workflow for modifying IOHK-owned extra-deps locally.

### LINKED ISSUE #309: Document workflow for modifying IOHK-owned extra-deps locally
There seems to be preference for doing a `stack-local.yaml` approach to managing subproject extra-deps, rather than something like Git submodules. It would be great to document this workflow, so that devs joining the project can easily make local changes to e.g. `cardano-prelude` and observe their effects in `cardano-chain` without committing/pushing/pulling via a remote to do so.

`stack-local.yaml` would resemble `stack.yaml` but instead of e.g.
```
extra-deps:
  - git: https://github.com/input-output-hk/cardano-prelude
    commit: 7f8544ae9e162b6664f07093227edb148f1c8a5b
    subdirs:
      - .
      - test
```
we would have
```
extra-deps:
  - ../cardano-prelude
  - ../cardano-prelude/test
```
***

I would also suggest adding a wrapper script to the repository, which would make `stack` workflows with local repositories easier. Something like:
```
$ cat local-stack.sh 
#!/bin/sh
export STACK_YAML=stack-local.yaml
stack --work-dir=.stack-work-local $@

$ ./local-stack.sh build
Getting project config file from STACK_YAML environment
<building>
```

dnadales:
Since #308 is on hold due to possible changes in the delegation spec, I'm creating the PR which addresses this particular issue.

Closes #309.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,6 @@ specs/**/.ghc.environment.x86_64-linux-*
 
 # gitinfo2 auxiliary file
 gitHeadLocal.gin
+
+# Local development
+stack-local.yaml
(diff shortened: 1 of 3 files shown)


ruhatch (approved):
Looks great, thank you!

PR #329, opened 2019-02-15
Create issue-template.md

ruhatch:


.github/issue-template.md
```diff
@@ -0,0 +1,3 @@
+---
+labels: inbox
```
dnadales, line 2: Uh?

dnadales (commented):
Why do we need issue templates? My guess is that at this stage we will be deleting the template in most of the cases ...

PR #335, opened 2019-02-21
Test the concrete validators against generated chains

### LINKED ISSUE #196: Elaborator from executable spec to real chain types (delegation)

### LINKED ISSUE #197: Create test harness for generating and validating chains with delegation

dnadales:
The chains are generated according to our formal spec rules.

I'm waiting on [this PR](https://github.com/input-output-hk/fm-ledger-rules/pull/259) to update the git hash of `fm-ledger-rules`.

Closes #197 
Closes #196


binary/src/Cardano/Binary/Class/Primitive.hs
```diff
@@ -141,8 +141,12 @@ decodeFull' = decodeFull . BSL.fromStrict
 
 decodeFullDecoder
   :: Text
+  -- ^ Label for error reporting
   -> (forall s . D.Decoder s a)
+  -- ^ The parser for the @ByteString@ to decode. It should decode the given
+  -- @ByteString@ into a value of type @a@
   -> BSL.ByteString
+  -- ^ The @ByteString@ to decode
```
ruhatch, line 149: Nice, thank you

test/Test/Cardano/Chain/Block/Validation/Spec.hs
```diff
@@ -0,0 +1,86 @@
...
+passConcreteValidation
+  :: MonadTest m
+  => Trace CHAIN -> m ()
+passConcreteValidation tr = void $ evalEither res
+  where
+    config = abEnvToCfg (tr ^. traceEnv)
+
+    initSt = either (panic . show) identity $ initialChainValidationState config
+
+    res = foldM elaborateAndUpdate initSt $ preStatesAndSignals OldestFirst tr
```
ruhatch: Personally I like to order `where`  statements in a kind of depth first way, so I'd do `res` first as it shows up first in the body, then `elaborateAndUpdate` as that's first in `res`, etc. That way the next thing to look for is (usually) next in line!

test/Test/Cardano/Chain/Elaboration/Block.hs
```diff
@@ -0,0 +1,243 @@
...
+        (Genesis.configProtocolMagicId config)
+        prevHash
+        0
+        sid
+        ssk
+        cDCert
+        bb0
+        extraHeaderData
+
+    emptyAttrs = Common.Attributes () (Common.UnparsedFields [])
```
ruhatch: `Common.mkAttributes ()`

test/Test/Cardano/Chain/Elaboration/Block.hs
```diff
@@ -0,0 +1,243 @@
...
+  -> Abstract.Block
+  -> Concrete.ABlock ByteString
+elaborateBS config aenv ast st ab
+  = annotateBlock $ elaborate config aenv ast st ab
+
+annotateBlock
+  :: Concrete.Block
+  -> Concrete.ABlock ByteString
+annotateBlock block =
+  let res =
```
ruhatch: Some of these short names are a little confusing. I like some short names, but here I think `decodedABlockOrBoundary` would be fine!

ruhatch: That way I can read the code in multiple passes:
1) Assuming sensibly named things do what I expect, and just reading the high-level
2) Checking the details and making sure they meet my assumptions, once I'm clear about the high-level

test/Test/Cardano/Chain/Elaboration/Block.hs
```diff
@@ -0,0 +1,243 @@
...
+rcDCert vk ast
+  = mkDCert vkg sigVkg vk (ast ^. epochL)
+  where
+    dm :: Map Abstract.VKeyGenesis Abstract.VKey
+    dm = ast ^. disL . delegationMap
+
+    vkg = fromMaybe err $ delegatorOf dm vk
+
+    err :: Abstract.VKeyGenesis
+    err = panic $ "No delegator found for key " <> show vk
```
ruhatch: I think it could be better to have an `ElaborationError` sum type around to avoid all these `panic`s. Then we can just have a single `panic` at the top level when we really do need to panic!

ruhatch (approved):
Looks great and so glad to see it all hooked up! I'd like to see some of the names lengthened, although I know it's hard to know what's unclear when you've already been looking at the code for so long!

test/Test/Cardano/Chain/Elaboration/Block.hs
dnadales: Great! Me like deleting characters :smile:

test/Test/Cardano/Chain/Elaboration/Block.hs
dnadales: You got it. I think in this case a long name name makes sense.

dnadales: I just added the characters I deleted when addressing the previous comment. I guess the universe just tends to stay in equilibrium ...

test/Test/Cardano/Chain/Elaboration/Block.hs
dnadales: Hmmm, I don't know. I guess in these cases we do want to panic as soon as possible, since this indicates an error in our generators logic (or in the spec). So, as far as I can see, bubbling up the error to other call sites won't give us any benefits but will clutter the code.

dnadales:
> Looks great and so glad to see it all hooked up! I'd like to see some of the names lengthened, although I know it's hard to know what's unclear when you've already been looking at the code for so long!

Thank you for the time you spent helping me out with this!

test/Test/Cardano/Chain/Block/Validation/Spec.hs
dnadales: Yes, that's my preference as well. Thanks for spotting this. I'll change it.

PR #346, opened 2019-03-03
Remove plutus-protoype from build

erikd:
The Plutus protoype code was never used on mainnet and is going to
be replaced by a new version at some later date.


dcoutts (approved):
Indeed, will never be used.

erikd:
Just have to fix the build :)

Issue #181, opened 2018-11-30
Change PublicKey and SecretKey to VerificationKey and SigningKey


ruhatch:
We currently have a bunch of keys that are named Public and Secret, but are used only for signature schemes. We can therefore rename them to Verification and Signing, to capture their semantic content.

ruhatch:
@erikd this isn't done yet! Any reason for closing?

erikd:
I thought that commit was the fix. Reopenning.

PR #347, opened 2019-03-04
[#321] Rename ProxySecretKey to ProxyVerificationKey

### LINKED ISSUE #321: Rename `ProxySecretKey`
[`ProxySecretKey`](https://github.com/input-output-hk/cardano-chain/blob/3b2282983c993c450a8a57db3a136015a3003931/crypto/src/Cardano/Crypto/Signing/Proxy/SecretKey.hs#L64) appears to have migrated from `cardano-sl` to `cardano-chain`. It recently came to @intricate and my attention that this name is misleading, because the datatype doesn't contain any "secret" information (no secret keys or similar).

Doing a little digging just now, I found [a document](https://github.com/input-output-hk/cardano-sl/blob/develop/docs/block-processing/types.md#proxy-key) which admits the name is a misnomer. Git blame [here](https://github.com/input-output-hk/cardano-sl/blame/develop/docs/block-processing/types.md#L1011) - this seems to have been known since 2018-01-11.

I'm not sure what a good replacement name would be. Perhaps something like `ProxyCertKey`?

intricate:
The name `ProxySecretKey` appears to be a misnomer as there is no private or secret information contained within this data type.

Closes #321


mhuesch (approved):
LGTM :shipit: 

```
$ ag ProxySecretKey
$
```

^ seems like they're all gone.

dnadales (approved):

ruhatch (commented):
If we're going to make this change I think we should call this `ProxyVerificationKey`, as @mhuesch identified. One other note is that this is only ever going to be used in the `Delegation.Certificate`s. Perhaps we should remove this entirely, as it represents premature abstraction of some kind.

In that case I guess we can go ahead and merge this and I'll open a new issue to discuss that.

intricate:
@ruhatch I have another PR that I'm raising after this in which I address #181. If you'd like, though, I can specifically rename this type in this PR.

ruhatch:
> @ruhatch I have another PR that I'm raising after this in which I address #181. If you'd like, though, I can specifically rename this type in this PR.

Yeah, I think that would be best, thanks

PR #349, opened 2019-03-06
Set main branch for Buildkite ci-cache

rvl:
This will make Buildkite branch builds faster because they can start with the `.stack-work` cache from `master`.


ruhatch (approved):
LGTM!

bors r+

PR #351, opened 2019-03-11
Add: `updateBody` method reflecting BBODY rule.

nc6:
This should ultimately replace `updateChain` in conjunction with other methods
reflecting header processing, as well as PBFT.

Currently this only calls down to delegation, as update and UTxO implementations are not complete yet.


src/Cardano/Chain/Block/Validation.hs
```diff
@@ -40,7 +43,7 @@ import Cardano.Chain.Block.Header
   (BlockSignature(..), HeaderHash, recoverSignedBytes, wrapBoundaryBytes)
 import Cardano.Chain.Common (BlockCount(..), StakeholderId, mkStakeholderId)
 import qualified Cardano.Chain.Delegation as Delegation
-import Cardano.Chain.Delegation.Payload (APayload(..))
+import qualified Cardano.Chain.Delegation.Payload as DlgPayload
```
ruhatch, line 51: Could just do `Delegation.getPayload` and not need this import, but that might be less clear

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -215,6 +225,48 @@ updateChainBoundary config cvs bvd = do
...
+updateBody config cvs b = do
+
+  -- Validate the block size
+  -- TODO Max block size should come from protocol params
+  let maxBlockSize :: Int64
+      maxBlockSize = 2e6 in
+    blockLength b <= maxBlockSize `orThrowError` ChainValidationBlockTooLarge
+
+  -- Validate the delegation payload signature
+  forM_ (DlgPayload.getPayload $ blockDlgPayload b)
```
ruhatch: These are checking in the `SDELEG` rule I believe, so probably shouldn't do it twice

ruhatch (approved):
Neat - I like the code line changes! Only concern is the delegation signature checking.

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -215,6 +225,48 @@ updateChainBoundary config cvs bvd = do
       $ boundaryHeaderBytes bvd
     }
 
+-- | This is an implementation of the BBODY rule as per the chain specification.
+--   Compared to `updateChain`, this does not validate any header level checks, nor
+--   does it carry out anything which might be considered part of the protocol.
+updateBody
```
dnadales, line 226: A minor thing only, but there seems to be some asymmetry in the names here: we have a function `updateChain`, which updates the chain state, and a function `updateBody`, which also seems to update the chain state. This function seems to be applying the block body to the chain state, so what about calling it something like `applyBody` or `applyBlockBody`?

src/Cardano/Chain/Block/Validation.hs
dnadales: Right. However this duplication can be traced back to the specs: we do have this verification in the `BBODY` rule:

image

and we also assume the signature comes in the delegation certificate:

image

It seems that we have to remove one of the checks in the rules. Any ideas which?

nc6: These are checking two different things. The check in `BBODY` is validating that the signature (in the block header) correctly validates the entire delegation payload. The check in `SDELEG` is verifying the body of the certificate against the public key of the delegator.

It looks like I'm probably checking the wrong thing in the code.

src/Cardano/Chain/Block/Validation.hs
nc6: I'm ultimately going to delete `updateChain`, but I can also rename these other things...

src/Cardano/Chain/Block/Validation.hs
dnadales: Oh, you're right. Then we should probably verify the delegation payload without needing the `forM_` (like you pointed out :) )

src/Cardano/Chain/Block/Validation.hs
ruhatch: Yeah, maybe `applyBody` or something like that. So we `apply` the contents of the body to the chain state?

nc6:
bors retry

cleverca22:
bors try

PR #353, opened 2019-03-13
Implement the update endorsement rules

### LINKED ISSUE #188: Implement rules for update endorsement

dnadales:
Closes #188


src/Cardano/Chain/Common/BlockCount.hs
```diff
@@ -15,7 +15,7 @@ import Formatting.Buildable (Buildable)
 import Cardano.Binary.Class (Bi(..))
 
 newtype BlockCount = BlockCount
-    { getBlockCount :: Word64
+    { unBlockCount :: Word64
```
mhuesch: Just curious - is `un<..>` preferred over `get<..>`? We seem to use both for newtype fields in this repo. I have no real preference, just wondering.

src/Cardano/Chain/Update/Validation/Endorsement.hs
```diff
@@ -0,0 +1,153 @@
...
+  -- ^ Chain stability parameter.
+  , currentSlot               :: !FlatSlotId
+  , delegationMap             :: !(Map StakeholderId StakeholderId)
+  , adoptedProtocolParameters :: !ProtocolParameters
+  , confirmedProposals        :: !(Map UpId FlatSlotId)
+  , registeredUpdateProposals :: !Registration.ProtocolUpdateProposals
+  , numGenesisKeys            :: !Word8
+  -- ^ Number of genesis keys. This is used in combination with the
+  -- 'ppUpdateProposalThd' protocol parameter to calculate the proportion of
+  -- genesis keys that need to endorse a new protocol version to considered for
```
mhuesch:
```suggestion
  -- genesis keys that need to endorse a new protocol version for it to be considered for
```

or perhaps

```suggestion
  -- genesis keys which must endorse a new protocol version for it to be considered for
```

?

src/Cardano/Chain/Update/Validation/Endorsement.hs
```diff
@@ -0,0 +1,153 @@
...
+  -> ProtocolParameters
+  -> Set Endorsement
+  -> ProtocolVersion
+  -> Bool
+canAdopt n pps endorsements protocolVersion = t <= numberOfEndorsements
+ where
+  ProtocolParameters { ppUpdateProposalThd } = pps
+
+  t :: Double
+  t = lovelacePortionToDouble ppUpdateProposalThd * fromIntegral n
```
mhuesch: I was pretty confused by this part, but I think I figured it out. I want to check my understanding, though.

The spec says
```
canAdopt pps bvs bv = upAdptThd |→ t ∈ pps ∧ t ≤ |{bv} <| bvs|

-- (`<|` being domain restriction)
```

which seems to indicate `t` will be an integer number Genesis Keys. We must have that number, or more, of endorsements for a specific BlockVersion, for it to be adopt-able.

***

In this code, we seem to be making an assumption that stake is evenly distributed among Genesis Keys (I think this is true, but still an assumption). Thus we can determine the threshold number `t` of Genesis Keys needed to adopt an update. We filter the set to find endorsements for a given BlockVersion, and compare that to `t`.

***

Since we're using LovelacePortion, I was expecting we would compute that actual stake of the Genesis Keys and compare that against the threshold. However that would be more complex, so this seems tidier, given that the equal-stake assumption holds.

***

```suggestion
  -- Here we assume that stake is evenly distributed among the Genesis Keys.
  -- Thus, the LovelacePortion can be multiplied by the number of keys to yield
  -- the endorsement threshold.
  t = lovelacePortionToDouble ppUpdateProposalThd * fromIntegral n
```

A comment like ☝️might be helpful for future readers.

mhuesch (approved):
Looks good! I noted some optional tweaks and questions.

src/Cardano/Chain/Common/BlockCount/JSON.hs
```diff
@@ -0,0 +1,9 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Cardano.Chain.Common.BlockCount.JSON (unToGet) where
+
+import Cardano.Prelude
+
+unToGet :: [Char] -> [Char]
```
nc6: This is a very weird function unless the context behind it is explained. So can we have that context in some comments for it, please?

src/Cardano/Chain/Common/BlockCount.hs
dnadales: I cannot recall why we preferred `un<..>` over `get<..>`, but considering that we have to adapt the Aeson instances, maybe I'd just revert this. @ruhatch: any thoughts?

src/Cardano/Chain/Update/Validation/Endorsement.hs
```diff
@@ -0,0 +1,156 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE NamedFieldPuns   #-}
+
+module Cardano.Chain.Update.Validation.Endorsement
+  ( registerEndorsement
```
nc6, line 5: This module is going to be rather tricky to use, since the only exported function requires values of type `Environment` and `State` but offers no way to create them :-)

src/Cardano/Chain/Update/Validation/Endorsement.hs
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+data State = State
+  { futureProtocolVersions :: [CandidateProtocolVersion]
+  , registeredEndorsements :: Set Endorsement
+  }
+
+
+data CandidateProtocolVersion = CandidateProtocolVersion
+  { fpvSlot               :: FlatSlotId
```
nc6: [minor] `cpv` rather than `fpv` here? I'm guessing this used to be called `FutureProtocolVersion`?

nc6: Can we document what slot this refers to? Things like `fpvProtocolParameters` are self-explanatory, but slots are inevitably not... is this the slot it was submitted on, the one it will come into force on, or the one it expires on?

src/Cardano/Chain/Update/Validation/Endorsement.hs
dnadales: Now that I think about it, I'm not sure that we even need to assume even distribution of the stake.

If I understand this correctly, the problem is that we don't have a `upAdptThd` protocol parameter, so we need to use one of the existing parameters. 

I think the assumption here is that `lovelacePortionToDouble ppUpdateProposalThd` will be a number around `0.5`, so if we have `7` genesis keys, we'd need more than `t = 3.5` endorsements, which would be the same as having an `upAdptThd` parameter of `4`. 

Now the problem seems to be that `ppUpdateProposalThd` (called `updateProposalThd` in `cardano-sl` if I'm not mistaken) is set to `100000000000000`, so `lovelacePortionToDouble ppUpdateProposalThd` will be `0.1`!

From `cardano-sl/lib/mainnet-genesis.json`:

```json
    "blockVersionData": {
        "heavyDelThd": "300000000000",
        "maxBlockSize": "2000000",
        "maxHeaderSize": "2000000",
        "maxProposalSize": "700",
        "maxTxSize": "4096",
        "mpcThd": "20000000000000",
        "scriptVersion": 0,
        "slotDuration": "20000",
        "softforkRule": {
            "initThd": "900000000000000",
            "minThd": "600000000000000",
            "thdDecrement": "50000000000000"
        },
        "txFeePolicy": {
            "multiplier": "43946000000",
            "summand": "155381000000000"
        },
        "unlockStakeEpoch": "18446744073709551615",
        "updateImplicit": "10000",
        "updateProposalThd": "100000000000000",
        "updateVoteThd": "1000000000000"
    },
```

So it seems we might need to change the `canAdopt` function.

I'm tempted to use `softforkRule.minThd` since ` (600000000000000 / 1e15) * 7 = 4.2`,

src/Cardano/Chain/Update/Validation/Endorsement.hs
```diff
@@ -0,0 +1,156 @@
...
+        then do
+          let
+            fpv = CandidateProtocolVersion
+              { fpvSlot = currentSlot
+              , fpvProtocolVersion = pv
+              , fpvProtocolParameters = pps'
+              }
+            fpvs' = updateCandidateProtocolVersions futureProtocolVersions fpv
+          pure $ State
+            { futureProtocolVersions = fpvs'
```
nc6: Any reason we're using `fpvs` here rather than `fads`, as used in the latex spec?

src/Cardano/Chain/Update/Validation/Endorsement.hs
```diff
@@ -0,0 +1,156 @@
...
+    _ -> throwError $ MultipleProposalsForProtocolVersion pv
+ where
+  Environment { k, currentSlot, delegationMap, confirmedProposals, registeredUpdateProposals, numGenesisKeys }
+    = env
+
+  isConfirmedAndStable upId = upId `M.member` scps
+   where
+      -- Stable and confirmed proposals.
+    scps = M.filter (currentSlot - _2 k <=) confirmedProposals
+    _2 x = 2 * unBlockCount x
```
nc6: I find this a rather confusing notation (I would automatically assume `_2` to be the projection lens from a tuple), though I guess it's close enough to the site to be okay. But why not just inline this?

nc6 (changes requested):
Some small changes requested, but otherwise looks good.

src/Cardano/Chain/Common/BlockCount/JSON.hs
dnadales: I reverted the change. I think it isn't worth the extra complexity.

src/Cardano/Chain/Update/Validation/Endorsement.hs
dnadales: Yeah, indeed :) As I hook this module to others I'll start adding the required exports to that list :)

src/Cardano/Chain/Update/Validation/Endorsement.hs
dnadales: Good catch! Bad programmer, bad programmer :newspaper_roll:

dnadales: Good point! Thanks for bringing this to my attention. I'll add some documentation now.

src/Cardano/Chain/Update/Validation/Endorsement.hs
dnadales: I guess since we called that field of the state "future protocol versions ", `fpvs` is a suitable acronym, although it differs from the spec.

src/Cardano/Chain/Update/Validation/Endorsement.hs
dnadales: Indeed. I just changed that.

ruhatch (approved):
LGTM! Thanks for finishing this and tidying up!

src/Cardano/Chain/Update/Validation/Endorsement.hs
```diff
@@ -35,15 +35,21 @@ data Environment = Environment
...
-  { fpvSlot               :: FlatSlotId
-  , fpvProtocolVersion    :: ProtocolVersion
-  , fpvProtocolParameters :: ProtocolParameters
+  { cpvSlot               :: FlatSlotId
+    -- ^ Slot at which this protocol version and parameters gathered enough
+    -- endorsements and became a candidate. This is used to check which
+    -- versions became candidates 2k slots before the end of an epoch (and only
+    -- those can be adopted at that epoch). Versions that became candidates
+    -- later than 2k slots before the end of an epoch can be adopted in
+    -- following epochs.
```
nc6, line 51: Great, thanks!

nc6 (approved):

src/Cardano/Chain/Update/Validation/Endorsement.hs
```diff
@@ -132,18 +133,23 @@ canAdopt
...
-  -- Thus, the LovelacePortion can be multiplied by the number of keys to yield
-  -- the endorsement threshold.
-  t = lovelacePortionToDouble ppUpdateProposalThd * fromIntegral n
+  -- In Byron we do not have a @upAdptThd@ protocol parameter, so we have to
+  -- use the existing ones.
+  --
+  -- @lovelacePortionToDouble . srMinThd . ppSoftforkRule@ will give us the
+  -- ratio (in the interval @[0, 1]@) of the total stake that has to endorse a
+  -- protocol version to become adopted. In genesis configuration, this ratio
+  -- will evaluate to @0.6@, so if we have 7 genesis keys, @upAdptThd = 4@.
```
mhuesch, line 145: Thanks! I find this implementation more clear.

I now am confused by a different point - rounding of the `stakeRatio * fromIntegral ngk`. Shouldn't that be `ceiling`? If we insist that `0.6 * 7 = 4.2` or more nodes agree, shouldn't we round up to 5?

Alternately we could stay with `floor` but say `upAdptThd < numberOfEndorsements` as the final conditional (strict GT).

mhuesch (commented):
Another Q 🙂

src/Cardano/Chain/Update/Validation/Endorsement.hs
mhuesch: > Now that I think about it, I'm not sure that we even need to assume even distribution of the stake.
> 
> If I understand this correctly, the problem is that we don't have a upAdptThd protocol parameter, so we need to use one of the existing parameters.

Isn't this a separate issue, though? If we're translating from "percentage of total stake which must endorse this proposal" to "percentage of genesis keys, which all together control the total stake, which must endorse this proposal", it seems to me that we need to know how much stake they control (or assume it's evenly split).

I believe that `softforkRule.minThd` is appropriate to use here. In `cardano-sl`, update adoption was triggered when the percentage of stake endorsing the new BlockVersion exceeded the softforkRule threshold (although it used the moving target, which we're not).

src/Cardano/Chain/Update/Validation/Endorsement.hs
dnadales: I don't know what the right value should be: `4` or `5`, but I guess we can tweak this as needed.

src/Cardano/Chain/Update/Validation/Endorsement.hs
dnadales: > Isn't this a separate issue, though? If we're translating from "percentage of total stake which must endorse this proposal" to "percentage of genesis keys, which all together control the total stake, which must endorse this proposal", it seems to me that we need to know how much stake they control (or assume it's evenly split).

Well the thing is that in this case we don't care about the stake they control, all we care is that they are genesis keys.

However if this brings too much confusion we might need to introduce a protocol parameter with a default value.

> I believe that softforkRule.minThd is appropriate to use here. In cardano-sl, update adoption was triggered when the percentage of stake endorsing the new BlockVersion exceeded the softforkRule threshold (although it used the moving target, which we're not).

Great! Thanks for confirming this :)

PR #354, opened 2019-03-14
Disable nix-tools.tests.cardano-chain.cardano-chain-test.x86_64-darwin

jbgi:
as the job build freezes on the darwin hydra agents.


ruhatch (approved):
LGTM!

Issue #189, opened 2018-11-30
Implement ledger part of update interface


ruhatch:
implement the update interface rules:

- [x] UPIREG Update-proposal-interface registration.
- [x] UPIVOTE Update-proposal-interface vote.
- [x] UPIEND Update-proposal-interface endorsement.
- [x] UPIEC Update-proposal-interface epoch-change.

Plus:

- [x] Determine what the proposal parameter time to live should be
- [x] Rebase to master
- [x] Make sure our style is consistent
  - [x] Ensure consistency in the use of named field puns.
  - [x] Add an environment to `Cardano.Chain.Update.Validation.Registration`
  - [x] Make sure the other rules also have an environment.
  - [x] Use explicit imports.
- [x] Remove compiler warnings
- [x] Run hlint
- [x] Add more comments

PR #355, opened 2019-03-14
Implement UPIREG and UPIVOTE rules

dnadales:
Part of #189.

### DIFF
--- a/src/Cardano/Chain/Update/Validation/Interface.hs
+++ b/src/Cardano/Chain/Update/Validation/Interface.hs
@@ -1,0 +1,167 @@
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE NamedFieldPuns #-}
+
+-- | Blockchain interface validation rules.
+--
+module Cardano.Chain.Update.Validation.Interface
+  ( -- * Environment
+    Environment (..)
+    -- * State
+  , State (..)
+    -- *Error
+  , Error (..)
+    -- * Interface functions
+  , registerProposal
+  , registerVote
+  )
+where
+
+import Cardano.Prelude hiding (State)
+
+import qualified Data.Map.Strict as M
+
+import Cardano.Chain.Slotting (EpochIndex, FlatSlotId)
+import Cardano.Chain.Common.StakeholderId (StakeholderId)
+
+import Cardano.Chain.Update.ApplicationName (ApplicationName)
+import Cardano.Chain.Update.ProtocolParameters (ProtocolParameters)
+import Cardano.Chain.Update.ProtocolVersion (ProtocolVersion)
+import Cardano.Chain.Update.SoftwareVersion
+  ( NumSoftwareVersion
+  , SoftwareVersion
+  , svAppName
+  , svNumber
+  )
+import qualified Cardano.Chain.Update.Validation.Registration as Registration
+import qualified Cardano.Chain.Update.Validation.Voting as Voting
+import Cardano.Chain.Update.Vote (UpId, AProposal, recoverUpId, AVote)
+import Cardano.Crypto (ProtocolMagicId)
+
+data TODO
+
+data Environment = Environment
+  { protocolMagic :: ProtocolMagicId
+  , currentEpoch  :: !EpochIndex
+  , currentSlot   :: !FlatSlotId
+  , delegationMap :: !(Map StakeholderId StakeholderId)
+  }
+
+-- | Update interface state.
+data State = State
+  { prevEpoch                         :: !EpochIndex
+    -- ^ Previously seen epoch
+  , adoptedProtocolVersion            :: !ProtocolVersion
+  , adoptedProtocolParams             :: !ProtocolParameters
+    -- ^ Adopted protocol parameters
+  , futureAdopts                      :: !TODO -- We should take this from Cardano.Chain.Update.Validation.Endorsement
+    -- ^ Future protocol version adoptions
+  , appVersions                       :: !(Map ApplicationName (NumSoftwareVersion, FlatSlotId))
+    -- ^ Current application versions (by application name)
+  , registeredProtocolUpdateProposals :: !(Map UpId (ProtocolVersion, ProtocolParameters))
+    -- ^ Registered protocol update proposals
+  , registeredSoftwareUpdateProposals :: !(Map UpId SoftwareVersion)
+    -- ^ Registered software update proposals
+  , confirmedProposals                :: !(Map UpId FlatSlotId)
+    -- ^ Confirmed update proposals
+  , proposalVotes                     :: !(Map UpId (Set StakeholderId))
+    -- ^ Update proposals votes
+  , proposalsEndorsements             :: TODO -- We should take this from Cardano.Chain.Update.Validation.Endorsement
+                                              -- So this should be a @Set Endorsement@
+    -- ^ Update proposals endorsements
+  , proposalRegistrationSlot          :: Map UpId FlatSlotId
+    -- ^ Slot at which an update proposal was registered
+  }
+
+data Error
+  = Registration Registration.Error
+  | Voting Voting.Error
+
+-- | Register an update proposal.
+--
+-- This corresponds to the @UPIREG@ rules in the spec.
+registerProposal
+  :: MonadError Error m
+  => Environment
+  -> State
+  -> AProposal ByteString
+  -> m State
+registerProposal env st proposal = do
+  Registration.State rpus' raus' <-
+    Registration.registerProposal pm pv pps avs dms regSubSt proposal
+      `wrapError` Registration
+  pure $!
+    st { registeredProtocolUpdateProposals = rpus'
+       , registeredSoftwareUpdateProposals = raus'
+       , proposalRegistrationSlot = M.insert (recoverUpId proposal) currentSlot pws
+       }
+  where
+    Environment
+      { protocolMagic = pm
+      , currentSlot
+      , delegationMap = dms
+      } = env
+
+    State
+      { adoptedProtocolVersion = pv
+      , adoptedProtocolParams = pps
+      , appVersions = avs
+      , registeredProtocolUpdateProposals = rpus
+      , registeredSoftwareUpdateProposals = raus
+      , proposalRegistrationSlot = pws
+      } = st
+
+    regSubSt = Registration.State rpus raus
+
+-- | Register a vote for the given proposal.
+--
+-- If the proposal gets enough confirmations after adding the given vote, then
+-- it will get added to the set of confirmed proposals.
+-- This corresponds to the @UPIVOTE@ rules in the spec.
+--
+registerVote
+  :: MonadError Error m
+  => Environment
+  -> State
+  -> AVote ByteString
+  -> m State
+registerVote env st vote = do
+  Voting.State vts' cps' <-
+    Voting.registerVoteWithConfirmation pm subEnv subSt vote
+      `wrapError` Voting
+  let
+    avsNew =
+      M.fromList $! [ (svAppName sv, (svNumber sv, sn))
+                    | (pid, sv) <- M.toList raus
+                    , pid `elem` M.keys cps'
+                    ]
+  pure $!
+    st { confirmedProposals = cps'
+       , proposalVotes = vts'
+       , appVersions = M.union avsNew avs
+       , registeredSoftwareUpdateProposals = M.withoutKeys raus (M.keysSet cps)
+       }
+       -- TODO: consider using the `Relation` instances from `fm-ledger-rules` (see `Ledger.Core`)
+
+  where
+
+    Environment
+      { protocolMagic = pm
+      , currentSlot = sn
+      , delegationMap = dms
+      } = env
+
+    State
+      { adoptedProtocolParams = pps
+      , proposalRegistrationSlot
+      , proposalVotes = vts
+      , confirmedProposals = cps
+      , appVersions =  avs
+      , registeredSoftwareUpdateProposals = raus
+      } = st
+
+    rups = M.keysSet proposalRegistrationSlot
+
+    subEnv = Voting.Environment sn pps (Voting.RegistrationEnvironment rups dms)
+
+    subSt = Voting.State vts cps
--- a/cardano-chain.cabal
+++ b/cardano-chain.cabal
@@ -105,6 +105,7 @@ library
                        Cardano.Chain.Update.SystemTag
                        Cardano.Chain.Update.Undo
                        Cardano.Chain.Update.Validation.Endorsement
+                       Cardano.Chain.Update.Validation.Interface
                        Cardano.Chain.Update.Validation.Registration
                        Cardano.Chain.Update.Validation.Voting
                        Cardano.Chain.Update.Vote
--- a/src/Cardano/Chain/Update/Validation/Registration.hs
+++ b/src/Cardano/Chain/Update/Validation/Registration.hs
@@ -5,14 +5,14 @@
 --   This is an implementation of the rules defined in the Byron ledger
 --   specification
 module Cardano.Chain.Update.Validation.Registration
-  ( RegistrationError
-  , RegistrationState(..)
+  ( Error
+  , State(..)
   , ProtocolUpdateProposals
   , registerProposal
   )
 where
 
-import Cardano.Prelude
+import Cardano.Prelude hiding (State)
 
 import qualified Data.ByteString as BS
 import qualified Data.Map.Strict as M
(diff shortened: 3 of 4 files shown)


src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -0,0 +1,177 @@
...
+
+data Environment
+  = Environment
+  { _protocolMagic :: ProtocolMagicId
+  , _currentEpoch :: !EpochIndex
+  , _currentSlot :: !FlatSlotId
+  , _delegationMap :: !(Map StakeholderId StakeholderId)
+  }
+
+makeLenses ''Environment
```
ruhatch: Sorry, I don't think we're going to use lenses in the implementation anywhere!

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -0,0 +1,177 @@
...
+    { _prevEpoch :: !EpochIndex
+      -- ^ Previously seen epoch
+    , _adoptedProtocolVersion :: !ProtocolVersion
+    , _adoptedProtocolParams :: !ProtocolParameters
+      -- ^ Adopted protocol parameters
+    , _futureAdopts :: !TODO -- We should take this from Cardano.Chain.Update.Validation.Endorsement
+      -- ^ Future protocol version adoptions
+    , _appVersions :: !(Map ApplicationName (NumSoftwareVersion, FlatSlotId))
+      -- ^ Current application versions (by application name)
+    , _registeredProtocolUpdateProposals :: !(Map UpId (ProtocolVersion, ProtocolParameters))
```
ruhatch: There's a `RegisteredProtocolUpdateProposals` (or something) type synonym in `Cardano.Chain.Update.Validation.Registration` that you could use here

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -0,0 +1,177 @@
...
+                                    -- So this should be a @Set Endorsement@
+      -- ^ Update proposals endorsements
+    , _proposalRegistrationSlot :: Map UpId FlatSlotId
+      -- ^ Slot at which an update proposal was registered
+    }
+
+makeLenses ''State
+
+data Error
+  = Registration Registration.Error
```
ruhatch, line 77: Nice

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -0,0 +1,177 @@
...
+-- This corresponds to the @UPIREG@ rules in the spec.
+registerProposal
+  :: MonadError Error m
+  => Environment
+  -> State
+  -> AProposal ByteString
+  -> m State
+registerProposal env st proposal = do
+  let
+    pm = env ^. protocolMagic
```
ruhatch: To stay in line with the style of all the other validation rules, please unpack the environment and state in a where clause using `NamedFieldPuns`

ruhatch (changes requested):
The general structure looks good, but I don't want to use lenses in the implementation at all! Just need to recast in the style of the other rules so far.

src/Cardano/Chain/Update/Validation/Interface.hs
mhuesch: 👍  [`ProtocolUpdateProposals`](https://github.com/input-output-hk/cardano-chain/blob/8194af3534721859f4fc4b5246f6820bca2bd299/src/Cardano/Chain/Update/Validation/Registration.hs#L37)

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -0,0 +1,177 @@
...
+    , _adoptedProtocolVersion :: !ProtocolVersion
+    , _adoptedProtocolParams :: !ProtocolParameters
+      -- ^ Adopted protocol parameters
+    , _futureAdopts :: !TODO -- We should take this from Cardano.Chain.Update.Validation.Endorsement
+      -- ^ Future protocol version adoptions
+    , _appVersions :: !(Map ApplicationName (NumSoftwareVersion, FlatSlotId))
+      -- ^ Current application versions (by application name)
+    , _registeredProtocolUpdateProposals :: !(Map UpId (ProtocolVersion, ProtocolParameters))
+      -- ^ Registered protocol update proposals
+    , _registeredSoftwareUpdateProposals :: !(Map UpId SoftwareVersion)
```
mhuesch: There's also a synonym for this: [`SoftwareUpdateProposals`](https://github.com/input-output-hk/cardano-chain/blob/8194af3534721859f4fc4b5246f6820bca2bd299/src/Cardano/Chain/Update/Validation/Registration.hs#L38)

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -0,0 +1,177 @@
...
+    pv = st ^. adoptedProtocolVersion
+    pps = st ^. adoptedProtocolParams
+    avs = st ^. appVersions
+    dms = env ^. delegationMap
+    rpus = st ^. registeredProtocolUpdateProposals
+    raus = st ^. registeredSoftwareUpdateProposals
+    regSubSt = Registration.State rpus raus
+  Registration.State rpus' raus' <-
+    (Registration.registerProposal pm pv pps avs dms regSubSt proposal)
+      `wrapError` Registration
```
mhuesch, line 92: I like how this reads 🙂

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -0,0 +1,177 @@
...
+                      | (pid, sv) <- M.toList raus
+                      , pid `elem` M.keys cps'
+                      ]
+      raus = st ^. registeredSoftwareUpdateProposals
+  pure $! st
+       & confirmedProposals .~ cps'
+       & proposalVotes .~ vts'
+       & appVersions %~ M.union avsNew
+       & registeredSoftwareUpdateProposals %~ (`M.withoutKeys` S.fromAscList (M.keys cps))
+       -- TODO: consider using the `Relation` instances from `fm-ledger-rules` (see `Ledger.Core`)
```
mhuesch, line 144: Seems like a reasonable idea to me - we already depend on `cs-ledger`.

mhuesch (commented):
Looks good. I added a few comments. Since Ru requested de-lens-ing, I'll wait to approve.

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -0,0 +1,177 @@
...
+        M.fromList $! [ (svAppName sv, (svNumber sv, sn))
+                      | (pid, sv) <- M.toList raus
+                      , pid `elem` M.keys cps'
+                      ]
+      raus = st ^. registeredSoftwareUpdateProposals
+  pure $! st
+       & confirmedProposals .~ cps'
+       & proposalVotes .~ vts'
+       & appVersions %~ M.union avsNew
+       & registeredSoftwareUpdateProposals %~ (`M.withoutKeys` S.fromAscList (M.keys cps))
```
mhuesch: There is also [`keysSet`](https://www.stackage.org/haddock/lts-12.17/containers-0.5.11.0/Data-Map-Lazy.html#v:keysSet)

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -0,0 +1,177 @@
...
+  pure $! st
+       & registeredProtocolUpdateProposals .~ rpus'
+       & registeredSoftwareUpdateProposals .~ raus'
+       & proposalRegistrationSlot %~ M.insert (recoverUpId proposal) (env ^. currentSlot)
+
+-- | Register a vote for the given proposal.
+--
+-- If the proposal gets enough confirmations after adding the given vote, then
+-- it will get added to the set of confirmed proposals.
+--
```
mhuesch, line 121: In keeping with the other rules, we could write

```suggestion
-- This corresponds to the @UPIVOTE@ rules in the spec.
--
```

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -0,0 +1,177 @@
...
+  => Environment
+  -> State
+  -> AVote ByteString
+  -> m State
+registerVote env st vote = do
+  let pm = env ^. protocolMagic
+      subEnv = Voting.Environment sn pps (Voting.RegistrationEnvironment rups dms)
+      sn = env ^. currentSlot
+      pps = st ^. adoptedProtocolParams
+      rups = S.fromAscList $! M.keys (st ^. proposalRegistrationSlot)
```
mhuesch: There is also [`keysSet`](https://www.stackage.org/haddock/lts-12.17/containers-0.5.11.0/Data-Map-Lazy.html#v:keysSet)

mhuesch (commented):
I missed a few spots in my last review...

dnadales:
@ruhatch if we don't want to use lenses at all maybe we'd like to make this explicit in the README or CONTRIBUTING documents. 

I also noticed that we do depend on `lens` in `cardano-chain` (I didn't add this dependency), do we want to make an issue to remove this?

src/Cardano/Chain/Update/Validation/Interface.hs
dnadales: I don't like these synonyms: they introduce additional notation that the reader has to lookup. If we use the right-hand side of the type alias we are not adding much more characters and we remove one level of indirection.

@mhuesch, @ruhatch: any thoughts?

src/Cardano/Chain/Update/Validation/Interface.hs
dnadales: Thanks for this tip! I should have imagined that there must be a function already defined for this common pattern :man_facepalming:

src/Cardano/Chain/Update/Validation/Interface.hs
dnadales: I owe you a :beer:

src/Cardano/Chain/Update/Validation/Interface.hs
dnadales: Oh, I commented on this below. Please let me know what do you think.

src/Cardano/Chain/Update/Validation/Interface.hs
dnadales: Now it is a bit better without the parentheses :D

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -122,21 +99,32 @@ registerProposal
...
+    Environment { protocolMagic, currentSlot, delegationMap } = env
+    State
+      { adoptedProtocolVersion
+      , adoptedProtocolParams
+      , appVersions
+      , registeredProtocolUpdateProposals
+      , registeredSoftwareUpdateProposals
+      , proposalRegistrationSlot
+      } = st
+    pm = protocolMagic
```
dnadales: @ruhatch @mhuesch let me know if you're ok with these double bindings, otherwise I can split the call to `registerProposal` over several lines (or we can ask for 4K monitors :D )

src/Cardano/Chain/Update/Validation/Interface.hs
mhuesch: That's fair - the full types are fairly simple and terse, and they're only used once (as opposed to in `Cardano.Chain.Update.Validation.Registration` where they're repeated a few times).

Having the type inlined also makes it easier to compare directly with the types in the ledger-spec, which is what I did while reviewing.

dnadales: > Having the type inlined also makes it easier to compare directly with the types in the ledger-spec, which is what I did while reviewing.

Yeah, that's also a good point. Consistency is important. So we either drop the type synonyms in `Cardano.Chain.Update.Validation.Registration` or I add them in `Cardano.Chain.Update.Validation.Interface`. I'd prefer the latter, but I can also do the former :) @ruhatch, what do you think?

src/Cardano/Chain/Update/Validation/Interface.hs
ruhatch: If you want to rename them, just do:
```
Environment { protocolMagic = pm, currentSlot, delegationMap = dms }
```

Also, I'd still prefer the unpacking to be done in the `where` clause, as that allows us to read the meat of the definition and look up names as needed in some sort of index

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -0,0 +1,176 @@
...
+    subEnv = Voting.Environment sn pps (Voting.RegistrationEnvironment rups dms)
+    sn = currentSlot
+    pps = adoptedProtocolParams
+    rups = M.keysSet proposalRegistrationSlot
+    dms = delegationMap
+    subSt = Voting.State vts cps
+    vts = proposalVotes
+    cps = confirmedProposals
```
ruhatch: Again, renaming using records, and put into `where` clause

src/Cardano/Chain/Update/Validation/Interface.hs
ruhatch: How would you use `Relation` instances here?

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -0,0 +1,176 @@
...
+  , currentEpoch :: !EpochIndex
+  , currentSlot :: !FlatSlotId
+  , delegationMap :: !(Map StakeholderId StakeholderId)
+  }
+
+-- | Update interface state.
+data State
+  = State
+    { prevEpoch :: !EpochIndex
+      -- ^ Previously seen epoch
```
ruhatch: This is really pedantic, but could you move these back two?  Also if you have stylish-haskell, could you run it to line up the type definitions here?

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -0,0 +1,176 @@
...
+  , svNumber
+  )
+import qualified Cardano.Chain.Update.Validation.Registration as Registration
+import qualified Cardano.Chain.Update.Validation.Voting as Voting
+import Cardano.Chain.Update.Vote (UpId, AProposal, recoverUpId, AVote)
+import Cardano.Crypto (ProtocolMagicId)
+
+data TODO
+
+data Environment
```
ruhatch: I would prefer keeping the `= Environment` on the same line as `data Environment`, to keep in line with the rest of the code. You're right, I need to publish some style documents as I clearly have some styles!

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -0,0 +1,176 @@
...
+  , State (..)
+    -- *Error
+  , Error (..)
+    -- * Interface functions
+  , registerProposal
+  , registerVote
+  )
+where
+
+import Cardano.Prelude
```
ruhatch: Please don't use explicit imports in `Cardano.Prelude` - use `hiding` if something clashes

ruhatch (changes requested):

src/Cardano/Chain/Update/Validation/Interface.hs
dnadales: For the record, what I don't like of the 

```haskell
data MyData = MyData 
  { blah ...
```

style is that the fields end to the left of the constructor. 

I'll change this as suggested, but if we write some style document it'd be nice to ask others for their input, in particular @erikd, @edsko, and the more experienced folks around us.

src/Cardano/Chain/Update/Validation/Interface.hs
dnadales: I'll remove this, but again for the record, the problem with the implicit imports in `Cardano.Prelude` is that there are a bunch of non-standard functions what we do not know where do they come from (think of `wrapError` for instance). Tracking these down is quite hard (mostly because `Cardano.Prelude` will re-export a lot of things implicitly).

It is a pain to explicitly import symbols from `Cardano.Prelude`, but I think we're saving pain to our readers.

src/Cardano/Chain/Update/Validation/Interface.hs
dnadales: Sure.

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -139,38 +126,42 @@ registerVote
...
+      , delegationMap = dms
+      } = env
+
+    State
+      { adoptedProtocolParams = pps
+      , proposalRegistrationSlot
+      , proposalVotes = vts
+      , confirmedProposals = cps
+      , appVersions =  avs
+       , registeredSoftwareUpdateProposals = raus
```
dnadales: Ups I missed this one. I deleted this space and pushed the changes.

ruhatch (approved):
LGTM!

src/Cardano/Chain/Update/Validation/Interface.hs
ruhatch: Hmm, I think as long as everything else is using `qualified` or explicit imports, then it's relatively clear that it can only come from `Cardano.Prelude`

ruhatch: But I do understand and like with most things, I'm back and forth!

dnadales: > I think as long as everything else is using qualified or explicit imports, then it's relatively clear that it can only come from Cardano.Prelude

Yeah, that's a good point :)

dnadales:
bors retry

PR #357, opened 2019-03-14
[312][313] Build internal haddocks in CI

### LINKED ISSUE #312: Haddocks for cardano chain cannot be produced
We should remove the haddock errors (that might have carried over). Otherwise we cannot produce haddock documentation, and therefore we cannot use `hoogle`.

```
    src/Cardano/Chain/Update/Vote.hs:307:7: error:
        parse error on input ‘-- ^ This byte is part of the signed payload, but is not part of the transmitted payload.
          -- This is an implementation artifact of the previous’
        |
    307 |       -- ^ This byte is part of the signed payload, but is not part of the transmitted payload.
        |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```

### LINKED ISSUE #313: Check `haddock` output on CI
We should check that `haddock` can generate documentation for our project in CI. By being able to produce documentation we are also able to use `hoogle`, which I find helps a lot when ramping up with a new code-base.

mhuesch:
Closes #312
Closes #313

### DIFF
--- a/scripts/buildkite/rebuild.hs
+++ b/scripts/buildkite/rebuild.hs
@@ -58,7 +58,13 @@ buildStep testArgs = do
  where
   cfg = ["--dump-logs", "--color", "always"]
   stackBuild args = run "stack" $ cfg ++ ["build", "--fast"] ++ args
-  buildArgs    = ["--bench", "--no-run-benchmarks", "--no-haddock-deps"]
+  buildArgs =
+    [ "--bench"
+    , "--no-run-benchmarks"
+    , "--haddock"
+    , "--haddock-internal"
+    , "--no-haddock-deps"
+    ]
   buildAndTest = stackBuild $ ["--tests"] ++ buildArgs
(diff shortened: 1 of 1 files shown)


scripts/buildkite/rebuild.hs
```diff
@@ -58,7 +58,8 @@ buildStep testArgs = do
  where
   cfg = ["--dump-logs", "--color", "always"]
   stackBuild args = run "stack" $ cfg ++ ["build", "--fast"] ++ args
-  buildArgs    = ["--bench", "--no-run-benchmarks", "--no-haddock-deps"]
+  buildArgs    = [ "--bench", "--no-run-benchmarks", "--haddock"
```
dnadales: Apparently brittany doesn't like this :man_shrugging: Looks ok to me though ...

dnadales (approved):
Thank you Michael!

PR #358, opened 2019-03-14
[#285] Implement bulk chain validation with logging and UTxO size estimation

### LINKED ISSUE #285: Initial cardano-shell integration
To include:

 - Corresponding blockchain layer and feature
 - An application that uses them to verify existing blockchain

To be added later:

 - Integration with the logging layer

### LINKED ISSUE #387: Log UTxO size during bulk chain validation
_Relates to issue #285 and PR #358._ 

We'd like to add functionality for logging the current `UTxO` size after processing each block during bulk chain validation. We're most interested in the `Data.Map.size` as well as the size in `Cardano.Prelude.HeapWords`. We would also want to hide the printouts behind a conditional so they only fire during debugging.

Jimbo4350:
This PR aims to: 

- Incorporate `cardano-shell`'s logging capability into `cardano-ledger`. 
- Implement bulk chain validation and `UTxO` size estimation into the blockchain feature.

`cardano-shell` exposes the [logging feature](https://github.com/input-output-hk/cardano-shell/blob/develop/app/Cardano/Shell/Features/Logging.hs) which is used in conjunction with the blockchain feature (defined in `cardano-ledger`) to log the output of bulk chain validation and `UTxO` size estimation.

Relevant background: https://github.com/input-output-hk/cardano-shell/blob/develop/README.md
Closes #285 
Closes #387


src/Cardano/Chain/Epoch/Validation.hs
```diff
@@ -0,0 +1,78 @@
...
+    fst
+    stream
+  where stream = parseEpochFileWithBoundary fp
+
+-- | Check that a list of epochs' `Block`s are valid.
+validateEpochFiles
+  :: Genesis.Config
+  -> ChainValidationState
+  -> [FilePath]
+  -> IO (Either EpochError ChainValidationState)
```
ksaric, line 52: I don't want to go into details before other folks give a review, but wouldn't it be simpler to just use `MonadThrow/MonadCatch m` here?

app/Main.hs
```diff
@@ -0,0 +1,72 @@
...
+  let
+    chainValidationApp
+      :: LoggingLayer -> BlockchainLayer -> CardanoFeature -> IO ()
+    chainValidationApp ll bcl bf = do
+      logTrace' <- appendName "cardano-chain" logTrace
+      let
+        loop = chainValidationStatus bcl >>= \case
+          Nothing ->
+            logNotice logTrace' "No results yet!"
+              >> threadDelay 10000000
```
ksaric: 10 seconds, right? Hard to guess when they are all crammed in like that.

src/Cardano/Shell/Features/Blockchain.hs
```diff
@@ -0,0 +1,97 @@
...
+import Cardano.Shell.Types
+  (ApplicationEnvironment(..), CardanoEnvironment, CardanoFeature(..))
+
+-- `BlockchainLayer` enables us to get the value from an MVar
+-- which holds the result of the bulk chain validation.
+data BlockchainLayer = BlockchainLayer
+  { chainValidationStatus
+      :: forall m
+       . MonadIO m
+      => m (Maybe (Either EpochError ChainValidationState))
```
ksaric: Again, seems like a simple `MonadCatch` would do here.

src/Cardano/Shell/Features/Blockchain.hs
```diff
@@ -0,0 +1,97 @@
...
+   . MonadIO m
+  => BlockchainConfiguration
+  -> ApplicationEnvironment
+  -> ChainValidationState
+  -> MVar (Either EpochError ChainValidationState)
+  -> m ()
+init config appEnv initialCVS resultVar = do
+  -- Validate epoch files
+  files <- case appEnv of
+    Development -> take 10 <$> liftIO mainnetEpochFiles
```
ksaric: Take the first 10 mainnet epoch files?

src/Cardano/Chain/Epoch/Validation.hs
```diff
@@ -0,0 +1,78 @@
...
+  where stream = parseEpochFilesWithBoundary fps
+
+-- | Fold chain validation over a 'Stream' of 'Blund's
+foldChainValidationState
+  :: Genesis.Config
+  -> ChainValidationState
+  -> Stream (Of (ABlockOrBoundary ByteString)) (ExceptT ParseError ResIO) ()
+  -> ExceptT EpochError ResIO ChainValidationState
+foldChainValidationState config cvs blocks = S.foldM_
+  (\c b ->
```
ksaric: Write down a function for readability?

src/Cardano/Chain/Epoch/Validation.hs
```diff
@@ -0,0 +1,78 @@
...
+    fst
+    stream
+  where stream = parseEpochFilesWithBoundary fps
+
+-- | Fold chain validation over a 'Stream' of 'Blund's
+foldChainValidationState
+  :: Genesis.Config
+  -> ChainValidationState
+  -> Stream (Of (ABlockOrBoundary ByteString)) (ExceptT ParseError ResIO) ()
+  -> ExceptT EpochError ResIO ChainValidationState
```
ksaric, line 62: :scream:

src/Cardano/Chain/Epoch/Validation.hs
```diff
@@ -0,0 +1,78 @@
...
+validateEpochFiles
+  :: Genesis.Config
+  -> ChainValidationState
+  -> [FilePath]
+  -> IO (Either EpochError ChainValidationState)
+validateEpochFiles config cvs fps =
+  runResourceT . runExceptT . foldChainValidationState config cvs $ S.map
+    fst
+    stream
+  where stream = parseEpochFilesWithBoundary fps
```
ksaric: Repetition?

app/Main.hs
```diff
@@ -0,0 +1,72 @@
...
+  -- Application we are going to execute.
+  let
+    chainValidationApp
+      :: LoggingLayer -> BlockchainLayer -> CardanoFeature -> IO ()
+    chainValidationApp ll bcl bf = do
+      logTrace' <- appendName "cardano-chain" logTrace
+      let
+        loop = chainValidationStatus bcl >>= \case
+          Nothing ->
+            logNotice logTrace' "No results yet!"
```
ruhatch: Could just use `do` notation here

app/Main.hs
```diff
@@ -0,0 +1,72 @@
...
+              >> threadDelay 10000000
+              >> loop
+          Just results -> case results of
+            Left epochError -> logNotice logTrace' $ show epochError
+            Right _ ->
+              logNotice logTrace' "Epoch validation successful!" >> exitSuccess
+      logNotice logTrace' "Begin validating epoch files..."
+      loop
+      featureStart bf
```
ruhatch: I think you don't need to do this part - the `runCardanoApplicationWithFeatures` is supposed to start all the features for you!

app/Main.hs
```diff
@@ -0,0 +1,72 @@
...
+    cardanoEnvironment
+    cardanoConfiguration
+  (blockchainLayer, blockchainFeature) <- createBlockchainFeature
+    cardanoEnvironment
+    cardanoConfiguration
+    Development
+
+  -- Application we are going to execute.
+  let
+    chainValidationApp
```
ruhatch: Maybe move this to the top level (i.e. out of the `main` function) to keep the definitions smaller

cardano-chain.cabal
```diff
@@ -25,13 +25,16 @@ library
...
                        Cardano.Chain.Epoch.File
+                       Cardano.Chain.Epoch.Validation
                        Cardano.Chain.Genesis
                        Cardano.Chain.ProtocolConstants
                        Cardano.Chain.Slotting
                        Cardano.Chain.Ssc
                        Cardano.Chain.Txp
                        Cardano.Chain.Update
 
+                       Cardano.Shell.Features.Blockchain
```
ruhatch: Aha, could you actually move this to the application, I don't think it should be in the `cardano-chain` library

cardano-chain.cabal
```diff
@@ -121,9 +124,12 @@ library
                      , canonical-json
                      , cardano-binary
                      , cardano-crypto-wrapper
+                     , cardano-mainnet-mirror
```
ruhatch: Then you can move this dep to the application too, I really don't want this in the library dependencies!

cardano-chain.cabal
```diff
@@ -153,6 +159,27 @@ library
   if (!flag(development))
     ghc-options:         -Werror
 
+executable cardano-chain-validation-exe
+  main-is: Main.hs
+  hs-source-dirs:
+      app
+  build-depends:
+      base >=4.7 && <5
```
ruhatch: Can you align these like the others in this file?

src/Cardano/Shell/Features/Blockchain.hs
```diff
@@ -0,0 +1,97 @@
...
+
+import Cardano.Chain.Block (initialChainValidationState, ChainValidationState)
+import Cardano.Chain.Epoch.Validation (EpochError, validateEpochFiles)
+import qualified Cardano.Chain.Genesis.Config as Genesis
+import Cardano.Mirror (mainnetEpochFiles)
+import Cardano.Shell.Constants.Types (CardanoConfiguration)
+import Cardano.Shell.Types
+  (ApplicationEnvironment(..), CardanoEnvironment, CardanoFeature(..))
+
+-- `BlockchainLayer` enables us to get the value from an MVar
```
ruhatch: You can use `-- | ...` to make this nice for Haddock

src/Cardano/Shell/Features/Blockchain.hs
```diff
@@ -0,0 +1,97 @@
...
+          (Genesis.mkConfigFromFile "test/mainnet-genesis.json" Nothing)
+
+  let
+    blockchainConf =
+      BlockchainConfiguration "cardano-mainnet-mirror/epochs" config
+
+  -- Create initial `ChainValidationState`.
+  initCVS <- either (panic . show) identity
+    <$> runExceptT (initialChainValidationState config)
+  -- Create MVar that will hold the result of the bulk chain validation.
```
ruhatch: Could use a bit more white space here (i.e. newlines above the comments on the next couple of lines)

ruhatch (changes requested):
Structure looks good, just a few formatting/style suggestions. Are you going to change the config stuff in this PR or do another one for that?

ksaric (commented):
I'll add my own comments here, but they are optional, of course, since my jurisdiction ends when you cross the `cardano-shell` county.

src/Cardano/Chain/Epoch/Validation.hs
ruhatch: Yeah, although we prefer `MonadError`, so `(MonadIO m, MonadError EpochError m) => ...`

app/Main.hs
ruhatch: Yeah, I'd suggest `NumDecimals`, so you can write `1e7`

src/Cardano/Shell/Features/Blockchain.hs
ruhatch: I'm not sure about this one actually, I'd probably leave it as it is

cardano-chain.cabal
Jimbo4350: Will do!

src/Cardano/Shell/Features/Blockchain.hs
Jimbo4350: @ksaric what's your rationale here?

src/Cardano/Chain/Epoch/Validation.hs
mhuesch: Jordan asked me for input on this -

I don't see that `MonadError` gets us anything here. We have a function which _could_ be generalized, but is only used [here](https://github.com/input-output-hk/cardano-chain/pull/358/files#diff-fed5b16ab90a046803ddac97cd51193dR59) at its specialized type, in order to return an `Either EpochError ChainValidationState` which goes into an MVar, which is specialized to that type.

So, generalizing in order to specialize doesn't make sense to me.

I agree that `MonadError` is useful in many situations, because of ease of composition, etc. But I don't see that here.

src/Cardano/Shell/Features/Blockchain.hs
mhuesch: If we keep this as a concrete `Either EpochError ChainValidationState`, then I don't see the point in generalizing [above](https://github.com/input-output-hk/cardano-chain/pull/358#discussion_r265922486), since we would just have to re-specialize [here](https://github.com/input-output-hk/cardano-chain/blob/426db475a1f3dae561af9b81075c8cc68bb1b38f/app/Features/Blockchain.hs#L59).

And I don't really see the point in generalizing this part, since there's only one usage site for the `chainValidationStatus` accessor. In short, I think leaving them both as concrete `Either EpochError ChainValidationState` types is best.

ksaric: This is up to you guys to decide, just wrote down my opinion.
My rationale here, and in general, when it's concerning `Either` is that it seems to me to be good for smaller chunks of code, that don't propagate over several functions.
My experience is that `Either` doesn't compose nicely into the transformer stack and when you think about it, all the functions where you drag `Either` are functions which you "infect" with that "effect", which means you "need" transformer stack composition there anyway.

cardano-chain.cabal
```diff
@@ -122,8 +123,10 @@ library
                      , cardano-binary
                      , cardano-crypto-wrapper
                      , cardano-prelude
+                     , cardano-shell
```
ruhatch: You should be able to remove these now, right?

cardano-chain.cabal
```diff
@@ -152,6 +155,31 @@ library
...
+                     , cardano-chain
+                     , cardano-mainnet-mirror
+                     , cardano-shell
+                     , cardano-prelude
+                     , formatting
+
+  default-language:    Haskell2010
+  default-extensions:  NoImplicitPrelude
+
+  ghc-options:         -Wall
```
ruhatch: See the other sections of this cabal file for the standard `ghc-options` we use

test/Test/Cardano/Chain/Block/Validation.hs
```diff
@@ -40,10 +40,10 @@ import qualified Hedgehog.Range as Range
 
 import Cardano.Chain.Block
   ( ABlockOrBoundary(..)
+  , blockSlot
```
ruhatch: Can you revert the changes to this file, they don't seem particularly relevant!

ruhatch (approved):
Looks good, just a couple of things to address!

src/Cardano/Shell/Features/Blockchain.hs
Jimbo4350: Decided to keep things explicit for now, the `Either` isn't propagating very far at the moment and if we need to generalize in the future we can.

test/Test/Cardano/Chain/Block/Validation.hs
Jimbo4350: Needed here: https://github.com/input-output-hk/cardano-chain/pull/358/files#diff-8fb0ac17f82ff048bc35a27ffe8a766aR78

ruhatch: From what I can see this just moves an import around and removes some whitespace? Can you see something different?

src/Cardano/Chain/Epoch/Validation.hs
erikd: i personally consider `MonadThrow/MonadCatch` much more painful than simply dealing with the  `Either`.

In my experience, the only people who prefer `MonadError/MonadThrow/MonadCatch` are the people who don't know how to use `EitherT/ExceptT` and/or do not have access to a suitable library of useful combinators for `EitherT/ExceptT`.

src/Cardano/Shell/Features/Blockchain.hs
erikd: Sorry, I *really* have to disagree here, `Either` composes far better than any `Monad*` typeclass constraint, as long as you are aware of and use the right techniques.

The problem is that the vast majority of the publicly available example code uses the `Monad*` stuff and there is almost no code that uses what I consider a better alternative.

@ksaric if you ever get into a situation where you consider `Either` a problem, please let me know and I will help you sort it out.

src/Cardano/Chain/Epoch/Validation.hs
erikd: What is the problem here???? As far as I am concerned that is best practice!

ksaric: I guess it comes down to this - it is my impression that using `mtl` is the "standard" and recommended way of using transformer stacks. Getting out of the whole "uh, I forgot to add another `lift` to `lift . lift . lift $`". 
So you prefer not to use `mtl` and like using this style, I presume?

test/Test/Cardano/Chain/Block/Validation.hs
Jimbo4350: :man_facepalming: yep, overlooked this!

src/Cardano/Chain/Epoch/Validation.hs
erikd: I think we are talking about completely different things.

I am specifically talking about a preference for using `ExceptT` (with the right combinators) in preference to `MonadCatch/MonadThrow`. Using `ExceptT` makes errors far more explicit.

app/Features/Blockchain.hs
```diff
@@ -0,0 +1,143 @@
...
+
+-- | `BlockchainLayer` provides a window of sorts that
+-- enables us to access values from various 'MVar's
+-- that hold the results of specific applications.
+data BlockchainLayer = BlockchainLayer
+  { chainValidationStatus
+      :: forall m
+       . MonadIO m
+      => m (Maybe (Either EpochError ChainValidationState))
+  , currentUTxOSize
```
ruhatch, line 40: I think it might be better to do the logging for this one internal to the `BlockchainLayer`. This would also show what's it's like to have the `LoggingLayer` as a dependency of the `BlockchainFeature`. I think there's an example in `cardano-shell` that shows what it would look like to have one feature take the other as a dependency. So then you would just output `UTxO` size information after each block from inside the layer.

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -287,3 +319,76 @@ updateBlock config cvs b = do
...
+  -> MVar (HeapSize UTxO, UTxOSize, SlotId)
+  -> UTxO
+  -> ABlock ByteString
+  -> ExceptT Error ResIO UTxO
+applyBlockUTxOMVar pm sizeMVar utxo block = do
+  resResult <- liftIO . runResourceT . runExceptT $ foldUTxOBlock pm utxo block
+  case resResult of
+    Left e -> throwE e
+    Right updatedUtxo -> do
+              _ <- liftIO $ takeMVar sizeMVar
```
ruhatch, line 375: This is slightly odd, I don't think you should be `take`ing from this `MVar`, because it can only be consumed by one thread. In this case the other thread might read it before you and then you'd block here forever! If you want to wait for them to read it out then `putMVar` does exactly that, it blocks until the `MVar` is empty.

ruhatch (approved):
Looks good! I'd suggest removing the whole `MVar` thing for the UTxO size and just use the logging layer internally, but maybe leave that for another PR!

app/Features/Blockchain.hs
Jimbo4350: Yep that sounds reasonable. I can do that in a follow up PR.

src/Cardano/Chain/Block/Validation.hs
Jimbo4350: This is read with `tryReadMVar` in the blockchain layer so it shouldn't block because `tryReadMVar`doesn't empty the `MVar`. I agree this design could be improved though!

app/Features/Blockchain.hs
```diff
@@ -0,0 +1,143 @@
...
+
+    -- Validate epoch files.
+
+    -- Currently `validateEpochFiles` is restricted to the first 5 epochs.
+    result <- liftIO $ validateEpochFiles (genesisConfig config) initialCVS (take 5 files)
+    liftIO $ putMVar cvsVar result
+
+    -- Output UTxO size
+    let genesisData = configGenesisData $ genesisConfig config
+    let pm = ProtocolMagic (gdProtocolMagicId genesisData) RequiresNoMagic
```
mhuesch: I think this was added after you started the PR, but you could grab the `ProtocolMagic` out of `Genesis.Config` using https://github.com/input-output-hk/cardano-ledger/blob/8ef7bf8049bbcd201ef19c4038ae42ed79722be6/src/Cardano/Chain/Genesis/Config.hs#L161

app/Features/Blockchain.hs
```diff
@@ -0,0 +1,143 @@
...
+      , featureShutdown = cleanup
+      }
+
+  -- Create `BlockchainLayer` which allows us to see the status of
+  -- the blockchain feature.
+  let
+    bcLayer =
+      BlockchainLayer
+        { chainValidationStatus = liftIO $ tryReadMVar cvsVar
+        , currentUTxOSize       = liftIO $ tryReadMVar utxoSizeVar
```
mhuesch, line 137: I _think_ this could be significantly simplified if a Queue datatype (e.g. [`TBQueue`](https://www.stackage.org/haddock/lts-13.4/stm-2.5.0.0/Control-Concurrent-STM-TBQueue.html)) were used instead of an MVar. That approach would also eliminate the need for an initial dummy value, since the queue could just be initialized as empty.

app/Main.hs
```diff
@@ -0,0 +1,116 @@
...
+    utxoSizeAfterBlockLoop lgTrace =
+      currentUTxOSize bcl >>= \case
+        Nothing -> do
+          logNotice lgTrace "No results yet for UTxO size!"
+          threadDelay 1e6
+          utxoSizeAfterBlockLoop lgTrace
+        Just size -> do
+          logNotice lgTrace $ show size
+          threadDelay 1e6
+          utxoSizeAfterBlockLoop lgTrace
```
mhuesch, line 111: This retry logic could be cut out and a simple blocking read used, such as [`readTBQueue`](https://www.stackage.org/haddock/lts-13.4/stm-2.5.0.0/Control-Concurrent-STM-TBQueue.html#v:readTBQueue)

This should address Ru's concern about MVar coordination to avoid deadlock, also.

src/Cardano/Chain/Block/Validation.hs
mhuesch: I think a Queue datatype (see my other comment) would clean this up. It would remove the concern about synchronization & deadlocks. If, for example, write was missed, or a read was missed, neither end would get "stuck".

src/Cardano/Chain/Common/NetworkMagic.hs
```diff
@@ -37,3 +38,8 @@ makeNetworkMagic :: ProtocolMagic -> NetworkMagic
 makeNetworkMagic pm = case getRequiresNetworkMagic pm of
     RequiresNoMagic -> NetworkMainOrStage
     RequiresMagic   -> NetworkTestnet (getProtocolMagic pm)
+
+parseReqNetworkMag :: Text -> RequiresNetworkMagic
+parseReqNetworkMag "NetworkMainOrStage" = RequiresNoMagic
+parseReqNetworkMag _                    = RequiresMagic
```
mhuesch, line 44: Ack. Is there not a more-typed way to do this? Granted, that probably requires changes `cardano-shell` datatypes.

At the very least, I think we could explicitly match both cases and signal an error somehow (`panic` perhaps) on unrecognized input.

test/Test/Cardano/Chain/Config.hs
```diff
@@ -21,13 +21,8 @@ import Cardano.Chain.Slotting (EpochSlots(EpochSlots))
...
       identity
     <$> runExceptT
-          (Genesis.mkConfigFromFile RequiresNoMagic "test/mainnet-genesis.json" Nothing)
+          (Genesis.mkConfigFromFile RequiresNoMagic "mainnet-genesis.json" Nothing)
 
--- | Slots per epoch used in mainnet
---
--- This number has been fixed throughout the Byron era.
-mainnetEpochSlots :: EpochSlots
-mainnetEpochSlots = EpochSlots 21600
```
mhuesch, line 33: Assuming this was moved from `test/` because it was needed in the `src/` tree.

stack.yaml
```diff
@@ -32,8 +32,27 @@ extra-deps:
   - sequence-0.9.8
   - tasty-hedgehog-0.2.0.0
 
-  # To be able to use `stack hoogle`
-  - haskell-src-exts-1.21.0
```
mhuesch: Do we want to get rid of this?

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -188,6 +209,17 @@ data ChainValidationError
...
 
+updateChainBlockOrBoundary
+  :: MonadError ChainValidationError m
+  => Genesis.Config
+  -> ChainValidationState
+  -> ABlockOrBoundary ByteString
+  -> m ChainValidationState
+updateChainBlockOrBoundary config c b = case b of
+  ABOBBoundary bvd   -> updateChainBoundary config c bvd
+  ABOBBlock    block -> updateBlock config c block
```
mhuesch, line 220: 👍

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -287,3 +319,76 @@ updateBlock config cvs b = do
...
+-- | Fold 'updateUTxO' over the transactions in a single 'Block'
+foldUTxOBlock
+  :: ProtocolMagic -> UTxO -> ABlock ByteString -> ExceptT Error ResIO UTxO
+foldUTxOBlock pm utxo block =
+  withExceptT (ErrorUTxOValidationError . unflattenSlotId mainnetEpochSlots $ blockSlot block)
+    $ foldM (updateUTxOWitness pm) utxo (aUnTxPayload $ blockTxPayload block)
+
+newtype HeapSize a =
+  HeapSize { unHeapSize :: Int}
+  deriving Show
```
mhuesch, line 354: It would be nice to add a docstring, something like
```
-- | Size of a heap value, in words
```

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -287,3 +319,76 @@ updateBlock config cvs b = do
...
+
+newtype UTxOSize =
+  UTxOSize { unUTxOSize :: Int}
+  deriving Show
```
mhuesch, line 359: It would be nice to add a docstring, something like
```
-- | Number of entries in the UTxO
```

mhuesch (commented):
I don't want to block merge since this has been a long-running issue and is great work, modulo a few things I'd like to see tweaked.

Happy to have these go in a separate PR, if we decide their good suggestions.

app/Main.hs
erikd: Possibly doesn't even need a `TBQueue`, a `TVar` should do the trick as well and avoid `MVar` deadlocks.

Jimbo4350: I will do some investigation here and submit a follow up PR.

src/Cardano/Chain/Common/NetworkMagic.hs
Jimbo4350: Totally agree. We need to reconcile the [`CardanoConfiguration`](https://github.com/input-output-hk/cardano-shell/blob/develop/src/Cardano/Shell/Constants/Types.hs#L36) type with the `Config` type that exists in `cardano-ledger`. I'll raise an issue in both repos and link them.

test/Test/Cardano/Chain/Config.hs
Jimbo4350: It made more sense to be in `src/` imo

stack.yaml
Jimbo4350: I missed this, thanks for pointing it out!

src/Cardano/Chain/Common/NetworkMagic.hs
mhuesch: Sounds good 👍

app/Main.hs
mhuesch: To check my understanding of the various STM constructs: `TVar` would enforce coupling between the size-logger and the size-printer processes. Whereas `TBQueue` would give them a buffer to get slightly out of sync with each other. Is that right @erikd?

In this case it seems there's no harm in them being in sync.

app/Features/Blockchain.hs
mhuesch: If we use a TVar we would still require a dummy value, as well.

PR #359, opened 2019-03-14
Update issue templates

ruhatch:
I would like to try out the following issue template as part of my GitHub Issue Wrangler role in the Ledger Circle.


dnadales (approved):
Looks good, but I'm just wondering what happens if an issue cannot be described according to the template. For instance I can imagine that people might not always know what to fill in as proposed solution.

But if the template is not compulsory I guess we can delete the parts that are irrelevant or we don't know how to fill.

ruhatch:
Yeah, I should add a comment in the template to suggest leaving the solution part, even if you can't think of one and adding a label that it's needed, like `needs-planning`

PR #361, opened 2019-03-15
[#356] Remove AddrStakeDistribution

### LINKED ISSUE #356: Remove `AddrStakeDistribution`
## Background

The `AddrStakeDistribution` type would have been used to represent different ways of assigning stake in Ouroboros Classic. It was never actually used on chain, apart from the `BootstrapEraDistr` nullary constructor. We will not use this type in the Shelley era as our address format will be changing, so we should simply drop the whole thing.

## Proposed Solution

We should remove this type entirely. To preserve compatibility with cardano-sl, we need to decode and encode the empty list, which was used to represent the `BootstrapEraDistr` constructor.

ruhatch:
Closes #356


dnadales (approved):

src/Cardano/Chain/Common/AddrAttributes.hs
```diff
@@ -83,22 +69,13 @@ instance Bi (Attributes AddrAttributes) where
...
     go n v acc = case n of
-      0 -> (\distr -> Just $ acc { aaStakeDistribution = distr })
-        <$> Bi.deserialize v
       1 -> (\deriv -> Just $ acc { aaPkDerivationPath = Just deriv })
         <$> Bi.deserialize v
       _ -> pure Nothing
-
-instance HeapWords AddrAttributes where
-  heapWords (AddrAttributes derivationPath stakeDistr) =
-    heapWords2 derivationPath stakeDistr
```
mhuesch: Encoding and decoding of `AttrAttributes` looks to be `Bi` compatible with the old format 👍

mhuesch, line 104: Encoding and decoding of AttrAttributes looks to be Bi compatible with the old format 👍

mhuesch (commented):
Late to the party - LGTM

PR #364, opened 2019-03-17
Rename SlotCount to EpochSlots

erikd:
The type is related to slots and counts, but the name `SlotCount` gave zero hint as to what this type actually represents. The new name is far more likely to help the code reader figure out its purpose.

Also get rid of some potentially dangerous (and totally un-needed) typeclass instances like `Num`, and `Integral`.

Deriving these typeclasses is dangerous because it allows implicit conversion which in turn allows operations on otherwise incompatible types which is *exactly* what `newtype`s are supposed to prevent. For instance if types `X` and `Y` both have `Num` and `Integral` instances this will compile:
```
    wibble = (10 :: X) + (10 :: Y)
````
and be as meaningful as doing addition one two quantities, one in meters and the other in seconds.

Explicit is always better than implicit

### CODE REVIEW

File: src/Cardano/Chain/Slotting/EpochSlots.hs [hunk 1]
```diff
@@ -0,0 +1,26 @@
...
+
+import Cardano.Prelude
+
+import Data.Aeson (ToJSON(..))
+import Formatting.Buildable (Buildable)
+
+import Cardano.Binary.Class (Bi(..))
+
+-- | The number of slots per epoch.
+newtype EpochSlots = EpochSlots
```
dnadales, line 18: Wouldn't `SlotsPerEpoch` be a more appropriate name? I'm not a native English speaker though, so that might be the reason why it took me a bit to parse `EpochSlots`.

src/Cardano/Chain/Slotting/LocalSlotIndex.hs
```diff
@@ -24,40 +24,28 @@ import Cardano.Prelude
...
-             , Generic
-             , B.Buildable
-             , NFData
-             , Num
-             , Integral
-             , Real
-             , Enum
-             )
+  { unLocalSlotIndex :: Word16
+  } deriving (Eq, Ord, Generic, NFData, Show, B.Buildable)
```
dnadales, line 37: :tada:

dnadales (approved):
:sparkles:

ruhatch (approved):
:+1: - my only concern is removing the overflow errors from `flattenSlotId`, but I guess you're right that it will never be a real problem

dnadales:
@ruhatch we need to rebase for `bors` to succeed. See https://github.com/input-output-hk/cardano-chain/pull/366

ruhatch:
Oh, does `bors` use hydra and buidlkite?

PR #365, opened 2019-03-18
Concretise the type inside the `GenesisHash` newtype wrapper

### LINKED ISSUE #332: Concretise the type inside the `GenesisHash` newtype wrapper
## Background

`GenesisHash` is defined as

```
newtype GenesisHash = GenesisHash { getGenesisHash :: forall a . Hash a }
```

To construct a value of this type one must use `coerce` from `Data.Coerce`. There's no obvious justification for this complicated approach.

## Proposed Solution

A `GenesisHash` should really be a hash of the bytes of the genesis configuration data. This is what we refer to as the genesis block. We use the type `Raw` to represent `ByteString`s that shouldn't be serialised before `hash`ing, so we should simply change the definition to use a `Hash Raw`.

dnadales:
Closes #332

### DIFF
--- a/src/Cardano/Chain/Block/Header.hs
+++ b/src/Cardano/Chain/Block/Header.hs
@@ -47,6 +47,9 @@ module Cardano.Chain.Block.Header
   , ConsensusData
   , consensusData
   , verifyConsensusData
+
+  -- * Utility functions
+  , genesisHeaderHash
   )
 where
 
@@ -54,6 +57,7 @@ import Cardano.Prelude
 
 import Control.Monad.Except (MonadError, liftEither)
 import qualified Data.ByteString as BS
+import Data.Coerce (coerce)
 import Formatting (Format, bprint, build, int)
 import qualified Formatting.Buildable as B
 
@@ -202,11 +206,15 @@ mkHeader
   -> Header
 mkHeader pm prevHeader = mkHeaderExplicit pm prevHash difficulty
  where
-  prevHash   = either unGenesisHash hashHeader prevHeader
+  prevHash   = either genesisHeaderHash hashHeader prevHeader
   difficulty = either
     (const 0)
     (succ . consensusDifficulty . headerConsensusData)
     prevHeader
+
+-- | Extract the genesis hash and cast it into a header hash.
+genesisHeaderHash :: GenesisHash -> HeaderHash
+genesisHeaderHash = coerce . unGenesisHash
 
 -- | Make a 'Header' for a given slot, with a given body, parent hash,
 --   and difficulty. This takes care of some signing and consensus data.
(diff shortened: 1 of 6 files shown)


src/Cardano/Chain/Block/Header.hs
```diff
@@ -202,12 +206,16 @@ mkHeader
...
  where
-  prevHash   = either unGenesisHash hashHeader prevHeader
+  prevHash   = either genesisHeaderHash hashHeader prevHeader
   difficulty = either
     (const 0)
     (succ . consensusDifficulty . headerConsensusData)
     prevHeader
 
+-- | Extract the genesis hash and cast it into a header hash.
+genesisHeaderHash :: GenesisHash -> HeaderHash
```
ruhatch, line 216: Neat - the `coerce` is in one spot :D

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -194,7 +198,7 @@ updateChainBoundary
 updateChainBoundary config cvs bvd = do
   let
     prevHash =
-      fromMaybe (unGenesisHash $ configGenesisHash config) (cvsPreviousHash cvs)
+      fromMaybe (genesisHeaderHash $ configGenesisHash config) (cvsPreviousHash cvs)
```
ruhatch: Might even be worth defining a `configGenesisHeaderHash` because this pattern is used three times, not sure though

ruhatch (approved):
LGTM! Paying off that technical debt :)

src/Cardano/Chain/Block/Validation.hs
dnadales: Yeah, I thought that too! I saw it only two times in our `src`, so it was [less than 3](https://blog.codinghorror.com/rule-of-three/), but since we're using in in our test code as well should I just define a function:

```haskell
configGenesisHeaderHash :: Genesis.Config -> HeaderHash
```

src/Cardano/Chain/Genesis/Hash.hs
```diff
@@ -8,11 +8,12 @@ where
...
 import Cardano.Prelude
 
+import Cardano.Binary.Class (Raw)
 import Cardano.Crypto.Hashing (Hash)
 
 
 newtype GenesisHash = GenesisHash
-  { unGenesisHash :: forall a. Hash a
+  { unGenesisHash :: Hash Raw
   }
```
Jimbo4350, line 17: :tada:

Jimbo4350 (approved):
LGTM!

src/Cardano/Chain/Block/Validation.hs
ruhatch: Yeah, I reckon that would be nice - you can put it in `Genesis.Config` with the other config accessors

ruhatch (approved):
LGTM!

PR #366, opened 2019-03-18
Remove cardano-chain-test.x86_64-darwin from required hydra jobs

jbgi:
this job is not built by hydra and should have been removed from
 the required aggregate job in #354, otherwise the required
 job never finish.
Other PRs need to be rebased on master once this is merged.


dnadales (approved):
Thank you @jbgi!

As a side note it'd be nice if we'd have tools in place to detect these kind of dependencies.

PR #367, opened 2019-03-18
Remove brittany from buildkite pipeline for now

### LINKED ISSUE #340: Do we want to keep using brittany?
As programmers we should use our good judgement in formatting the code, optimizing it for readability. If we rely on automatic tools we get results like the following:

image

where the arguments to a function appear below and to the left of the function name. Or results like this one:

image

which produces a weird `case` alignment just because the tool wants to obey the `80` characters limit.

Paraphrasing Edsko:

> We take great care in our layout, which we use to emphasize the structure.

So we shouldn't let a tool take precedence over our sense of judgement.

But like always, feel free to close this issue if the majority wants to keep `brittany` :)

ruhatch:
Closes #340

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -16,12 +16,12 @@ steps:
     agents:
       system: x86_64-linux
 
-  - label: 'brittany'
-    command:
-      - "nix-build scripts/brittany -o check-brittany"
-      - "./check-brittany"
-    agents:
-      system: x86_64-linux
+  # - label: 'brittany'
+  #   command:
+  #     - "nix-build scripts/brittany -o check-brittany"
+  #     - "./check-brittany"
+  #   agents:
+  #     system: x86_64-linux
 
   - label: 'nix-tools'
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -16,12 +16,12 @@ steps:
     agents:
       system: x86_64-linux
 
-  - label: 'brittany'
-    command:
-      - "nix-build scripts/brittany -o check-brittany"
-      - "./check-brittany"
-    agents:
-      system: x86_64-linux
+  # - label: 'brittany'
```
dnadales, line 19: Some people would object "why commenting this out if we have version control" ;) I would have commented this out as well, somehow I find it easier to bring things back in this way.

dnadales (approved):
image

.buildkite/pipeline.yml
ruhatch: Yeah, I think it also shows the intention to bring it back. If it's not used in a couple of months, then it can go properly!

PR #374, opened 2019-03-19
Disable cache-s3

ruhatch:
This drops build times because of limitations in the download times of cache-s3


dnadales (approved):
I guess @rvl would be able to provide substantial feedback. Other than that LGTM!

PR #379, opened 2019-03-21
Update references from cardano-base to cardano-binary

ruhatch:


nc6 (approved):
Seems to fix downstream builds.

PR #381, opened 2019-03-21
Update nix files to use lib.nix for pkgs

ruhatch:


scripts/brittany/default.nix
```diff
@@ -1,14 +1,9 @@
...
-}:
-
-with pkgs.lib;
+with import ../../lib.nix;
 with pkgs;
 
 writeScript "check-brittany" ''
   #!${stdenv.shell}
-  export PATH=${lib.makeBinPath [ findutils git haskellPackages.brittany ]}:$PATH
+  export PATH=${lib.makeBinPath [ findutils git haskell.packages.ghc844.brittany ]}:$PATH
```
nc6: Why GHC 8.4.4? Surely we want 8.6.4? Or is it that Brittany is broken on 8.6.4? In which case we would expect it to fail on a bunch of files...

scripts/buildkite/default.nix
```diff
@@ -1,14 +1,7 @@
...
-
-with pkgs.lib;
+with import ../../lib.nix;
 with pkgs;
 
 let
-  cache-s3 = callPackage ./cache-s3.nix {};
 
   stack-hpc-coveralls = pkgs.haskell.lib.dontCheck
     (haskell.packages.ghc844.callPackage ./stack-hpc-coveralls.nix {});
```
nc6: Particular reason for pinning 844 here?

nc6 (approved):
LGTM, a couple of questions re pinning packages in old GHC versions.

PR #383, opened 2019-03-22
 Prepare for integration with ouroboros-network

erikd:
These changes to make integration with `ouroboros-network` easier.

Also want to add a test that asserts that the previous header hash of the zeroth block on the mainnet chain is `Nothing` but will do that in a separate PR.

### DIFF
--- a/src/Cardano/Chain/Block/Header.hs
+++ b/src/Cardano/Chain/Block/Header.hs
@@ -85,7 +85,7 @@ import Cardano.Chain.Block.ExtraBodyData (ExtraBodyData)
 import Cardano.Chain.Block.ExtraHeaderData
   (ExtraHeaderData(..), ExtraHeaderDataError, verifyExtraHeaderData)
 import Cardano.Chain.Block.Proof (Proof(..), mkProof)
-import Cardano.Chain.Common (Attributes, ChainDifficulty)
+import Cardano.Chain.Common (Attributes, ChainDifficulty (..))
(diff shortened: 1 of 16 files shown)


src/Cardano/Chain/Block/Header.hs
```diff
@@ -130,8 +131,19 @@ data AHeader a = AHeader
   , headerAnnotation    :: a
   } deriving (Eq, Show, Generic, NFData, Functor)
 
-headerPrevHash :: AHeader a -> HeaderHash
-headerPrevHash = unAnnotated . aHeaderPrevHash
+-- | Return the 'HeaderHash' of the previous header if it exists.
```
ruhatch: This is actually not quite right. The first block has a prevHash of the header of the first epoch boundary block. That block is the one that has no real previous header, but we deal with them separately anyway.

I think we should revert this, but maybe comment that sometimes those `HeaderHash`es will be from EBBs.

ruhatch (dismissed):
Looks neat in general, but need to look at `headerPrevHash` further

src/Cardano/Chain/Block/Header.hs
erikd: The test I am working on would have validated this.

erikd: I removed this commit for the PR.

ruhatch (approved):
LGTM!

bors r+

PR #389, opened 2019-03-26
Miscellaneous tweaks made while getting up to speed on the repo

mhuesch:
Small changes to a few files.

### DIFF
--- a/src/Cardano/Chain/Common/LovelacePortion.hs
+++ b/src/Cardano/Chain/Common/LovelacePortion.hs
@@ -70,6 +70,9 @@ instance Bi LovelacePortion where
   encode = encode . getLovelacePortion
   decode = LovelacePortion <$> decode
 
+-- The Canonical and Aeson instances for LovelacePortion are inconsistent -
+-- Canonical reads/writes an integer, but Aeson reads/write a Real in range [0,1]
+-- This is because 'canonical-json' only supports numbers of type @Int54@.
 instance Monad m => ToJSON m LovelacePortion where
   toJSON = toJSON . getLovelacePortion
 
--- a/src/Cardano/Chain/Genesis/AvvmBalances.hs
+++ b/src/Cardano/Chain/Genesis/AvvmBalances.hs
@@ -19,7 +19,9 @@ import Cardano.Chain.Common (Lovelace)
 import Cardano.Crypto.Signing.Redeem (RedeemPublicKey)
 
 
--- | Predefined balances of avvm entries.
+-- | Predefined balances of AVVM (Ada Voucher Vending Machine) entries.
+-- People who purchased Ada at a pre-sale were issued a certificate during
+-- the pre-sale period. These certificates allow customers to redeem ADA.
 newtype GenesisAvvmBalances = GenesisAvvmBalances
   { getGenesisAvvmBalances :: Map RedeemPublicKey Lovelace
   } deriving (Show, Eq, Semigroup)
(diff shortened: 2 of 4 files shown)


src/Cardano/Chain/Genesis/Config.hs
```diff
@@ -113,14 +114,13 @@ instance FromJSON StaticConfig where
...
-            kV
-            protocolMagicV
-            initializerV
+        either fail (pure . GCSpec) $ mkGenesisSpec
+          avvmDistrV
+          heavyDelegationV
+          protocolParametersV
+          kV
+          protocolMagicV
+          initializerV
```
mhuesch, line 123: I made this edit on the assumption that we shouldn't ever have invalid datatypes specified in config. It's that we want that, but it seems weird. If that is indeed the case, we should document why here.

intricate: Off-the-top, I can't imagine why we'd want to construct an invalid `GCSpec` so this seems like a good modification to me.

src/Cardano/Chain/Genesis/AvvmBalances.hs
```diff
@@ -19,7 +19,10 @@ import Cardano.Chain.Common (Lovelace)
 import Cardano.Crypto.Signing.Redeem (RedeemPublicKey)
 
 
--- | Predefined balances of avvm entries.
+-- | Predefined balances of AVVM (Ada Voucher Vending Machine) entries.
```
intricate, line 22: Good addition. Will definitely save someone the time of having to look into this.

intricate (approved):

src/Cardano/Chain/Common/LovelacePortion.hs
```diff
@@ -70,6 +70,9 @@ instance Bi LovelacePortion where
   encode = encode . getLovelacePortion
   decode = LovelacePortion <$> decode
 
+-- The Canonical and Aeson instances for LovelacePortion are inconsistent -
```
dnadales, line 73: Nice :)

dnadales (approved):
Thanks for cleaning things a bit, and adding the comments!

src/Cardano/Chain/Common/LovelacePortion.hs
ruhatch: :+1:

src/Cardano/Chain/Genesis/AvvmBalances.hs
```diff
@@ -19,7 +19,10 @@ import Cardano.Chain.Common (Lovelace)
...
+-- People who purchased Ada at a pre-sale were issued a certificate at the
```
ruhatch:
```suggestion
-- People who purchased Ada at a pre-sale were issued a certificate
```

src/Cardano/Chain/Genesis/Config.hs
ruhatch: Seems good :)

ruhatch (approved):
Thanks @mhuesch

Jimbo4350 (approved):
LGTM!

src/Cardano/Chain/Genesis/AvvmBalances.hs
mhuesch: Thanks for the grammar catch. I just dropped the extra words and then rewrapped the text: https://github.com/input-output-hk/cardano-ledger/pull/389/files#diff-633568df21fc9c3cf311dbc4c323a689R24

PR #391, opened 2019-03-27
Update HeapWords instance for data type AbstractHash

intricate:
The value 12 was hardcoded for the word overhead of an `AbstractHash` alloc'd on the heap. However, the implementation of the `Digest` data type in `cryptonite` was modified such that it no longer utilizes a `UArray Word8` but a `Block Word8`, therefore further optimizing memory usage.

### DIFF
--- a/crypto/src/Cardano/Crypto/Hashing.hs
+++ b/crypto/src/Cardano/Crypto/Hashing.hs
@@ -115,7 +115,31 @@ instance (Typeable algo, Typeable a, HashAlgorithm algo) => Bi (AbstractHash alg
     in fromInteger (toInteger (withWordSize realSz + realSz))
 
 instance HeapWords (AbstractHash algo a) where
-  heapWords _ = 12 -- Hardcoded as 8 word overhead of Digest + 4 words for hash
+  heapWords _
+    -- We have
+    --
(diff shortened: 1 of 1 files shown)


dnadales (approved):

crypto/src/Cardano/Crypto/Hashing.hs
```diff
@@ -115,7 +115,31 @@ instance (Typeable algo, Typeable a, HashAlgorithm algo) => Bi (AbstractHash alg
     in fromInteger (toInteger (withWordSize realSz + realSz))
 
 instance HeapWords (AbstractHash algo a) where
-  heapWords _ = 12 -- Hardcoded as 8 word overhead of Digest + 4 words for hash
+  heapWords _
+    -- We have
```
dnadales, line 119: Nice :+1:

ruhatch (approved):
LGTM!

bors r+

crypto/src/Cardano/Crypto/Hashing.hs
mhuesch: Yeah, this is great! Thanks for making it very clear.

mhuesch (approved):

Issue #385, opened 2019-03-22
Implement address discrimination via `NetworkMagic`


intricate:
As it stands, there is no method by which we can discriminate between `Address`es created and utilized on different networks (`mainnet`, `testnet`, `staging`, etc.). As a result, a user could, for example, attempt to send funds from a `mainnet` address to a `testnet` address.

Address discrimination was implemented in `cardano-sl` by introducing a new data type, `NetworkMagic`, and adding a field of this type to the `AddrAttributes` data type. This functionality should be ported over here as well.

intricate:
Relevant PRs from `cardano-sl`:

- https://github.com/input-output-hk/cardano-sl/pull/3712
- https://github.com/input-output-hk/cardano-sl/pull/3715
- https://github.com/input-output-hk/cardano-sl/pull/3733
- https://github.com/input-output-hk/cardano-sl/pull/3756
- https://github.com/input-output-hk/cardano-sl/pull/3775

PR #392, opened 2019-03-27
[#385] Address Discrimination Part 1: altering configuration

mhuesch:
Relates to #385.

I think this is fairly well explained in the two commit messages:
```
    [#385] Revamp ProtocolMagic/Id in Genesis datatypes

    Instead of storing a @ProtocolMagic@ in @Genesis.Data@, and being
    forced to pick a default for the omitted @RequiresNetworkMagic@ field,
    switch to a @ProtocolMagicId@.

    Add a @RequiresNetworkMagic@ field to @Genesis.Config@.

    Leave @GenesisSpec@ the same - use its @ProtocolMagic@ to fill the
    @ProtocolMagicId@ of @Genesis.Data@ and the @RequiresNetworkMagic@
    of @Genesis.Config@.

    This seems to simplify the JSON instances considerably. We will no
    longer be backwards compatible with `cardano-sl`'s configuration,
    but that is ok.
```
and
```
    [#385] Use arbitrary @ProtocolMagic@s in arbitrary instances
```

### DIFF
--- a/crypto/src/Cardano/Crypto/ProtocolMagic.hs
+++ b/crypto/src/Cardano/Crypto/ProtocolMagic.hs
@@ -23,9 +23,6 @@ import Cardano.Prelude
 
 import qualified Data.Aeson as A
 import Data.Aeson ((.:), (.=))
-import Data.Aeson.Types (typeMismatch)
-import Data.List (lookup)
-import Data.String (String)
 import Text.JSON.Canonical (FromJSON(..), JSValue(..), ToJSON(..), expected)
 
 import Cardano.Binary.Class (Bi)
@@ -47,6 +44,12 @@ newtype ProtocolMagicId = ProtocolMagicId
     deriving newtype Bi
     deriving anyclass NFData
 
+instance A.ToJSON ProtocolMagicId where
+  toJSON = A.toJSON . unProtocolMagicId
+
+instance A.FromJSON ProtocolMagicId where
+  parseJSON v = ProtocolMagicId <$> A.parseJSON v
+
 -- mhueschen: For backwards-compatibility reasons, I redefine this function
 -- in terms of the two record accessors.
 getProtocolMagic :: ProtocolMagic -> Int32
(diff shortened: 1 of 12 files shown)


mhuesch:
For context, this is the first of 3 PR's. If this is PR0, then here are the branches for [PR1](https://github.com/input-output-hk/cardano-ledger/tree/mhuesch/385_addr_discrim_implement_nm_1) and [PR2](https://github.com/input-output-hk/cardano-ledger/tree/mhuesch/385_addr_discrim_tests_2). They'll have to rebased over each other, of course.

dnadales:
> If this is PR0, then ...

Happy to see I'm not the only one starting [numbering at zero](https://www.cs.utexas.edu/users/EWD/transcriptions/EWD08xx/EWD831.html) :)

dnadales (commented):
Nice. I'll like the simplifications on the `To/FromJson` instances. However I'll leave the approval to somebody that is more familiar with the code :)

src/Cardano/Chain/Genesis/Config.hs
```diff
@@ -133,6 +133,8 @@ data Config = Config
     -- ^ The hash of the canonical JSON representation of the 'GenesisData'
     , configGeneratedSecrets  :: Maybe GeneratedSecrets
     -- ^ Secrets needed to access 'GenesisData' in testing
+    , configReqNetMagic       :: RequiresNetworkMagic
+    -- ^ Differentiates between Testnet and Mainet/Staging
     --
     --   TODO: Figure out how to split testing and mainnet needs
```
Jimbo4350, line 139: Should we remove this comment and turn it into an issue?

crypto/src/Cardano/Crypto/ProtocolMagic.hs
```diff
@@ -56,79 +59,19 @@ instance A.ToJSON ProtocolMagic where
...
-
-and the new format
-
-```
-protocolMagic:
-  pm: 12345678
-  requiresNetworkMagic: RequiresNoMagic
-```
--}
+instance Monad m => ToJSON m ProtocolMagicId where
```
Jimbo4350, line 69: :+1:

Jimbo4350 (approved):
This is significantly cleaner, good work!

ruhatch (approved):
Neat - great idea to move it up into the `Config` :+1:

intricate (approved):
LGTM

src/Cardano/Chain/Genesis/Config.hs
mhuesch: I am not sure what the comment means, so I left it. If it just refers to what we're doing with adding `NetworkMagic`, then perhaps it is / will soon be done.

Looks like it was added by @ruhatch in b403a0a5 - Ru can you confirm whether this is done or if a separate issue needs to be opened?

PR #393, opened 2019-03-28
[#385] Introduce & implement NetworkMagic

mhuesch:
NetworkMagic allows us to differentiate between Addresses which
come from different networks. This prevents users from accidentally
sending funds from a mainnet wallet to a testnet address, which would
technically be a valid address on mainnet, but would be controlled by
unknown keys not under the user's control (thus rendering the funds
unusable). When addresses are created, the @ProtocolMagicId@ of the
machine is inserted into the @NetworkMagic@, which is added to the
@Address@ datatype.

In this commit I introduce NetworkMagic, wire it up to @ProtocolMagic@
in the configuration, and modify functions so that @NetworkMagic@ can
be passed down to both address creation and transaction validation. I
add a new constructor to @TxValidationError@
- `TxValidationNetworkMagicMismatch` - which is thrown when we
encounter a @NetworkMagic@ coming from an address which doesn't match
our cluster's @NetworkMagic@.

Relates to #385.

### DIFF
--- a/src/Cardano/Chain/Common/AddrAttributes.hs
+++ b/src/Cardano/Chain/Common/AddrAttributes.hs
@@ -1,6 +1,6 @@
+{-# LANGUAGE DeriveAnyClass             #-}
 {-# LANGUAGE DeriveGeneric              #-}
 {-# LANGUAGE FlexibleInstances          #-}
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE OverloadedStrings          #-}
 
 module Cardano.Chain.Common.AddrAttributes
@@ -18,14 +18,21 @@ import Cardano.Binary.Class (Bi, decode, encode)
 import qualified Cardano.Binary.Class as Bi
 import Cardano.Chain.Common.Attributes
   (Attributes(..), decodeAttributes, encodeAttributes)
+import Cardano.Chain.Common.NetworkMagic (NetworkMagic(..))
 import Cardano.Crypto.HD (HDAddressPayload)
+
 
 -- | Additional information stored along with address. It's intended
 -- to be put into 'Attributes' data type to make it extensible with
 -- softfork.
-newtype AddrAttributes = AddrAttributes
-    { aaPkDerivationPath  :: Maybe HDAddressPayload
-    } deriving (Eq, Ord, Show, Generic, NFData, HeapWords)
+data AddrAttributes = AddrAttributes
+    { aaPkDerivationPath  :: !(Maybe HDAddressPayload)
+    , aaNetworkMagic      :: !NetworkMagic
+    } deriving (Eq, Ord, Show, Generic, NFData)
+
+instance HeapWords AddrAttributes where
+  heapWords aa = 3 + heapWords (aaPkDerivationPath aa)
+                   + heapWords (aaNetworkMagic aa)
 
 instance B.Buildable AddrAttributes where
   build aa = bprint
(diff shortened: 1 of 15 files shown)


src/Cardano/Chain/Txp/Validation.hs
```diff
@@ -112,6 +127,12 @@ validateTxIn utxo txIn
   | txIn `UTxO.member` utxo = pure ()
   | otherwise = throwError $ TxValidationMissingInput txIn
 
+-- | Validate the NetworkMagic of a TxOut
+validateTxOutNM :: MonadError TxValidationError m => NetworkMagic -> TxOut -> m ()
```
Jimbo4350, line 131: :100:

Jimbo4350 (approved):
LGTM!

src/Cardano/Chain/Common/NetworkMagic.hs
```diff
@@ -0,0 +1,43 @@
...
+import           Cardano.Crypto.ProtocolMagic (ProtocolMagic (..),
+                     RequiresNetworkMagic (..), getProtocolMagic)
+
+
+--------------------------------------------------------------------------------
+-- NetworkMagic
+--------------------------------------------------------------------------------
+
+-- Although it doesn't make sense for an identifier to have a sign, we're opting
+-- to maintain consistency with `ProtocolMagicId`, rather than risk subtle
```
intricate, line 24: Since we didn't want to make this change in `cardano-sl`, maybe we could do it in this repo (in another PR at some point)? There really is no reason for an identifier like `ProtocolMagicId` to have a sign, imo.

src/Cardano/Chain/Common/AddrAttributes.hs
```diff
@@ -18,14 +17,23 @@ import Cardano.Binary.Class (Bi, decode, encode)
...
 -- softfork.
-newtype AddrAttributes = AddrAttributes
-    { aaPkDerivationPath  :: Maybe HDAddressPayload
-    } deriving (Eq, Ord, Show, Generic, NFData, HeapWords)
+data AddrAttributes = AddrAttributes
+    { aaPkDerivationPath  :: !(Maybe HDAddressPayload)
+    , aaNetworkMagic      :: !NetworkMagic
+    } deriving (Eq, Ord, Show, Generic)
+
+instance NFData AddrAttributes
```
ruhatch: You should be able to derive this via `anyclass` :+1:

src/Cardano/Chain/Common/AddrAttributes.hs
```diff
@@ -18,14 +17,23 @@ import Cardano.Binary.Class (Bi, decode, encode)
...
+
+instance HeapWords AddrAttributes where
+  heapWords aa = 3 + heapWords (aaPkDerivationPath aa)
```
ruhatch, line 34: I think you should save some words here from unpacking?

ruhatch: Oh no, they're both multi-constructor, so I think you're right

src/Cardano/Chain/Common/NetworkMagic.hs
```diff
@@ -0,0 +1,43 @@
...
+-- conversion bugs.
+data NetworkMagic
+    = NetworkMainOrStage
+    | NetworkTestnet {-# UNPACK #-} !Int32
+    deriving (Show, Eq, Ord, Generic)
+
+instance NFData NetworkMagic
```
ruhatch: Should also be able to derive this via anyclass

src/Cardano/Chain/Txp/Validation.hs
```diff
@@ -112,6 +127,12 @@ validateTxIn utxo txIn
...
+validateTxOutNM nm txOut =
+  let addrNm = addrNetworkMagic . txOutAddress $ txOut
+   in unless (nm == addrNm) $
```
ruhatch: We use a function `orThrowError` for this `unless pred $ throwError` pattern, so:
```
(nm == addrNm)
  `orThrowError` TxValidationNetworkMagicMismatch nm addrNm
```

ruhatch (approved):
LGTM, just some small comments :)

src/Cardano/Chain/Common/AddrAttributes.hs
mhuesch: Sweet, just incorporated, thanks for the heads up.

src/Cardano/Chain/Txp/Validation.hs
mhuesch: Clean! Just incorporated.

intricate (approved):
LGTM

src/Cardano/Chain/Common/NetworkMagic.hs
mhuesch: Opened issue: https://github.com/input-output-hk/cardano-ledger/issues/396

Issue #209, opened 2018-12-06
Investigate epochSlots size


ruhatch:
The `LocalSlotIndex` type, which represents an index within an epoch is a `Word16`, but the `epochSlots` constant is a `SlotCount`, which is a `Word64`. In some tests I ran into a problem with generating fake `epochSlots` values using `genSlotCount`. If `epochSlots` should be a `Word16` then we should encode that in the types somehow.

ruhatch:
In Shelley `EpochSlots` will indeed take on a value that is greater than `maxBound :: Word16`, so we should leave it as it is. We should keep track of the fact that it is constrained by `LocalSlotIndex` in the Byron era, as this has an impact on test value generation.

Issue #170, opened 2018-11-27
Give `SlotId` a more specific name, and `FlatSlotId` a better name.


dnadales:
A slot-id could represent the absolute index of a slot across epochs, or the index of the slot within an epoch. To remove ambiguity regarding what `SlotId` means we should give it a more specific name. Furthermore, `FlatSlotId` should be given a better name. The formal specifications should be updated to reflect the new naming conventions.

## Proposed Solution

First we need to rename `SlotId` to something more appropriate, since it is now defined as:

```haskell
data SlotId = SlotId
  { siEpoch :: !EpochIndex
  , siSlot  :: !LocalSlotIndex
  } deriving (Show, Eq, Ord, Generic)
```

This data structure should be placed in a different module from `Cardano.Chain.Slotting.SlotId`, to reflect the fact that it is a legacy data structure, and we shouldn't use it in any new code.

Instead of `LocalSlotIndex` we can use something like `SlotCount`, which clearly reflects what this type denotes.

`FlatSlotId` should be renamed to `SlotNo` to be consistent with [`ouroboros-network`](https://github.com/input-output-hk/ouroboros-network/pull/374).

Then we need to make sure the `SlotId` is only used at the block header (since we cannot change the block structure that is already persisted), and use `SlotNo` everywhere else. This means that when we decode a block we need to translate the epoch-and-slot pair (represented by the `SlotId` data structure above), to a `SlotNo`, which in turn will require that we use a fixed number of slots-per-epoch.

By hard-coding the number of slots per epoch we won't be able to use a smaller number in the tests, so we could add a new protocol parameter, with a default value.

Additionally, for the sake of consistency, `EpochIndex` should be renamed to `EpochNo`.

## Next Actions

- [ ] Identify all the places in which a translation from `SlotId` to `FlatSlotId` is needed, so that the use of `SlotId` can be restricted to the block serialization and de-serialization. Note that we use old names here since renaming will/should take place at a later stage.
- [ ] Bump the version of `cardano-ledger-specs`.
- [ ] Rename `SlotId` to `EpochAndSlotCount`, and place it in a module that reflects the fact that it should be used in new code. Use `SlotCount` instead of `LocalSlotIndex`.
- [ ] Rename `EpochIndex` to `EpochNo`.
- [ ] Rename `FlatSlotId` to `SlotNo`.
[...]

PR #394, opened 2019-03-28
Make slot numbers absolute

### LINKED ISSUE #169: Don't panic on something that won't happen, but check division by `0`
In `src/Cardano/Chain/Slotting/SlotId.hs`, we have this code:

```haskell

-- | Construct a 'SlotId' from a flattened variant, using a given 'SlotCount'
--   modulus
unflattenSlotId :: SlotCount -> FlatSlotId -> SlotId
unflattenSlotId es n = SlotId {siEpoch = fromIntegral epoch, siSlot = lsi}
 where
  (epoch, slot) = n `divMod` fromIntegral es
  lsi           = case mkLocalSlotIndex es (fromIntegral slot) of
    Left err -> panic
      $ sformat ("The impossible happened in unflattenSlotId: " . build) err
    Right lsi' -> lsi'

mkLocalSlotIndex_ :: SlotCount -> Word16 -> Maybe LocalSlotIndex
mkLocalSlotIndex_ es idx
  | idx < fromIntegral es = Just (UnsafeLocalSlotIndex idx)
  | otherwise             = Nothing

```

As far as I understand, the `panic` branch won't ever be called since `slot` is the remainder of the division by `es`. On the other hand, we should check division by zero (or use a type wrapper for non-zero naturals). The `mkLocalSlotIndex` doesn't seem to be needed anywhere else.

dnadales:
This is part of #170. In particular this addresses the following sub-task:

> Identify all the places in which a translation from `SlotId` to `FlatSlotId` is needed, so that the use of `SlotId` can be restricted to the block serialization and de-serialization. Note that we use old names here since renaming will/should take place at a later stage.

At first at thought about creating a draft PR, but since I had to rebase twice today, I'd like to speed up the integration of this to master, provided that what I did makes sense.

Closes #169 .

### DIFF
--- a/src/Cardano/Chain/Block/Block.hs
+++ b/src/Cardano/Chain/Block/Block.hs
@@ -49,6 +49,7 @@ where
 import Cardano.Prelude
 
 import qualified Data.ByteString as BS
+import Data.Text.Lazy.Builder (Builder)
 import Formatting (bprint, build, int, shown)
 import qualified Formatting.Buildable as B
 
@@ -86,6 +87,7 @@ import Cardano.Chain.Block.Header
   , ToSign
   , decodeAHeader
   , dropBoundaryHeader
+  , encodeHeader'
   , genesisHeaderHash
   , hashHeader
   , headerAttributes
@@ -106,7 +108,11 @@ import Cardano.Chain.Block.Proof (Proof(..))
 import Cardano.Chain.Common (Attributes, ChainDifficulty (..), mkAttributes)
 import qualified Cardano.Chain.Delegation as Delegation
 import Cardano.Chain.Genesis.Hash (GenesisHash(..))
-import Cardano.Chain.Slotting (SlotId(..))
+import Cardano.Chain.Slotting
+  ( EpochSlots
+  , FlatSlotId
+  , WithEpochSlots(WithEpochSlots)
+  )
 import Cardano.Chain.Ssc (SscPayload)
 import Cardano.Chain.Txp.TxPayload (ATxPayload)
 import Cardano.Chain.Update.ProtocolVersion (ProtocolVersion)
@@ -130,8 +136,13 @@ data ABlock a = ABlock
   , blockAnnotation :: a
   } deriving (Eq, Show, Generic, NFData, Functor)
 
-instance B.Buildable Block where
-  build block = bprint
+instance B.Buildable (WithEpochSlots Block) where
+  build (WithEpochSlots es block) = renderBlock es block
+
+
+renderBlock :: EpochSlots -> Block -> Builder
+renderBlock es block =
+  bprint
     ( "Block:\n"
     . "  " . build . "  transactions (" . int . " items): " . listJson . "\n"
     . "  " . build . "\n"
@@ -139,37 +150,41 @@ instance B.Buildable Block where
     . "  update payload: " . build . "\n"
     . "  " . build
     )
-    (blockHeader block)
+    (WithEpochSlots es $ blockHeader block)
     (length txs)
     txs
     (blockDlgPayload block)
     (blockSscPayload block)
     (blockUpdatePayload block)
     (blockExtraData block)
-    where txs = bodyTxs $ blockBody block
+  where txs = bodyTxs $ blockBody block
 
-instance Bi Block where
-  encode block =
-    encodeListLen 3
-      <> encode (blockHeader block)
-      <> encode (blockBody block)
-      <> encode (blockExtraData block)
 
-  decode = void <$> decodeABlock
+-- | Encode a block, given a number of slots-per-epoch.
+--
+-- Unlike 'encodeBlock', this function does not take the deprecated epoch
+-- boundary blocks into account.
+--
+encodeBlockWithoutBoundary :: EpochSlots -> Block -> Encoding
+encodeBlockWithoutBoundary epochSlots block
+  =  encodeListLen 3
+  <> encodeHeader' epochSlots (blockHeader block)
+  <> encode (blockBody block)
+  <> encode (blockExtraData block)
 
-decodeABlock :: Decoder s (ABlock ByteSpan)
-decodeABlock = do
+decodeABlock :: EpochSlots -> Decoder s (ABlock ByteSpan)
+decodeABlock epochSlots = do
   Annotated (header, body, ed) byteSpan <-
     annotatedDecoder $ do
       enforceSize "Block" 3
-      (,,) <$> decodeAHeader <*> decodeABody <*> decodeAnnotated
+      (,,) <$> decodeAHeader epochSlots <*> decodeABody <*> decodeAnnotated
   pure $ ABlock header body ed byteSpan
 
 
 -- | Encode a 'Block' accounting for deprecated epoch boundary blocks
-encodeBlock :: Block -> Encoding
-encodeBlock block = encodeListLen 2 <> encode (1 :: Word) <> encode block
-
+encodeBlock :: EpochSlots -> Block -> Encoding
+encodeBlock epochSlots block =
+  encodeListLen 2 <> encode (1 :: Word) <> encodeBlockWithoutBoundary epochSlots block
 
 data ABlockOrBoundary a
   = ABOBBlock (ABlock a)
@@ -183,18 +198,19 @@ data ABlockOrBoundary a
 --   now deprecated these explicit boundary blocks, but we still need to decode
 --   blocks in the old format. In the case that we find a boundary block, we
 --   drop it using 'dropBoundaryBlock' and return a 'Nothing'.
-decodeABlockOrBoundary :: Decoder s (ABlockOrBoundary ByteSpan)
-decodeABlockOrBoundary = do
+decodeABlockOrBoundary :: EpochSlots -> Decoder s (ABlockOrBoundary ByteSpan)
+decodeABlockOrBoundary epochSlots = do
   enforceSize "Block" 2
   decode @Word >>= \case
     0 -> ABOBBoundary <$> dropBoundaryBlock
-    1 -> ABOBBlock <$> decodeABlock
+    1 -> ABOBBlock <$> decodeABlock epochSlots
     t -> cborError $ DecoderErrorUnknownTag "Block" (fromIntegral t)
 
-decodeBlockOrBoundary :: Decoder s (Maybe Block)
-decodeBlockOrBoundary = decodeABlockOrBoundary >>= \case
-  ABOBBoundary _ -> pure Nothing
-  ABOBBlock    b -> pure . Just $ void b
+decodeBlockOrBoundary :: EpochSlots -> Decoder s (Maybe Block)
+decodeBlockOrBoundary epochSlots =
+  decodeABlockOrBoundary epochSlots >>= \case
+    ABOBBoundary _ -> pure Nothing
+    ABOBBlock    b -> pure . Just $ void b
 
 --------------------------------------------------------------------------------
 -- BoundaryBlock
@@ -236,7 +252,8 @@ mkBlock
   -> ProtocolVersion
   -> SoftwareVersion
   -> Either GenesisHash Header
-  -> SlotId
+  -> EpochSlots
+  -> FlatSlotId
   -> SecretKey
   -- ^ The 'SecretKey' used for signing the block
   -> Maybe Delegation.Certificate
@@ -244,9 +261,10 @@ mkBlock
   --   right to sign this block
   -> Body
   -> Block
-mkBlock pm bv sv prevHeader = mkBlockExplicit pm bv sv prevHash difficulty
+mkBlock pm bv sv prevHeader epochSlots =
+  mkBlockExplicit pm bv sv prevHash difficulty epochSlots
  where
-  prevHash   = either genesisHeaderHash hashHeader prevHeader
+  prevHash   = either genesisHeaderHash (hashHeader epochSlots) prevHeader
   difficulty = either (const $ ChainDifficulty 0) (succ . headerDifficulty) prevHeader
 
 -- | Smart constructor for 'Block', without requiring the entire previous
@@ -259,7 +277,8 @@ mkBlockExplicit
   -> SoftwareVersion
   -> HeaderHash
   -> ChainDifficulty
-  -> SlotId
+  -> EpochSlots
+  -> FlatSlotId
   -> SecretKey
   -- ^ The 'SecretKey' used for signing the block
   -> Maybe Delegation.Certificate
@@ -267,11 +286,12 @@ mkBlockExplicit
   --   right to sign this block
   -> Body
   -> Block
-mkBlockExplicit pm bv sv prevHash difficulty slotId sk mDlgCert body = ABlock
-  (mkHeaderExplicit pm prevHash difficulty slotId sk mDlgCert body extraH)
-  body
-  (Annotated extraB ())
-  ()
+mkBlockExplicit pm bv sv prevHash difficulty epochSlots slotId sk mDlgCert body =
+  ABlock
+    (mkHeaderExplicit pm prevHash difficulty epochSlots slotId sk mDlgCert body extraH)
+    body
+    (Annotated extraB ())
+    ()
  where
   extraB :: ExtraBodyData
   extraB = ExtraBodyData (mkAttributes ())
@@ -282,8 +302,8 @@ mkBlockExplicit pm bv sv prevHash difficulty slotId sk mDlgCert body = ABlock
 -- Block accessors
 --------------------------------------------------------------------------------
 
-blockHash :: Block -> HeaderHash
-blockHash = hashHeader . blockHeader
+blockHash :: EpochSlots -> Block -> HeaderHash
+blockHash epochSlots = hashHeader epochSlots . blockHeader
 
 blockHashAnnotated :: ABlock ByteString -> HeaderHash
 blockHashAnnotated = hashDecoded . fmap wrapHeaderBytes . blockHeader
@@ -297,7 +317,7 @@ blockPrevHash = headerPrevHash . blockHeader
 blockProof :: ABlock a -> Proof
 blockProof = headerProof . blockHeader
 
-blockSlot :: ABlock a -> SlotId
+blockSlot :: ABlock a -> FlatSlotId
 blockSlot = headerSlot . blockHeader
 
 blockLeaderKey :: ABlock a -> PublicKey
@@ -306,8 +326,8 @@ blockLeaderKey = headerLeaderKey . blockHeader
 blockDifficulty :: ABlock a -> ChainDifficulty
 blockDifficulty = headerDifficulty . blockHeader
 
-blockToSign :: ABlock a -> ToSign
-blockToSign = headerToSign . blockHeader
+blockToSign :: EpochSlots -> ABlock a -> ToSign
+blockToSign epochSlots = headerToSign epochSlots . blockHeader
 
 blockSignature :: ABlock a -> BlockSignature
 blockSignature = headerSignature . blockHeader
--- a/src/Cardano/Chain/Block/Header.hs
+++ b/src/Cardano/Chain/Block/Header.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE FlexibleInstances    #-}
 {-# LANGUAGE LambdaCase           #-}
 {-# LANGUAGE OverloadedStrings    #-}
+{-# LANGUAGE ScopedTypeVariables  #-}
 {-# LANGUAGE TypeApplications     #-}
 {-# LANGUAGE TypeFamilies         #-}
 {-# LANGUAGE TypeSynonymInstances #-}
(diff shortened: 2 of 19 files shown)


src/Cardano/Chain/Block/Block.hs
```diff
@@ -130,46 +136,60 @@ data ABlock a = ABlock
...
     . "  " . build . "\n"
     . "  " . shown . "\n"
     . "  update payload: " . build . "\n"
     . "  " . build
     )
-    (blockHeader block)
+    (WithEpochSlots es $ blockHeader block) -- TODO: ask the more experienced
+                                            -- devs around: `build` is heavily
+                                            -- overloaded. Are there better
+                                            -- ways to do this?
```
ruhatch: This is a perfect example of why @erikd is right about `Buildable` in general. We should really replace these calls to build with explicit `renderX` functions for the sub-components. It would be much easier to read and understand, as well as more robust to code changes. If I accidentally change one of the arguments, I want there to be a type error!

src/Cardano/Chain/Block/Block.hs
```diff
@@ -130,46 +136,60 @@ data ABlock a = ABlock
...
-decodeABlock :: Decoder s (ABlock ByteSpan)
-decodeABlock = do
+-- | Encode a block, given a number of slots-per-epoch.
+--
+-- Unlike 'encodeBlock', this function does not take the deprecated epoch
+-- boundary blocks into account.
+--
+-- TODO: do we want to rename any of these functions ('encodeBlock'' or
+-- 'encodeBlock')? If so, what are good names?
+encodeBlock' :: EpochSlots -> Block -> Encoding
```
ruhatch: This should be `encodeBlock` and you could then rename the one that takes boundaries into account `encodeBlockWithBoundary`? My concern is that if someone has a block they want to encode, and they find `encodeBlock`, they'll assume that's the right one, even though we need to take boundaries into account.

In the Shelley era we'd probably want `encodeBlock` to be what it is now and be able to dispatch to the new version. So actually, I think `encodeBlock'` is good!

src/Cardano/Chain/Slotting/EpochSlots.hs
```diff
@@ -26,3 +27,12 @@ instance Bi EpochSlots where
...
 
 deriving instance ToJSON EpochSlots
+
+-- | Data with an accompanying slots per epoch context.
+data WithEpochSlots a = WithEpochSlots
+  { epochSlots       :: EpochSlots
+  , unWithEpochSlots :: a
+  }
+
+deriving instance Show a => Show (WithEpochSlots a)
```
ruhatch: You can derive these on the type directly

ruhatch (commented):
Need to do another round through and address some more stuff, but don't have time today

test/Test/Cardano/Chain/Block/Bi.hs
```diff
@@ -90,46 +114,84 @@ import Test.Cardano.Crypto.Gen (feedPM)
...
+  withTests 1 . property $ do
+    bs <- liftIO $ BS.readFile path
+    let target = decodeBase16 bs
+    compareHexDump bs bs'
+    let
+      fullDecoder :: BS.ByteString -> Either DecoderError a
+      fullDecoder = decodeFullDecoder "" dec -- TODO: do we want to pass the label as parameter?
+    fmap fullDecoder target === Just (Right x)
+
+-- TODO: discuss with Ru whether we want to keep this, as it seems to be the
```
ruhatch: Yeah, the separate ones were for `encodeBlock` and the `encode` Bi instance of Block. Now you'd want one for `encodeBlock` and one for `encodeBlock'`, but maybe that's overkill :man_shrugging:

test/Test/Cardano/Chain/Block/Gen.hs
```diff
@@ -76,14 +84,21 @@ genBody pm =
     <*> Delegation.genPayload pm
     <*> Update.genPayload pm
 
+genHeaderWithEpochSlots :: ProtocolMagicId -> Gen (WithEpochSlots Header)
```
ruhatch: Do you need these if you have the `genWithEpochSlots` wrapper?

test/Test/Cardano/Chain/Block/Validation.hs
```diff
@@ -109,7 +109,8 @@ epochValid
   :: Genesis.Config -> IORef ChainValidationState -> FilePath -> Property
 epochValid config cvsRef fp = withTests 1 . property $ do
   cvs <- liftIO $ readIORef cvsRef
-  let stream = parseEpochFileWithBoundary fp
+  -- TODO: put the hardcoded value in some central place.
+  let stream = parseEpochFileWithBoundary mainnetEpochSlots fp
```
ruhatch: Why aren't you taking this from the config?

test/Test/Cardano/Chain/Epoch/File.hs
```diff
@@ -24,7 +25,11 @@ propDeserializeEpochs :: Property
 propDeserializeEpochs = H.withTests 1 $ H.property $ do
   files <- take 10 <$> liftIO mainnetEpochFiles
   H.assert $ not (null files)
-  let stream = parseEpochFiles files
+  -- TODO: the property cannot take any parameters (if we want discoverPrefix
+  -- to work). Now the question is whether it is OK to use an hardcoded value
+  -- for the number of slots per epoch, and if so in which module should we
+  -- store this constant?
```
ruhatch, line 31: Maybe we should have the full mainnet config in Haskell-land. Then we could also compare the decoded `mainnet-genesis.json` to the Haskell version. I think `Test.Cardano.Chain.Config` is a reasonable location though.

test/Test/Cardano/Chain/Slotting/Example.hs
```diff
@@ -29,9 +32,15 @@ exampleEpochSlottingData :: EpochSlottingData
...
   EpochSlottingData {esdSlotDuration = 100e-6, esdStartDiff = 100e-6}
 
-exampleSlotId :: SlotId
-exampleSlotId = SlotId (EpochIndex 11) lsi
-  where Right lsi = mkLocalSlotIndex (EpochSlots 50) 47
+exampleFlatSlotId :: EpochSlots -> FlatSlotId
+exampleFlatSlotId es =
+  flattenSlotId es (exampleSlotId es)
+
+-- TODO: es was (EpochSlots 50)
```
ruhatch: Why do you need to add parameters here, these are just examples that don't really need to be changed ever!

test/Test/Cardano/Chain/Slotting/Gen.hs
```diff
@@ -76,10 +78,28 @@ genLocalSlotIndex epochSlots = mkLocalSlotIndex'
...
+-- slot-count, which represents a local index of a slot within the epoch and is
+-- represented using a 'Word16', is calculated taking the reminder of dividing
+-- the slot number by the number of slots-per-epoch ('EpochSlots'). So if the
+-- generated epoch would be greater than @2^16@ we couldn't guarantee that the
+-- local-index would fit inside its representation.
+--
 genEpochSlots :: Gen EpochSlots
-genEpochSlots = EpochSlots <$> Gen.word64 Range.constantBounded
+genEpochSlots =
+  EpochSlots . succ . fromIntegral <$> Gen.word16 Range.constantBounded
```
ruhatch: Won't the upper bound be one too high?

test/Test/Cardano/Chain/Slotting/Gen.hs
```diff
@@ -144,5 +164,5 @@ genSlottingDataInvalidIndicies = unsafeSlottingData <$> genSlottingDataMap
 feedPMEpochSlots :: (ProtocolMagicId -> EpochSlots -> Gen a) -> Gen a
 feedPMEpochSlots genA = do
   pm         <- genProtocolMagicId
-  epochSlots <- EpochSlots . fromIntegral <$> Gen.word16 Range.constantBounded
+  epochSlots <- EpochSlots . fromIntegral <$> Gen.word16 Range.constantBounded -- TODO: why not genEpochSlots???
```
ruhatch: `genEpochSlots` should be fine here :+1:

ruhatch (approved):
Just a few small tweaks, but looks good, thanks and sorry it was painful!

src/Cardano/Chain/Block/Block.hs
```diff
@@ -259,19 +282,21 @@ mkBlockExplicit
...
-  (mkHeaderExplicit pm prevHash difficulty slotId sk mDlgCert body extraH)
-  body
-  (Annotated extraB ())
-  ()
+mkBlockExplicit pm bv sv prevHash difficulty epochSlots slotId sk mDlgCert body =
+  ABlock
+    (mkHeaderExplicit pm prevHash difficulty epochSlots slotId sk mDlgCert body extraH)
+    body
+    (Annotated extraB ())
+    ()
```
mhuesch, line 294: 👍

src/Cardano/Chain/Block/Header.hs
```diff
@@ -205,9 +235,10 @@ mkHeader
   -> Body
   -> ExtraHeaderData
   -> Header
-mkHeader pm prevHeader = mkHeaderExplicit pm prevHash difficulty
+mkHeader pm prevHeader epochSlots =
```
mhuesch, line 238: This looks good. One thing that confuses me, tho: how can `mkHeader` not be referenced anywhere in the codebase?

```
mhueschen-MBP :: Projects/IOHK-work/cardano-ledger ‹dnadales/170-make-slot-numbers-absolute*› » ag -w mkHeader
src/Cardano/Chain/Block/Header.hs
19:  , mkHeader
221:mkHeader
238:mkHeader pm prevHeader epochSlots =
259:  -- ^ See 'mkHeader'.
```
Perhaps I am missing something.

src/Cardano/Chain/Block/Header.hs
```diff
@@ -233,28 +266,30 @@ mkHeaderExplicit
...
     Just dlgCert ->
       BlockPSignatureHeavy $ proxySign pm SignMainBlockHeavy sk dlgCert toSign
 
-  leaderPk  = maybe (toPublic sk) pskIssuerPk mDlgCert
+  leaderPk = maybe (toPublic sk) pskIssuerPk mDlgCert
 
   consensus = consensusData slotId leaderPk difficulty signature
 
-headerSlot :: AHeader a -> SlotId
+headerSlot :: AHeader a -> FlatSlotId
```
mhuesch, line 292: 👍

test/Test/Cardano/Chain/Txp/Validation.hs
```diff
@@ -78,15 +78,16 @@ tests scenario = do
...
   deriving (Eq, Show)
 
 
 -- | Check that a single epoch's transactions are valid by folding over 'Blund's
 epochValid :: ProtocolMagicId -> IORef UTxO -> FilePath -> Property
 epochValid pm utxoRef fp = withTests 1 . property $ do
   utxo <- liftIO $ readIORef utxoRef
-  let stream = parseEpochFile fp
+  -- TODO: put this hardcoded module in some central file.
+  let stream = parseEpochFile mainnetEpochSlots fp
```
mhuesch: `mainnetEpochSlots` should probably be replaced by an argument to `epochValid` which can come out of the `GenesisData`.

Now that my Address Discrimination PR has been merged to `master`, it's actually a `Genesis.Config`: https://github.com/input-output-hk/cardano-ledger/blob/06e0f69f5e061d2b544f175411cfe87446af4652/test/Test/Cardano/Chain/Txp/Validation.hs#L53. But the same `gdK` is accessible from inside the enclosed `GenesisData`.

test/Test/Cardano/Chain/Block/Validation.hs
mhuesch: Agree with @ruhatch, I don't see why we are hardcoding `k` [here](https://github.com/input-output-hk/cardano-ledger/pull/394/files#diff-6fa26d8b6c8d61b4f8ac49754e7508b5R33).

What I suggest in my comment [here](https://github.com/input-output-hk/cardano-ledger/pull/394/files#diff-1d9bf76305b28311e265c6151f3adc5cR90) should be applicable here too.

test/Test/Cardano/Chain/Epoch/File.hs
mhuesch: Yeah, interesting idea. Since `test/mainnet-genesis.json` is frozen (in genesis block), it shouldn't change. So we could just have the full config in Haskell. We could even add a property test to check that they're in sync (is that what you're suggesting @ruhatch?)

test/Test/Cardano/Chain/Slotting/Gen.hs
mhuesch: I think that Damian's code is correct here, because there's a type mismatch between `EpochSlots` (Word64) and `LocalSlotIndex` (Word16). So he's generate the full range of the (ordinal) index, then adding 1, which should get us the corresponding cardinals.

test/Test/Cardano/Chain/Slotting/Gen.hs
mhuesch: Agree.

test/Test/Cardano/Chain/Txp/Validation.hs
mhuesch: If we go the route Ru is suggesting [here](https://github.com/input-output-hk/cardano-ledger/pull/394/files#r270530544) then ignore my comment ⬆️ !

test/Test/Cardano/Chain/Block/Validation.hs
mhuesch: If we go the route Ru is suggesting [here](https://github.com/input-output-hk/cardano-ledger/pull/394/files#r270530544) then ignore my comment ⬆️ !

src/Cardano/Chain/Block/Block.hs
mhuesch: I agree, I find this confusing and slightly dangerous (it is "just" printout code, though) because it's so polymorphic.

mhuesch (commented):
LGTM, I think we might want to change a few things before merging though. Or we could do it in a follow-on PR?

mhuesch:
Separate point - if we're "containing" `LocalSlotIndex`, and using `SlotNo` in the actual Haskell datatype, does that mean we'll close #209 and keep `EpochSlots` and `LocalSlotIndex` as mismatched types (Word64 and Word16, respectively)?

test/Test/Cardano/Chain/Epoch/File.hs
ruhatch: @mhuesch yeah, that's exactly what I was suggesting :ok_hand:

src/Cardano/Chain/Block/Block.hs
dnadales: Yup. Me too :) What do we want to do in the meantime? Do we leave the `Buildable` instance of `WithEpochSlots` for the time being? I don't know which other alternatives we have...

src/Cardano/Chain/Block/Block.hs
dnadales: Alright :)

src/Cardano/Chain/Slotting/EpochSlots.hs
dnadales: :man_facepalming: Thanks Ru!

dnadales:
> Just a few small tweaks, but looks good, thanks and sorry it was painful!

Actually it was quite instructive to me :) Thank you @ruhatch and @mhuesch for your help and feedback!

test/Test/Cardano/Chain/Block/Bi.hs
dnadales: Do I kill the test then?

test/Test/Cardano/Chain/Block/Gen.hs
dnadales: Absolutely not! Thanks for catching this Ru!

test/Test/Cardano/Chain/Block/Bi.hs
dnadales: I mean, since we're indirectly testing the `f'` versions though their `f` counterparts...

test/Test/Cardano/Chain/Block/Validation.hs
dnadales: Links overload :dizzy_face: :) 

> Why aren't you taking this from the config?

 Because I was doing things fast and being stupid. Thanks @ruhatch !

@mhuesch : I'll comment on the other remarks [here](https://www.youtube.com/watch?v=oHg5SJYRHA0)

test/Test/Cardano/Chain/Epoch/File.hs
dnadales: This seems like a great idea! 

My suggestion would be, use `Test.Cardano.Chain.Config.mainnetEpochSlots` for now, and create an issue to add a full mainet config as a Haskell data structure, plus equivalence tests w.r.t. the JSON config. Does this make sense?

test/Test/Cardano/Chain/Slotting/Example.hs
dnadales: With the addition of the epoch slots to the encode and decode function, we need to pass the former as a parameter to the latter. Now I could leave `50` hard-coded here, but then we'll have another `50` hard-coded in `Test.Cardano.Chain.Block.Bi`. This was my attempt to make the dependency of the golden test in that module on the magic number `50` explicit.

Should I revert back and have `50` in two places?

test/Test/Cardano/Chain/Slotting/Gen.hs
dnadales: > Won't the upper bound be one too high?

Yes! @erikd made the same observation [here](https://github.com/input-output-hk/cardano-ledger/commit/f5ffae72590f8698e27190116d29676e3fce2473#r32983105).

test/Test/Cardano/Chain/Slotting/Gen.hs
dnadales: Done. Thanks guys!

dnadales:
> LGTM, I think we might want to change a few things before merging though. Or we could do it in a follow-on PR?

Since this is my first big PR in this code-base there are some important review comments that I should address before merging. 

Thanks @mhuesch once more for your time!

src/Cardano/Chain/Block/Header.hs
dnadales: Hmmm, that's strange :thinking: I don't know whether we plan on using this later on (@ruhatch @erikd ?), otherwise we can remove it (maybe in a separate nano-pr, so that we can easily search for the commit that remove `mkHeader` and bring it back if needed).

test/Test/Cardano/Chain/Txp/Validation.hs
dnadales: I think having the mainnet config hardcoded as a Haskell structure makes sense. However in this particular case, we read the config in the tests, so what you're suggesting @mhuesch makes perfect sense!

dnadales: And it'd _make sense_ to expand my vocabulary as well :P

dnadales:
> Separate point - if we're "containing" LocalSlotIndex, and using SlotNo in the actual Haskell datatype, does that mean we'll close #209 and keep EpochSlots and LocalSlotIndex as mismatched types (Word64 and Word16, respectively)?

We still need the `EpochSlots` and `LocalSlotIndex` for converting a `SlotNo` into a `(epoch, slot-offset)` pair which is required by filling up the `ToSign` data structure. So I guess the issue is still valid. Now the question is whether we want to change the type of `LocalSlotIndex` to be a `Word64`, but I guess that'd affect the encoding/decoding...

src/Cardano/Chain/Block/Block.hs
dnadales: Renamed `encodeBlock'` to `encodeBlockWithoutBoundary`. I think that makes things clearer, while keeping `encodeBlock` :) Let me know if this is not right and I'll revert it back.

src/Cardano/Chain/Block/Block.hs
mhuesch: Whoops, I guess either I didn't get a notification for this or I lost track of it.

I think you were right to leave it as `Buildable` for now.

@ruhatch @dnadales I've created [an issue](https://github.com/input-output-hk/cardano-ledger/issues/427) to track this.

test/Test/Cardano/Chain/Epoch/File.hs
mhuesch: This was moved in Jordan's recently PR to `src/`: https://github.com/input-output-hk/cardano-ledger/pull/358/files#diff-b32539f2b1922895977cf767a883874eR76

I created an issue to capture this: https://github.com/input-output-hk/cardano-ledger/issues/432

dnadales: Great! Thanks for keeping an eye on this!

PR #397, opened 2019-03-29
[#385] Add tests to cover new NetworkMagic cases

### LINKED ISSUE #385: Implement address discrimination via `NetworkMagic`

mhuesch:
Add golden tests to cover the new cases introduced by @NetworkMagic@'s
two constructors.

Also add a property tests to demonstrate Addresses correctly hold the
NetworkMagic which was provided to `makeAddress`.

Closes #385 (final PR of 3).


test/Test/Cardano/Chain/Txp/Example.hs
```diff
@@ -75,26 +77,46 @@ exampleTxAux :: TxAux
...
 exampleTxId = exampleHashTx
 
 exampleTxInList :: (NonEmpty TxIn)
 exampleTxInList = fromList [exampleTxInUtxo]
 
+exampleTxInList1 :: (NonEmpty TxIn)
+exampleTxInList1 = fromList [exampleTxInUtxo, exampleTxInUtxo1]
+
 exampleTxInUtxo :: TxIn
 exampleTxInUtxo = TxInUtxo exampleHashTx 47 -- TODO: loop here
```
intricate, line 94: This is unrelated to your PR but any idea what is meant by this `TODO` comment?

test/Test/Cardano/Chain/Common/Address.hs
```diff
@@ -0,0 +1,31 @@
+{-# LANGUAGE DataKinds         #-}
```
intricate, line 1: Were the `Address`-related golden and round-trip tests not ported over to this codebase? We should probably write up an issue for this.

intricate (approved):
The content of this PR looks good but I've got a concern which I mentioned in this review.

test/Test/Cardano/Chain/Txp/Json.hs
```diff
@@ -33,17 +39,15 @@ goldenTxpConfiguration0 = goldenTestJSONPretty
...
--- TODO: This test likely fails due to the `NetworkMagic` field that exists
--- in `AddrAttributes` in the Byron release. As a result there are
--- serialization differences in the example addresses. When it has been
--- decided how we are differentiating between the different networks,
--- this should be fixed.
-{-
 goldenTxpConfiguration1 :: Property
 goldenTxpConfiguration1 =
     goldenTestJSONPretty exampleTxpConfiguration1
         "test/golden/json/txp/TxpConfiguration1"
```
mhuesch, line 45: I am not sure whether this `TxpConfiguration` test was failing due to `@NetworkMagic@` decoding, because it seems like the [example addresses](https://github.com/input-output-hk/cardano-ledger/blob/68bd2fa78fd3a06c350ae7907c2d1ceec871ebc3/test/Test/Cardano/Chain/Txp/Json.hs#L64) all has [@NetworkMagic@ constructors](https://github.com/input-output-hk/cardano-sl/blob/57a0d236eed59c48214ee70128549d6cfdbc73db/core/test/Test/Pos/Core/ExampleHelpers.hs#L211-L232) of `NetworkMainOrStage`, so they should have all enc/decoded to be consistent with the `@NetworkMagic@`-unaware `@Address@`s.

I suspected that it could be related to the [excision of AddrStakeDistribution](https://github.com/input-output-hk/cardano-ledger/commit/f4a95848b7ba22d3a2f2cf5b8dad6507b0530687), but that was more recent than the commit here, originally by @Jimbo.

At this stage, going back and diagnosing what the root failure was is doable, but a bit involved.

In this commit I opted to just modify both the example datatypes and golden files. So they will not give us guarantees of compatibility with `cardano-sl`, but will give us assurance going forwards around @NetworkMagic@ serialization.

ruhatch (approved):
Looks great, thanks!

test/Test/Cardano/Chain/Txp/Json.hs
Jimbo4350: This is fine! At that point address discrimination wasn't introduced into `cardano-ledger` and I wasn't sure what the final decision was going to be and therefore left the comment.

Jimbo4350 (approved):
LGTM!

test/Test/Cardano/Chain/Txp/Example.hs
mhuesch: I am not sure what it means.

test/Test/Cardano/Chain/Txp/Json.hs
mhuesch: Ok, sounds great!

test/Test/Cardano/Chain/Common/Address.hs
mhuesch: Just opened an issue: https://github.com/input-output-hk/cardano-ledger/issues/400

PR #401, opened 2019-04-01
Add compact data types to be used in optimizing UTxO

intricate:
This PR adds the compact/optimized data types:

- `CompactTxIn`
- `CompactTxId`
- `CompactTxOut`
- `CompactAddress`

In addition to this, for each of them, it also adds Hedgehog generators, round-trip tests, and tests which ensure that `heapWords` savings are gained.

_N.B. This is the first PR of two. The second PR will actually optimize/compact the `UTxO` data type by utilizing those which have been added in this PR._

### DIFF
--- a/src/Cardano/Chain/Common/Compact.hs
+++ b/src/Cardano/Chain/Common/Compact.hs
@@ -1,0 +1,43 @@
+{-# LANGUAGE DeriveAnyClass             #-}
+{-# LANGUAGE DeriveGeneric              #-}
+{-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE OverloadedStrings          #-}
+
+module Cardano.Chain.Common.Compact
+  ( CompactAddress
+  , toCompactAddress
+  , fromCompactAddress
+  )
+where
+
+import Cardano.Prelude
+
+import Cardano.Binary.Class (serialize', decodeFull')
+import qualified Data.ByteString.Short as BSS (fromShort, toShort)
+import Data.ByteString.Short (ShortByteString)
+
+import Cardano.Chain.Common.Address (Address(..))
+
+--------------------------------------------------------------------------------
+-- Compact Address
+--------------------------------------------------------------------------------
+
+-- | A compact in-memory representation for an 'Address'.
+--
+-- Convert using 'toCompactAddress' and 'fromCompactAddress'.
+--
+newtype CompactAddress = CompactAddress ShortByteString
+  deriving (Eq, Ord, Generic, Show)
+  deriving newtype HeapWords
+  deriving anyclass NFData
+
+toCompactAddress :: Address -> CompactAddress
+toCompactAddress addr =
+  CompactAddress (BSS.toShort (serialize' addr))
+
+fromCompactAddress :: CompactAddress -> Address
+fromCompactAddress (CompactAddress addr) =
+  case decodeFull' (BSS.fromShort addr) of
+    Left err      -> panic ("fromCompactAddress: impossible: " <> show err)
+    Right decAddr -> decAddr
--- a/src/Cardano/Chain/Txp/Compact.hs
+++ b/src/Cardano/Chain/Txp/Compact.hs
@@ -1,0 +1,143 @@
+-- | The UTxO is large and is kept in-memory. It is important to use as
+-- small a representation as possible to keep overall memory use reasonable.
+--
+-- This module provides a special compact representation for data types
+-- contained within the UTxO.
+--
+-- The idea here is that the compact representation is optimised only for
+-- storage size and does not have to be the same as the representation used
+-- when operating on the data. Conversion functions are to be used when
+-- inserting and retrieving values from the UTxO.
+--
+
+{-# LANGUAGE DeriveAnyClass             #-}
+{-# LANGUAGE DeriveGeneric              #-}
+{-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE OverloadedStrings          #-}
+
+module Cardano.Chain.Txp.Compact
+  ( CompactTxIn
+  , toCompactTxIn
+  , fromCompactTxIn
+
+  , CompactTxId
+  , toCompactTxId
+  , fromCompactTxId
+
+  , CompactTxOut
+  , toCompactTxOut
+  , fromCompactTxOut
+  )
+where
+
+import Cardano.Prelude
+
+import Crypto.Hash (digestFromByteString)
+import Cardano.Crypto.Hashing (AbstractHash (..))
+import Data.Binary.Get (Get, getWord64le, runGet)
+import Data.Binary.Put (Put, putWord64le, runPut)
+import qualified Data.ByteArray as BA (convert)
+import qualified Data.ByteString.Lazy as BSL (fromStrict, toStrict)
+
+import Cardano.Chain.Common.Compact
+  ( CompactAddress
+  , fromCompactAddress
+  , toCompactAddress
+  )
+import Cardano.Chain.Common.Lovelace (Lovelace)
+import Cardano.Chain.Txp.Tx (TxId, TxIn (..), TxOut (..))
+
+--------------------------------------------------------------------------------
+-- Compact TxIn
+--------------------------------------------------------------------------------
+
+-- | A compact in-memory representation for a 'TxIn'.
+--
+-- Convert using 'toCompactTxIn' and 'fromCompactTxIn'.
+--
+data CompactTxIn = CompactTxInUtxo {-# UNPACK #-} !CompactTxId
+                                   {-# UNPACK #-} !Word32
+  deriving (Eq, Ord, Generic, Show)
+  deriving anyclass NFData
+
+instance HeapWords CompactTxIn where
+  heapWords _ = 6
+
+toCompactTxIn :: TxIn -> CompactTxIn
+toCompactTxIn (TxInUtxo txId txIndex) =
+  CompactTxInUtxo (toCompactTxId txId) txIndex
+
+fromCompactTxIn :: CompactTxIn -> TxIn
+fromCompactTxIn (CompactTxInUtxo compactTxId txIndex) =
+  TxInUtxo (fromCompactTxId compactTxId) txIndex
+
+--------------------------------------------------------------------------------
+-- Compact TxId
+--------------------------------------------------------------------------------
+
+-- | A compact in-memory representation for a 'TxId'.
+--
+-- Convert using 'toCompactTxId' and 'fromCompactTxId'.
+--
+-- Compared to a normal 'TxId', this takes 5 heap words rather than 12.
+--
+data CompactTxId = CompactTxId {-# UNPACK #-} !Word64
+                               {-# UNPACK #-} !Word64
+                               {-# UNPACK #-} !Word64
+                               {-# UNPACK #-} !Word64
+  deriving (Eq, Generic, Ord, Show)
+  deriving anyclass NFData
+
+instance HeapWords CompactTxId where
+  heapWords _ = 5
+
+getCompactTxId :: Get CompactTxId
+getCompactTxId =
+  CompactTxId <$> getWord64le
+              <*> getWord64le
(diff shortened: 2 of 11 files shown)


src/Cardano/Chain/Common/Compact.hs
```diff
@@ -0,0 +1,40 @@
...
+--------------------------------------------------------------------------------
+
+-- | A compact in-memory representation for an 'Address'.
+--
+-- Convert using 'toCompactAddress' and 'fromCompactAddress'.
+--
+newtype CompactAddress = CompactAddress ShortByteString
+  deriving (Eq, Ord, Generic, Show)
+  deriving newtype HeapWords
+  deriving anyclass NFData
```
erikd, line 33: :100:

src/Cardano/Chain/Txp/Compact.hs
```diff
@@ -0,0 +1,202 @@
...
+-- small a representation as possible to keep overall memory use reasonable.
+--
+-- This module provides a special compact representation for data types
+-- contained within the UTxO.
+--
+-- The idea here is that the compact representation is optimised only for
+-- storage size and does not have to be the same as the representation used
+-- when operating on the data. Conversion functions are to be used when
+-- inserting and retrieving values from the UTxO.
+--
```
erikd, line 11: Love good comments.

src/Cardano/Chain/Txp/Compact.hs
```diff
@@ -0,0 +1,202 @@
...
+    --
+    -- - 1 word for the 'CompactTxInUtxo' object header
+    -- - 4 words (on a 64-bit arch) for the unpacked 'CompactTxId'
+    -- - 1 word for the unpacked 'Word32'
+    --
+    -- +---------------------------------------------+
+    -- │CompactTxInUtxo│Word#|Word#│Word#│Word#│Word#│
+    -- +---------------------------------------------+
+    --
+    = 6
```
erikd, line 78: Is there a way to write a test to validate this? Don't want this to be a PR blocker, but something to think about.

test/Test/Cardano/Chain/Common/Compact.hs
```diff
@@ -0,0 +1,60 @@
...
+
+roundTripCompactAddress :: Property
+roundTripCompactAddress =
+  eachOf 1000 genAddress (trippingCompact toCompactAddress fromCompactAddress)
+
+prop_heapWordsSavingsCompactAddress :: Property
+prop_heapWordsSavingsCompactAddress = withTests 1000 $ property $ do
+  addr <- forAll genAddress
+  let compactAddr = toCompactAddress addr
+  assert $ heapWords compactAddr < heapWords addr
```
erikd, line 42: Again not something that would block this PR, but it would be nice to come up with a stronger assertion there.

test/Test/Cardano/Chain/Txp/Compact.hs
```diff
@@ -0,0 +1,92 @@
...
+
+roundTripCompactTxOut :: Property
+roundTripCompactTxOut =
+  eachOf 1000 genTxOut (trippingCompact toCompactTxOut fromCompactTxOut)
+
+prop_heapWordsSavingsCompactTxOut :: Property
+prop_heapWordsSavingsCompactTxOut = withTests 1000 $ property $ do
+  txOut <- forAll genTxOut
+  let compactTxOut = toCompactTxOut txOut
+  assert $ heapWords compactTxOut < heapWords txOut
```
erikd, line 84: Same here. That is not a very strong assertion.

erikd (approved):

src/Cardano/Chain/Common/Compact.hs
```diff
@@ -0,0 +1,40 @@
...
+
+toCompactAddress :: Address -> CompactAddress
+toCompactAddress addr = do
+  CompactAddress (BSS.toShort (serialize' addr))
+
+fromCompactAddress :: CompactAddress -> Address
+fromCompactAddress (CompactAddress addr) =
+  unsafeDeserialize' (BSS.fromShort addr)
```
ruhatch: Not sure about using `unsafeDeserialize'` here, as I'd actually like to remove that function. I'd rather `decodeFull'` and then throw the exception yourself if that's what you want

src/Cardano/Chain/Txp/Compact.hs
ruhatch: :+1:

src/Cardano/Chain/Txp/Compact.hs
```diff
@@ -0,0 +1,202 @@
...
+getCompactTxId :: Get CompactTxId
+getCompactTxId =
+  CompactTxId <$> getWord64le
+              <*> getWord64le
+              <*> getWord64le
+              <*> getWord64le
+
+putCompactTxId :: CompactTxId -> Put
+putCompactTxId (CompactTxId a b c d) = do
+  putWord64le a >> putWord64le b
```
ruhatch, line 137: This seems like an odd combination of `do` and `>>` - I think just `do` would be fine

src/Cardano/Chain/Txp/Compact.hs
```diff
@@ -0,0 +1,202 @@
...
+    -- +----------------------+
+    -- │CompactTxOut│ * │Word#│
+    -- +--------------+-------+
+    --                |
+    --                v
+    --                +--------------+
+    --                │BA#│sz│payload│
+    --                +--------------+
+    --
+    = 3 + (heapWordsUnpacked compactAddr)
```
ruhatch: These brackets are redundant - gotta catch the important stuff in these review ;) haha

test/Test/Cardano/Chain/Common/Compact.hs
```diff
@@ -0,0 +1,60 @@
...
+import Test.Cardano.Chain.Common.Gen (genAddress)
+import Test.Cardano.Prelude (discoverRoundTrip, eachOf)
+
+--------------------------------------------------------------------------------
+-- Compact Address
+--------------------------------------------------------------------------------
+
+roundTripCompactAddress :: Property
+roundTripCompactAddress =
+  eachOf 1000 genAddress (trippingCompact toCompactAddress fromCompactAddress)
```
ruhatch, line 36: Does `tripping` not work here? Just wondering :)

test/Test/Cardano/Chain/Common/Compact.hs
```diff
@@ -0,0 +1,60 @@
...
+  addr <- forAll genAddress
+  let compactAddr = toCompactAddress addr
+  assert $ heapWords compactAddr < heapWords addr
+
+-------------------------------------------------------------------------------
+-- Tripping util
+-------------------------------------------------------------------------------
+
+trippingCompact
+  :: (MonadTest m, Show a, Eq a)
```
ruhatch: I think you'll want a `HasCallStack` on this - you can check by breaking something and seeing if you get the diffs out

test/Test/Cardano/Chain/Common/Compact.hs
```diff
@@ -0,0 +1,60 @@
...
+  => (a -> b) -> (b -> a) -> a -> m ()
+trippingCompact toCompact fromCompact x =
+  x === fromCompact (toCompact x)
+
+-------------------------------------------------------------------------------
+-- Main test export
+-------------------------------------------------------------------------------
+
+tests :: IO Bool
+tests = (&&) <$> H.checkParallel $$(discover) <*> H.checkParallel
```
ruhatch: `discover` brackets are redundant :)

test/Test/Cardano/Chain/Txp/Compact.hs
```diff
@@ -0,0 +1,92 @@
...
+
+-------------------------------------------------------------------------------
+-- Tripping util
+-------------------------------------------------------------------------------
+
+trippingCompact
```
ruhatch, line 90: This is duplicated, but maybe it's not worth moving :man_shrugging:

ruhatch (approved):
Looks good, thanks! Just a couple of comments and I agree with @erikd that we need something stronger for `HeapWords`. Did you mention @mhuesch is working on something?

src/Cardano/Chain/Txp/Compact.hs
intricate: Oops.

test/Test/Cardano/Chain/Common/Compact.hs
intricate: Oops.

src/Cardano/Chain/Txp/Compact.hs
intricate: Not sure why I did it like this. I might just remove the `do` and use `>>`s.

test/Test/Cardano/Chain/Common/Compact.hs
intricate: I originally was using Hedgehog's `tripping`.
```
tripping :: (MonadTest m, Applicative f, Show b, Show (f a), Eq (f a), HasCallStack) => a -> (a -> b) -> (b -> f a) -> m ()
```
but it seemed odd to use something like `(Identity . fromCompact*)` for its `(b -> f a)` parameter. What's your opinion on this?

ruhatch: I guess you could use `Identity` and `tripping` inside `trippingCompact`?

mhuesch: @ruhatch I was encouraging Luke to use `===` and function composition (without `tripping`) because the functions don't have functor wrappers. However `tripping` gives us good error messages. Is that your main preference for it? If so I think we can go with `tripping` and `Identity`

src/Cardano/Chain/Common/Compact.hs
mhuesch: Agree with @ruhatch - rethrowing the exception, or even `panic`-ing with a prefix like `"fromCompactAddress: impossible: " <> show exn` would make it easier to track the problem down, in case the impossible happened.

src/Cardano/Chain/Txp/Compact.hs
mhuesch: @erikd do you mean a test which checks the actual heap words at test-suite execution time? @intricate and I talked about using the `ghc-datasize`-based tool in a test, but thought it was too brittle (we'd then have to update the tools each time we bumped GHC versions) and inaccurate (the byte estimates given are often different from what `HeapWords` says, and the transformation between the two values is not straightforwards e.g. sharing isn't taken into account) to be worth it.

test/Test/Cardano/Chain/Common/Compact.hs
mhuesch: Yeah, I wonder if there's some linear function we could assert as the relationship between compact and non-compact versions.

mhuesch (commented):
LGTM, left a few comments.

test/Test/Cardano/Chain/Common/Compact.hs
intricate: The error messages for both `===` and `tripping` are pretty insightful but I've decided to go with `tripping` just to stay consistent with other round-trip tests in the codebase.

src/Cardano/Chain/Txp/Compact.hs
intricate: I've spoken to Duncan about this and he agrees that it's a "nice to have" check that we could postpone for now. I'll open an issue for this just so we don't totally forget about it.

test/Test/Cardano/Chain/Common/Compact.hs
intricate: I won't block the PR on this but will put some thought into it.

test/Test/Cardano/Chain/Txp/Compact.hs
intricate: https://github.com/input-output-hk/cardano-ledger/pull/401#discussion_r271446003

src/Cardano/Chain/Txp/Compact.hs
intricate: Can thank Duncan for this one :stuck_out_tongue:

src/Cardano/Chain/Common/Compact.hs
intricate: Decided to use `decodeFull'` with a `panic` in the error case.

PR #402, opened 2019-04-02
Use Data.Map.Strict everywhere

### LINKED ISSUE #382: Use Data.Map.Strict everywhere
## Background

We noticed performance issues from using lazy maps, but we've accidentally reintroduced `Data.Map`, where we should be using `Data.Map.Strict`.

## Proposed Solution

Replace all locations of `Data.Map` with `Data.Map.Strict`

erikd:
This arose as a result of suggestions from the auditors.

Closes #382


intricate (approved):

dcoutts (approved):

dnadales (approved):

ruhatch:
Closes #382

PR #403, opened 2019-04-02
[#398] Remove Undos from the codebase

### LINKED ISSUE #398: Remove Undos from the codebase
## Background

`Undo`s were used in `cardano-sl` to rollback the chain, but the new consensus layer does not make use of these structures. We should be able to remove them entirely.

## Proposed Solution

Remove the `Undo` data type and any other types that are only used in that part of the code.

ruhatch:
Closes #398


dnadales (approved):
:tada: :sparkles:

PR #406, opened 2019-04-03
[#314] Run weeder and pay off some other technical debt

### LINKED ISSUE #182: Rename BlockPSignatureHeavy constructor
We could rename this to `BlockProxySignature` as we only have one kind of delegation now and the `P` is not clear.

### LINKED ISSUE #314: Remove unused modules and other weeds
We have modules like `Test.Cardano.Chain.Genesis.Dummy` which do not seem to be used. We should run [`weeder`](http://hackage.haskell.org/package/weeder) to remove dead code.

### LINKED ISSUE #395: Remove the underscores from Tx's accessors
## Background

We do not use `lens` in `cardano-ledger`, so it seems there's no reason for starting the [accessor names with underscore](https://github.com/input-output-hk/cardano-ledger/blob/03cf7207c4961fec6ca22a074759b9cd4014fa06/src/Cardano/Chain/Txp/Tx.hs#L74). 

## Proposed Solution

Remove the underscores. Bonus points for removing other underscores in accessor names that might be still around.

ruhatch:
- Remove weed dependencies
- Remove dependency on lens in cardano-ledger
- Remove underscores in Tx accessors
- Rename some newtype accessors get -> un
- Remove BlockSignature constructor (we only use delegation signatures)

Closes #314
Closes #395
Closes #182


src/Cardano/Chain/Block/Header.hs
```diff
@@ -161,43 +154,53 @@ instance B.Buildable (WithEpochSlots Header) where
...
-    (consensusSlot consensus)
-    (unChainDifficulty $ consensusDifficulty consensus)
-    (consensusLeaderKey consensus)
-    (consensusSignature consensus)
-    (headerExtraData header)
-  where
-    headerHash :: HeaderHash
-    headerHash = hashHeader es header
-    consensus  = headerConsensusData header
+renderHeader es header = bprint
```
dnadales, line 157: I liked the old format better. But that's highly subjective.

src/Cardano/Chain/Block/Header.hs
```diff
@@ -161,43 +154,53 @@ instance B.Buildable (WithEpochSlots Header) where
...
 -- blocks.
 encodeHeader' :: EpochSlots -> Header -> Encoding
-encodeHeader' es h
-  =  encodeListLen 5
-  <> encode (headerProtocolMagicId h)
-  <> encode (headerPrevHash h)
-  <> encode (headerProof h)
-  <> encodeConsensusData es (headerConsensusData h)
-  <> encode (headerExtraData h)
+encodeHeader' es h =
```
dnadales, line 194: Maybe something else for the style guide? ;)

src/Cardano/Chain/Block/Header.hs
```diff
@@ -556,7 +548,7 @@ decodeAConsensus epochSlots = do
   -- Next, we decode a 'SlotId' into a 'FlatSlotId': the `SlotId` used in
   -- 'AConsensusData' is encoded as a epoch and slot-count pair.
   epochAndSlotCount :: Annotated SlotId ByteSpan <- decodeAnnotated
-  pk <- decode
+  pk           <- decode
```
dnadales, line 551: It seems like we inserted random spaces here :thinking: Did you intend to align all the arrows?

src/Cardano/Chain/Genesis/Generate.hs
```diff
@@ -135,11 +134,11 @@ generateGenesisData startTime genesisSpec = do
         <$> zip dlgIssuersSecrets richSecrets
 
   genesisDlg <-
-    liftEither
-    .  first GenesisDataGenerationDelegationError
-    $  mkGenesisDelegation
-    $  M.elems (unGenesisDelegation $ gsHeavyDelegation genesisSpec)
-    <> genesisDlgList
+    mkGenesisDelegation
```
dnadales, line 137: Nice!

test/golden/bi/block/ConsensusData
```diff
@@ -1,9 +1,22 @@
-00: 84820b182f58404b6d7977346c445345
-10: 3553346653483665744e6f756958657a
-20: 4379456a4b63337447346a61306b466a
-30: 4f38717a616932365a4d5055454a6645
-40: 7931356f78356b8119270f8200584071
-50: d1a202ba60ee5c91f7eaf29170e9b60c
-60: 83a53294ea85fed529df1a0905181d03
-70: e4b0a5aa1b7346c0a4127c70ccabb3ae
-80: a4fec7323e4782eca1c95cacd1a40e
+000: 84820b182f58404b6d7977346c445345
```
dnadales, line 1: There's something I don't understand: what is the idea of the golden tests? I thought we shouldn't change the contents of the reference file. But I was wrong apparently :)

dnadales (approved):
Kudos for the simplifications and addressing technical debt!

test/golden/bi/block/ConsensusData
intricate: Yes, if you're forced to alter the contents of a golden file in order to make the tests pass, it indicates some change in (de)serialization output and, typically, we don't want to do this without good reason (especially if instances of the type in question are serialized on the blockchain).

Are we certain that the removed `BlockSignature (Signature ToSign)` constructor isn't used within *any* blocks on mainnet or testnets?

intricate: Yes, if you're forced to alter the contents of a golden file in order to make the tests pass, it indicates some change in (de)serialization output and, typically, we don't want to do this without good reason (especially if instances of the type in question are serialized on the blockchain).

Are we certain that the removed `BlockSignature (Signature ToSign)` constructor isn't used within *any* blocks on mainnet or testnets?

intricate (dismissed):
If this review is irrelevant, I don't mind it being dismissed. Just a bit concerned about the modification of the golden files and would like to ensure that this isn't going to affect our compatibility with existing chains.

dnadales:
> If this review is irrelevant, I don't mind it being dismissed. Just a bit concerned about the modification of the golden files and would like to ensure that this isn't going to affect our compatibility with existing chains.

We have tests that check that we can decode (and validate) the existing mainnet mirror, so I guess we're covered here :) But maybe I'm wrong ...

src/Cardano/Chain/Block/Header.hs
ruhatch: Sorry, yeah, I just ran `brittany` on the whole file, because I was being lazy!

test/golden/bi/block/ConsensusData
ruhatch: @intricate Yeah, I ran a test over the whole of mainnet to see if we used it and we didn't.

@disassembler can you make sure that we never use the `BlockSignature` constructor (that is regular signatures, rather than delegated ones) going forward?

ruhatch: Also as @dnadales said we have a test that runs over mainnet (although we should probably update the mainnet mirror to include newer epochs)

ruhatch: @dnadales in general the point of the golden tests is to stop us changing the format *without us realizing we're doing it*. In this case the golden test for the `Header` and `ConsensusData` contained a `BlockSignature`, but we never actually used those on the chain, and we shouldn't (they're not in the specification). So I've replaced those with `ProxySignature` versions and updated the golden tests accordingly.

PR #407, opened 2019-04-03
Finish the implementation of the update mechanism.

### LINKED ISSUE #189: Implement ledger part of update interface

dnadales:
- Implement `PVBUMP`, `UPIEND` and `UPIEC` rules
- Set the update proposal time to live to `ppUpdateImplicit`.
- Use named field puns in the interface more consistently.
- Add an environment to `Cardano.Chain.Update.Validation.Registration`
- Use explicit exports in the update modules.

Closes #189.

### DIFF
--- a/src/Cardano/Chain/Update/ProtocolParameters.hs
+++ b/src/Cardano/Chain/Update/ProtocolParameters.hs
@@ -44,6 +44,16 @@ data ProtocolParameters = ProtocolParameters
   , ppUpdateVoteThd     :: !LovelacePortion
   , ppUpdateProposalThd :: !LovelacePortion
   , ppUpdateImplicit    :: !FlatSlotId
+  -- ^ Time to live for a protocol update proposal. This used to be the number
+  -- of slots after which the system made a decision regarding an update
+  -- proposal confirmation, when a majority of votes was not reached in the
+  -- given number of slots. If there were more positive than negative votes the
+  -- proposal became confirmed, otherwise it was rejected. Since in the
+  -- Byron-Shelley bridge we do not have negative votes, and we aim at
+  -- simplifying the update mechanism, 'ppUpdateImplicit' is re-interpreted as
+  -- the number of slots a proposal has to gather a majority of votes. If a
+  -- majority of votes has not been reached before this period, then the
+  -- proposal is rejected.
   , ppSoftforkRule      :: !SoftforkRule
   , ppTxFeePolicy       :: !TxFeePolicy
   , ppUnlockStakeEpoch  :: !EpochIndex
--- a/src/Cardano/Chain/Update/Validation/Interface.hs
+++ b/src/Cardano/Chain/Update/Validation/Interface.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE NamedFieldPuns #-}
 
@@ -14,18 +13,26 @@ module Cardano.Chain.Update.Validation.Interface
     -- * Interface functions
   , registerProposal
   , registerVote
+  , registerEndorsement
+  , registerEpoch
   )
 where
 
 import Cardano.Prelude hiding (State)
 
 import qualified Data.Map.Strict as M
+import Data.Set (union)
+import qualified Data.Set as S
 
 import Cardano.Chain.Slotting (EpochIndex, FlatSlotId)
 import Cardano.Chain.Common.StakeholderId (StakeholderId)
+import Cardano.Chain.Common.BlockCount (BlockCount)
 
 import Cardano.Chain.Update.ApplicationName (ApplicationName)
-import Cardano.Chain.Update.ProtocolParameters (ProtocolParameters)
+import Cardano.Chain.Update.ProtocolParameters
+  ( ProtocolParameters
+  , ppUpdateImplicit
+  )
 import Cardano.Chain.Update.ProtocolVersion (ProtocolVersion)
 import Cardano.Chain.Update.SoftwareVersion
   ( NumSoftwareVersion
@@ -33,29 +40,46 @@ import Cardano.Chain.Update.SoftwareVersion
   , svAppName
   , svNumber
   )
+import Cardano.Chain.Update.Validation.Endorsement
+  ( CandidateProtocolUpdate
+  , Endorsement
+  , endorsementProtocolVersion
+  )
+import qualified Cardano.Chain.Update.Validation.Endorsement as Endorsement
+import Cardano.Chain.Update.Validation.Interface.ProtocolVersionBump
+  ( tryBumpVersion
+  )
+import qualified Cardano.Chain.Update.Validation.Interface.ProtocolVersionBump as PVBump
 import qualified Cardano.Chain.Update.Validation.Registration as Registration
 import qualified Cardano.Chain.Update.Validation.Voting as Voting
-import Cardano.Chain.Update.Vote (UpId, AProposal, recoverUpId, AVote)
+import Cardano.Chain.Update.Vote (AProposal, AVote, UpId, recoverUpId)
 import Cardano.Crypto (ProtocolMagicId)
 
-data TODO
 
 data Environment = Environment
-  { protocolMagic :: ProtocolMagicId
-  , currentEpoch  :: !EpochIndex
+  { protocolMagic :: !ProtocolMagicId
   , currentSlot   :: !FlatSlotId
   , delegationMap :: !(Map StakeholderId StakeholderId)
+  , k             :: !BlockCount
+  -- ^ TODO: this is the chain security parameter, a.k.a. @stableAfter@, it is not part
+  -- of our protocol parameters, so it seems that we need to pass it in the
+  -- environment. However we need to double-check this with others.
+  , numGenKeys    :: !Word8
+  -- ^ Number of genesis keys. This is used to calculate the proportion of
+  -- genesis keys that need to endorse a new protocol version for it to be
+  -- considered for adoption. See
+  -- @Cardano.Chain.Update.Validation.Endorsement.Environment@.
   }
 
 -- | Update interface state.
 data State = State
-  { prevEpoch                         :: !EpochIndex
-    -- ^ Previously seen epoch
+  { currentEpoch                      :: !EpochIndex
+    -- ^ Current epoch
   , adoptedProtocolVersion            :: !ProtocolVersion
-  , adoptedProtocolParams             :: !ProtocolParameters
+  , adoptedProtocolParameters         :: !ProtocolParameters
     -- ^ Adopted protocol parameters
-  , futureAdopts                      :: !TODO -- We should take this from Cardano.Chain.Update.Validation.Endorsement
-    -- ^ Future protocol version adoptions
+  , candidateProtocolUpdates          :: ![CandidateProtocolUpdate]
+    -- ^ Candidate protocol versions
   , appVersions                       :: !(Map ApplicationName (NumSoftwareVersion, FlatSlotId))
     -- ^ Current application versions (by application name)
   , registeredProtocolUpdateProposals :: !(Map UpId (ProtocolVersion, ProtocolParameters))
@@ -66,20 +90,21 @@ data State = State
     -- ^ Confirmed update proposals
   , proposalVotes                     :: !(Map UpId (Set StakeholderId))
     -- ^ Update proposals votes
-  , proposalsEndorsements             :: TODO -- We should take this from Cardano.Chain.Update.Validation.Endorsement
-                                              -- So this should be a @Set Endorsement@
+  , registeredEndorsements            :: !(Set Endorsement)
     -- ^ Update proposals endorsements
-  , proposalRegistrationSlot          :: Map UpId FlatSlotId
+  , proposalRegistrationSlot          :: !(Map UpId FlatSlotId)
     -- ^ Slot at which an update proposal was registered
   }
 
 data Error
   = Registration Registration.Error
   | Voting Voting.Error
+  | Endorsement Endorsement.Error
+  | NumberOfGenesisKeysTooLarge (Registration.TooLarge Int)
 
 -- | Register an update proposal.
 --
--- This corresponds to the @UPIREG@ rules in the spec.
+-- This corresponds to the @UPIREG@ rule in the Byron ledger specification.
 registerProposal
   :: MonadError Error m
   => Environment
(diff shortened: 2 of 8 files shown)


src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -66,58 +85,74 @@ data State = State
...
+      { adoptedProtocolVersion
+      , adoptedProtocolParameters
+      , appVersions
+      , registeredProtocolUpdateProposals
+      , registeredSoftwareUpdateProposals
+      , proposalRegistrationSlot
       } = st
 
-    regSubSt = Registration.State rpus raus
+    subEnv =
```
ruhatch, line 141: Nice :)

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -126,42 +161,205 @@ registerVote
...
+  :: MonadError Error m
+  => Environment
+  -> State
+  -> Endorsement
+  -> m State
+registerEndorsement env st endorsement = do
+  -- TODO: The domain of the delegation map should contain all the genesis keys.
+  -- However I do not know whether we want to determine the number of genesis
+  -- keys like this.
+  ngk <- if M.size delegationMap <= fromIntegral (maxBound :: Word8)
```
ruhatch: Why `Word8` here?

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -126,42 +161,205 @@ registerVote
...
+    pidsKeep = nonExpiredPids `union` confirmedPids
+
+    nonExpiredPids =
+      M.keysSet $ M.filter (currentSlot - u <=) proposalRegistrationSlot
+
+    confirmedPids = M.keysSet confirmedProposals
 
-    subSt = Voting.State vts cps
+    registeredProtocolUpdateProposals' =
+      M.restrictKeys registeredProtocolUpdateProposals pidsKeep
```
ruhatch, line 245: Do we want to define the `(<|)` operator - I think it could be good here!

src/Cardano/Chain/Update/Validation/Interface.hs
ruhatch: I do think it's reasonable to use the size of the delegation map if we know the initial rule puts them all in it though

src/Cardano/Chain/Update/Validation/Interface/ProtocolVersionBump.hs
```diff
@@ -0,0 +1,73 @@
...
+  -> State
+tryBumpVersion env st lastSeenEpoch =
+  if currentEpoch < lastSeenEpoch && not (null stableCandidates)
+  then
+    let CandidateProtocolUpdate
+          { cpuProtocolVersion
+          , cpuProtocolParameters
+          } = fromMaybe err $ head stableCandidates
+        err :: CandidateProtocolUpdate
+        err = panic $  "The list of stable candidates shouldn't be empty,"
```
ruhatch: Hmm, I think it might be better just to do a pattern match on `stableCandidates` here - I know that means there are two places to return `st`, but I think that's safer than this `err`

ruhatch (approved):
Looks great!! :tada:

src/Cardano/Chain/Update/Validation/Interface.hs
dnadales: > Why Word8 here?

Because of the type of `numGenesisKeys`, which we need to pass to the endorsement environment.

```haskell
data Environment = Environment
  { k                                 :: !BlockCount
  -- ^ Chain stability parameter.
  , currentSlot                       :: !FlatSlotId
  , delegationMap                     :: !(Map StakeholderId StakeholderId)
  , adoptedProtocolParameters         :: !ProtocolParameters
  , confirmedProposals                :: !(Map UpId FlatSlotId)
  , registeredProtocolUpdateProposals :: !Registration.ProtocolUpdateProposals
  , numGenesisKeys                    :: !Word8
  -- ^ Number of genesis keys. This is used in combination with the
  -- 'ppUpdateProposalThd' protocol parameter to calculate the proportion of
  -- genesis keys that need to endorse a new protocol version for it to be
  -- considered for adoption.
  }

```

> I do think it's reasonable to use the size of the delegation map if we know the initial rule puts them all in it though

￼The problem is that the rules and the implementation differ in the calculation of `canAdopt`:

image

```haskell
canAdopt
  :: Word8
  -- ^ Number of genesis keys.
  -> ProtocolParameters
  -> Set Endorsement
  -> ProtocolVersion
  -> Bool
canAdopt ngk pps endorsements protocolVersion =
  upAdptThd <= numberOfEndorsements
 where
  -- In Byron we do not have a @upAdptThd@ protocol parameter, so we have to
  -- use the existing ones.
  --
  -- @lovelacePortionToDouble . srMinThd . ppSoftforkRule@ will give us the
  -- ratio (in the interval @[0, 1]@) of the total stake that has to endorse a
  -- protocol version to become adopted. In genesis configuration, this ratio
  -- will evaluate to @0.6@, so if we have 7 genesis keys, @upAdptThd = 4@.
  upAdptThd :: Int
  upAdptThd = floor $ stakeRatio * fromIntegral ngk
   where
    stakeRatio = lovelacePortionToDouble . srMinThd . ppSoftforkRule $ pps

  numberOfEndorsements :: Int
  numberOfEndorsements = length $ Set.filter
    ((== protocolVersion) . endorsementProtocolVersion)
    endorsements

```

So we have no `numGenesisKeys` parameter in the specs.

Then either we let the two specs differ (and we pass an extra parameter to the environment to `registerEndorsement`), or we pass this extra parameter and adapt the specs. Any preference?

src/Cardano/Chain/Update/Validation/Interface.hs
dnadales: Yes, there's a TODO in the code that we could use the [`Relation`](https://github.com/input-output-hk/cardano-ledger-specs/blob/ed8424e9a85b9a3be44bb72e69a404426c7e334f/byron/ledger/executable-spec/src/Ledger/Core.hs) class defined by @nc6 in `cardano-ledger-specs`. We should definitely use this here in that case.

src/Cardano/Chain/Update/Validation/Interface.hs
ruhatch: I would suggest that we update the formal spec - if our protocol relies on the number of genesis keys in some way, we should probably let that be known!

src/Cardano/Chain/Update/Validation/Interface.hs
ruhatch: :ok_hand:

src/Cardano/Chain/Update/ProtocolParameters.hs
```diff
@@ -44,6 +44,16 @@ data ProtocolParameters = ProtocolParameters
...
+  -- ^ Time to live for a protocol update proposal. This used to be the number
+  -- of slots after which the system made a decision regarding an update
+  -- proposal confirmation, when a majority of votes was not reached in the
+  -- given number of slots. If there were more positive than negative votes the
+  -- proposal became confirmed, otherwise it was rejected. Since in the
+  -- Byron-Shelley bridge we do not have negative votes, and we aim at
+  -- simplifying the update mechanism, 'ppUpdateImplicit' is re-interpreted as
+  -- the number of slots a proposal has to gather a majority of votes. If a
+  -- majority of votes has not been reached before this period, then the
+  -- proposal is rejected.
```
mhuesch, line 56: 👍 I think I've said this before, but this is a nice simplification.

src/Cardano/Chain/Update/Validation/Interface.hs
mhuesch: I do think this is an instance of the spec & the system differing, in a way that could be confusing. @dnadales and I discussed this in a previous PR review: https://github.com/input-output-hk/cardano-ledger/pull/353#discussion_r265211366

We seem to be converting between "stake required to adopt" and "number of keys required to adopt". This is a simplification, and I think it makes things tidier, but it's still a difference which isn't explicitly outlined in the spec.

The issue with formalizing this, I would imagine, is that the OBFT era simplification of just tracking genesis keys isn't backwards compatible with OClassic.

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -126,42 +161,205 @@ registerVote
...
+      Endorsement.State
+        candidateProtocolUpdates
+        registeredEndorsements
+
+    u = ppUpdateImplicit adoptedProtocolParameters
+
+-- | Register an epoch. Whenever an epoch number is seen on a block this epoch
+-- number should be passed to this function so that on epoch change the
+-- protocol parameters can be updated, provided that there is an update
+-- candidate that was accepted and endorsed by a majority of the genesis keys.
```
mhuesch, line 301: 👍

src/Cardano/Chain/Update/Validation/Interface/ProtocolVersionBump.hs
```diff
@@ -0,0 +1,73 @@
...
+data Environment = Environment
+  { k           :: !BlockCount
+  , currentSlot :: !FlatSlotId
+  }
+
+data State = State
+  { currentEpoch              :: !EpochIndex
+  , nextProtocolVersion       :: !ProtocolVersion
+  , nextProtocolParameters    :: !ProtocolParameters
+  , candidateProtocolVersions :: ![CandidateProtocolUpdate]
```
mhuesch: I noticed that this part is in agreement with the executable-spec, but not with the formal-spec (which has `fads`/`candidateProtocolVersions` in the Env and doesn't modify it). Is this related to the spec changes where much of the state is cleared out on epoch transitions in `UPIEC`?

src/Cardano/Chain/Update/Validation/Interface/ProtocolVersionBump.hs
mhuesch: Agree with ⬆️ 

I'm a fan of case statements on tuples, so I might go for something like this. Not everyone likes the style though.

```
  case (currentEpoch < lastSeenEpoch, stableCandidates) of
    (True, (winner:_)) ->
      let CandidateProtocolUpdate
            { cpuProtocolVersion
            , cpuProtocolParameters
            } = winner
      in
        st { currentEpoch = lastSeenEpoch
           , nextProtocolVersion = cpuProtocolVersion
           , nextProtocolParameters = cpuProtocolParameters
           , candidateProtocolVersions =
               filter ((currentSlot - twice k <) . cpuSlot) candidateProtocolVersions
           }
    _ -> st
```

src/Cardano/Chain/Update/Validation/Registration.hs
```diff
@@ -199,32 +241,31 @@ canUpdate adoptedPP newPP proposal = do
...
 
   adoptedScriptVersion = ppScriptVersion adoptedPP
-  newScriptVersion     = ppScriptVersion newPP
+  newScriptVersion     = ppScriptVersion proposedPP
   scriptVersionDiff    = newScriptVersion - adoptedScriptVersion
 
-
 -- | Check that a new 'SoftwareVersion' is valid
 --
 --   We check that:
```
mhuesch, line 271: 👍

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -126,42 +161,205 @@ registerVote
...
+        st { currentEpoch = currentEpoch'
+           , adoptedProtocolVersion = adoptedProtocolVersion'
+           , adoptedProtocolParameters = nextProtocolParameters'
+           , candidateProtocolUpdates = []
+           , registeredProtocolUpdateProposals = M.empty
+           , confirmedProposals = M.empty
+           , proposalVotes = M.empty
+           , registeredEndorsements = S.empty
+           , proposalRegistrationSlot = M.empty
+           }
```
mhuesch, line 339: 👍 implements the new filtering of sets/lists on epoch transition

src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -126,42 +161,205 @@ registerVote
...
+                    , pid `elem` M.keys confirmedProposals'
                     ]
   pure $!
-    st { confirmedProposals = cps'
-       , proposalVotes = vts'
-       , appVersions = M.union avsNew avs
-       , registeredSoftwareUpdateProposals = M.withoutKeys raus (M.keysSet cps)
+    st { confirmedProposals = confirmedProposals'
+       , proposalVotes = proposalVotes'
+       -- Note that is important that the new application versions are passed
```
mhuesch: Grammar police 🚔 reporting for duty
```suggestion
       -- Note that it's important that the new application versions are passed
```

mhuesch (approved):
Looks good - left a few comments & questions.

src/Cardano/Chain/Update/Validation/Interface.hs
dnadales: What I can do is to add the number of genesis keys as an extra parameter in the environment of the validation interface, and adapt the formal spec afterwards.

If you guys don't think this is a good idea let me know!

src/Cardano/Chain/Update/Validation/Interface.hs
dnadales: Very much appreciated @mhuesch (or should I say officer? :P) Here you can see that I'm a native Spanish speaker :/

src/Cardano/Chain/Update/Validation/Interface/ProtocolVersionBump.hs
dnadales: Indeed! Thanks for catching this up @mhuesch I changed this now.

src/Cardano/Chain/Update/Validation/Interface/ProtocolVersionBump.hs
dnadales: I never would have thought of writing this in that way @mhuesch . I like this a lot, and it improved the code by not having that unnecessary `panic`.

Thank you guys!

src/Cardano/Chain/Update/Validation/Interface.hs
dnadales: Created https://github.com/input-output-hk/cardano-ledger-specs/issues/377

src/Cardano/Chain/Update/Validation/Interface/ProtocolVersionBump.hs
```diff
@@ -44,30 +44,23 @@ tryBumpVersion
...
-    in
-      st { currentEpoch = lastSeenEpoch
-         , nextProtocolVersion = cpuProtocolVersion
-         , nextProtocolParameters = cpuProtocolParameters
-         , candidateProtocolVersions =
-             filter ((currentSlot - twice k <) . cpuSlot) candidateProtocolVersions
-         }
-  else st
+  case (currentEpoch < lastSeenEpoch, stableCandidates) of
+    (True, newestStable:_) ->
```
ruhatch, line 48: Neat!

ruhatch (approved):
Looks great, thanks :+1:

src/Cardano/Chain/Update/Validation/Interface.hs
mhuesch: Heh 🙂

Issue #213, opened 2018-12-11
Add RTS options to the tests to limit memory usage


dnadales:
We have a situation in which tests consume all available memory. To detect this we should add RTS options that limit memory usage.

ruhatch:
Approach described in [this blog post](http://neilmitchell.blogspot.com/2015/09/detecting-space-leaks.html)

dnadales:
Could this be as simple as [what I did here](https://github.com/input-output-hk/cardano-ledger-specs/blob/ed8424e9a85b9a3be44bb72e69a404426c7e334f/byron/ledger/executable-spec/cs-ledger.cabal#L107)?

(This is also a way of having people check that I didn't do something stupid)

PR #411, opened 2019-04-03
Utilize optimized/compact data types in UTxO

intricate:
This PR utilizes the optimized data types, `CompactTxIn` and `CompactTxOut`, in the `UTxO` data type. Functions exposed by the `Cardano.Chain.Txp.UTxO` module still accept the same parameters such as `Tx`, `TxIn`, `UTxO`, etc. but internally utilize conversion functions defined in `Cardano.Chain.Txp.Compact`.

Manual inspection with `ghc-heap-view` and `ghc-datasize` indicates a savings of around ~51% in terms of `UTxO`'s heap memory consumption.

_Closes #213._

_N.B. This is the final PR following PR #401._

### DIFF
--- a/src/Cardano/Chain/Txp/UTxO.hs
+++ b/src/Cardano/Chain/Txp/UTxO.hs
@@ -1,5 +1,8 @@
 {-# LANGUAGE FlexibleContexts           #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE DeriveAnyClass             #-}
+{-# LANGUAGE DeriveGeneric              #-}
+{-# LANGUAGE DerivingStrategies         #-}
 
 module Cardano.Chain.Txp.UTxO
   ( UTxO
@@ -20,16 +23,26 @@ import Cardano.Prelude
 
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Map.Strict as M
+import qualified Data.Set as S
 
 import Cardano.Chain.Common
   (Address, Lovelace, LovelaceError, isRedeemAddress, sumLovelace)
 import Cardano.Chain.Txp.Tx (Tx(..), TxId, TxIn(..), TxOut(..))
+import Cardano.Chain.Txp.Compact
+  ( CompactTxIn
+  , CompactTxOut
+  , fromCompactTxOut
+  , toCompactTxIn
+  , toCompactTxOut
+  )
 import Cardano.Crypto (hash)
 
 
 newtype UTxO = UTxO
-  { unUTxO :: Map TxIn TxOut
-  } deriving HeapWords
+  { unUTxO :: Map CompactTxIn CompactTxOut
+  } deriving (Generic, Show)
+    deriving newtype HeapWords
+    deriving anyclass NFData
 
 data UTxOError
   = UTxOMissingInput TxIn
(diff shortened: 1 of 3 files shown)


dnadales:
> Manual inspection with ghc-heap-view and ghc-datasize indicates a savings of around ~51% in terms of UTxO's heap memory consumption.

:tada: 

BTW, do we want to measure performance as well *in the near future*? I could imagine this is the trade-off. For this PR we're too late, but it would have been nice to have some benchmarks...

dnadales (approved):
LGTM.

I only have a couple of questions regarding the overall picture:

- Where do we need the "expanded" forms of the compact data? In other words, why are they still around?
- What is the relation between this and https://github.com/input-output-hk/cardano-ledger/issues/410

src/Cardano/Chain/Txp/UTxO.hs
```diff
@@ -20,32 +23,45 @@ import Cardano.Prelude
...
   = UTxOMissingInput TxIn
   | UTxOOverlappingUnion
   deriving (Eq, Show)
 
 fromList :: [(TxIn, TxOut)] -> UTxO
-fromList = UTxO . M.fromList
+fromList = UTxO . M.fromList . toCompactTxInTxOutList
+ where
+  toCompactTxInTxOutList :: [(TxIn, TxOut)] -> [(CompactTxIn, CompactTxOut)]
+  toCompactTxInTxOutList = map (\(k, v) -> (toCompactTxIn k, toCompactTxOut v))
```
ruhatch:
```suggestion
  toCompactTxInTxOutList = map (bimap toCompactTxIn toCompactTxOut)
```

ruhatch (approved):
Looks good, but I would like to know more about performance before merging. I just submitted #418, which enables full mainnet validation for UTxO in the tests. We'll see how long that takes in CI and then you could rebase on that and see how long your stuff takes. That would give us a slightly better idea of the kind of performance hit we're looking at.

ruhatch:
It would also be good to implement #213 with this, to make sure that CI tests don't run over a certain amount of memory now

Jimbo4350 (approved):
LGTM!

src/Cardano/Chain/Txp/UTxO.hs
intricate: Thanks.

intricate:
@dnadales:

> * Where do we need the "expanded" forms of the compact data? In other words, why are they still around?

I wouldn't say that they're particularly necessary but, in my opinion, they're much easier to understand and work with. Working with the compact data types seems like it would be a bit non-obvious, especially types like `data CompactTxId = CompactTxId !Word64 !Word64 !Word64 !Word64`. 

Even if it were documented very well, I feel like `type TxId = Hash Tx` is pretty straightforward and can be understood just by looking at it. Because of this, I believe that the compact data types should be strictly reserved for optimizing the in-memory representation of the `UTxO`. This is definitely up for debate, though.

> * What is the relation between this and #410

I haven't looked at #410 myself but I could discuss this with @mhuesch.


@ruhatch:

>We'll see how long that takes in CI and then you could rebase on that and see how long your stuff takes. That would give us a slightly better idea of the kind of performance hit we're looking at.

Took about 17 minutes and seems to have had little effect on the run time (which you mentioned in https://github.com/input-output-hk/cardano-ledger/pull/418#issuecomment-479821158).

mhuesch:
@dnadales:

> Where do we need the "expanded" forms of the compact data? In other words, why are they still around?

I agree with what @intricate said above. The non-compact datatypes are easier to understand and work with. Certain of them, like `Address` are absolutely necessary, because a `CompactAddress` is defined as the serialization of an `Address`. I think we need the non-compact datatypes to deserialize things out of the blockchain. The next step would likely be to compact them, but we still need the non-compact datatypes to perform the deserialization operation.

> What is the relation between this and #410?

I am unsure which datatypes we'd want to implement `Diff` instances for. It's possible that we would want to diff "inside" of types which become opaque in their compact forms (e.g. `Address`). However I don't know which types we want to diff "inside of", and which we can ignore.

mhuesch (approved):
LGTM

Issue #336, opened 2019-02-22
Select which test to run via command line arguments


dnadales:
Currently there is no way to select which test to run:

```haskell
main :: IO ()
main = do
  opts <- execParser optsParser
  let scenario = optsTestScenario opts
  runTests
    [ Test.Cardano.Chain.Block.Bi.tests
    , Test.Cardano.Chain.Block.Validation.tests scenario
    , Test.Cardano.Chain.Common.Lovelace.tests
    , Test.Cardano.Chain.Delegation.Bi.tests
    , Test.Cardano.Chain.Epoch.File.tests
    , Test.Cardano.Chain.Genesis.Json.tests
    , Test.Cardano.Chain.Elaboration.Delegation.tests
    , Test.Cardano.Chain.Ssc.Bi.tests
    , Test.Cardano.Chain.Txp.Bi.tests
    , Test.Cardano.Chain.Txp.Json.tests
    , Test.Cardano.Chain.Txp.Validation.tests scenario
    , Test.Cardano.Chain.Update.Bi.tests
    , Test.Cardano.Chain.Update.Json.tests
    , Test.Cardano.Chain.Block.Validation.Spec.tests
    ]
```

It would be nice if we could select the test to run using `stack` in combination with `file-watch` and `ta` as follows:

```sh
stack test --file-watch --ta "-m *somePattern*"
```

There are options such as using `ghcid`, `ghci` or a combination of both however they have several shortcomings:

- No recompilation when `stack.yaml` or `*.cabal` files are changed
- No recompilation when local dependencies files are changed
- Colored output breaks in some cases

Also it might be nice to add a test framework, to avoid rolling our own.

ruhatch:
We should be able to write a function that converts from a `hedgehog` `TestGroup` to a `testy-hedgehog` `TestTree`, which would then give us this functionality out of the box

PR #417, opened 2019-04-04
Make previous hash of a boundary block an Either

erikd:
This turned out the be the least invasive changes that was still able to represent the fact that that previous hash of the zerotth boundary block does not exist.

### DIFF
--- a/scripts/nix/stack-shell.nix
+++ b/scripts/nix/stack-shell.nix
@@ -4,5 +4,5 @@ with pkgs;
 haskell.lib.buildStackProject {
   name = "cardano-ledger-env";
   buildInputs = [ zlib openssl git ];
-  ghc = haskell.packages.ghc863.ghc;
+  ghc = haskell.packages.ghc864.ghc;
 }
--- a/src/Cardano/Chain/Block/Block.hs
+++ b/src/Cardano/Chain/Block/Block.hs
@@ -42,12 +42,12 @@ module Cardano.Chain.Block.Block
   -- * BoundaryBlock
   , ABlockOrBoundary(..)
   , BoundaryValidationData(..)
-  , dropBoundaryBlock
   )
 where
 
 import Cardano.Prelude
 
+import Data.Coerce (coerce)
 import qualified Data.ByteString as BS
 import Data.Text.Lazy.Builder (Builder)
 import Formatting (bprint, build, int, shown)
@@ -198,17 +198,17 @@ data ABlockOrBoundary a
 --   now deprecated these explicit boundary blocks, but we still need to decode
 --   blocks in the old format. In the case that we find a boundary block, we
 --   drop it using 'dropBoundaryBlock' and return a 'Nothing'.
-decodeABlockOrBoundary :: EpochSlots -> Decoder s (ABlockOrBoundary ByteSpan)
-decodeABlockOrBoundary epochSlots = do
+decodeABlockOrBoundary :: EpochSlots -> Bool -> Decoder s (ABlockOrBoundary ByteSpan)
+decodeABlockOrBoundary epochSlots isEpochZero = do
   enforceSize "Block" 2
   decode @Word >>= \case
-    0 -> ABOBBoundary <$> dropBoundaryBlock
+    0 -> ABOBBoundary <$> dropBoundaryBlock isEpochZero
     1 -> ABOBBlock <$> decodeABlock epochSlots
     t -> cborError $ DecoderErrorUnknownTag "Block" (fromIntegral t)
 
-decodeBlockOrBoundary :: EpochSlots -> Decoder s (Maybe Block)
-decodeBlockOrBoundary epochSlots =
-  decodeABlockOrBoundary epochSlots >>= \case
+decodeBlockOrBoundary :: EpochSlots -> Bool -> Decoder s (Maybe Block)
+decodeBlockOrBoundary epochSlots isEpcohZero =
+  decodeABlockOrBoundary epochSlots isEpcohZero >>= \case
     ABOBBoundary _ -> pure Nothing
     ABOBBlock    b -> pure . Just $ void b
 
@@ -217,18 +217,19 @@ decodeBlockOrBoundary epochSlots =
 --------------------------------------------------------------------------------
 
 data BoundaryValidationData a = BoundaryValidationData
-  { boundaryBlockLength :: Int64
+  { boundaryBlockLength :: !Int64
   -- ^ The length of the boundary block in bytes
-  , boundaryPrevHash    :: HeaderHash
-  -- ^ The hash of the previous block
-  , boundaryHeaderBytes :: a
+  , boundaryPrevHash    :: !(Either GenesisHash HeaderHash)
+  -- ^ The hash of the previous block. Should only be GenesisHash for the
+  -- initial boundary block.
+  , boundaryHeaderBytes :: !a
   -- ^ Annotation representing the header bytes
   } deriving (Functor)
 
 -- | A decoder that drops the boundary block, but preserves the 'ByteSpan' of
 --   the header for hashing
-dropBoundaryBlock :: Decoder s (BoundaryValidationData ByteSpan)
-dropBoundaryBlock = do
+dropBoundaryBlock ::  Bool -> Decoder s (BoundaryValidationData ByteSpan)
+dropBoundaryBlock isEpochZero = do
   Annotated (Annotated hh bs) (ByteSpan start end) <- annotatedDecoder $ do
     enforceSize "BoundaryBlock" 3
     aHeaderHash <- annotatedDecoder dropBoundaryHeader
@@ -237,10 +238,11 @@ dropBoundaryBlock = do
     pure aHeaderHash
   pure $ BoundaryValidationData
     { boundaryBlockLength = end - start
-    , boundaryPrevHash    = hh
+    -- For the zeroth boundary block this field needs to be a 'GenesisHash'
+    -- and for all subsequent blocks it's a 'HeaderHash'.
+    , boundaryPrevHash    = if isEpochZero then Left (coerce hh) else Right hh
     , boundaryHeaderBytes = bs
     }
-
 
 --------------------------------------------------------------------------------
 -- Block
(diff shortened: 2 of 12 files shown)


src/Cardano/Chain/Block/Validation.hs
```diff
@@ -163,7 +162,10 @@ data ChainValidationError
   | ChainValidationInvalidDelegation PublicKey PublicKey
   -- ^ The delegation used in the signature is not valid according to the ledger
 
-  | ChainValidationInvalidHash HeaderHash HeaderHash
+  | ChainValidationInvalidHash HeaderHash (Maybe HeaderHash)
+  -- ^ The hash of the previous block does not match the value in the header
+
+  | ChainValidationInvalidGenesisHash (Maybe HeaderHash)
```
dnadales: Just for my understanding: what do we mean by genesis hash in this context? Is the hash of the genesis data?

Also what would be the meaning of:

```haskell
ChainValidationInvalidGenesisHash Nothing
```

I think [we will never generate such value](https://github.com/input-output-hk/cardano-ledger/pull/417/files#diff-bb2a507bf438a3ec8b2c55e563a36f21R202).

src/Cardano/Chain/Block/Block.hs
```diff
@@ -217,11 +217,11 @@ decodeBlockOrBoundary epochSlots =
 --------------------------------------------------------------------------------
 
 data BoundaryValidationData a = BoundaryValidationData
-  { boundaryBlockLength :: Int64
+  { boundaryBlockLength :: !Int64
   -- ^ The length of the boundary block in bytes
-  , boundaryPrevHash    :: HeaderHash
+  , boundaryPrevHash    :: !(Maybe HeaderHash)
```
dnadales: Could we add a comment that explains what the meaning of a `Nothing` value in this field means?

src/Cardano/Chain/Block/Validation.hs
erikd: > Is the hash of the genesis data?

Yes.

> ChainValidationInvalidGenesisHash Nothing

That is not valid, its just the result of `boundaryPrevHash bvd`. There may be a better way to do that. Will have a think.

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -190,18 +192,19 @@ data ChainValidationError
...
-      fromMaybe (configGenesisHeaderHash config) (cvsPreviousHash cvs)
-
-  -- Validate the previous block hash of 'b'
-  (boundaryPrevHash bvd == prevHash)
-    `orThrowError` ChainValidationInvalidHash prevHash (boundaryPrevHash bvd)
+updateChainBoundary cvs bvd = do
+  case cvsPreviousHash cvs of
+    Nothing ->
+      -- The case for the zeroth boundary block.
+      (boundaryPrevHash bvd /= Nothing)
```
ruhatch: I'm slightly confused - why don't we compare this to the `GenesisHash` anymore?

src/Cardano/Chain/Block/Validation.hs
dnadales: > > Is the hash of the genesis data?
>
> Yes.

Would it make sense to write `GenesisDataHash` then?



> That is not valid, its just the result of boundaryPrevHash bvd. There may be a better way to do that. Will have a think.

Maybe change the type of `ChainValidationInvalidGenesisHash HeaderHash`? :man_shrugging:

src/Cardano/Chain/Block/Validation.hs
erikd: We don't compare it to the genesis hash because for the in memory representation, the previous hash for the zeroth boundary block is `Nothing`. We still check this indirectly when we check the hash of the zeroth boundary block. It may still be nice to check the hash of the genesis hash directly,

erikd: Working on a better way to do this. It will actually check the genesis hash and it will correctly distinguish between the zeroth boundary block have a previous hash that is the `GenesisHash` and a non-zeroth boundary bbock  having a `HeaderHash` for the previous hash.

dnadales:
I don't know whether this could be converted into a [draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) BTW...

dnadales (commented):
LGTM. Although @ruhatch has more insight in our current code base :)

dnadales (approved):

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -65,7 +65,7 @@ import Cardano.Crypto
   , hashRaw
   )
 
-
+import qualified Prelude
```
ruhatch: Don't need this, use `panic` from `Cardano.Prelude` instead of `error`. If you get problems with `Show`, you can import qualified `Text.Show`.

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -195,22 +208,22 @@ updateChainBoundary
...
-    `orThrowError` ChainValidationInvalidHash prevHash (boundaryPrevHash bvd)
-
+  case (cvsPreviousHash cvs, boundaryPrevHash bvd) of
+    (Left expected, Left actual) ->
+        (expected == actual && actual == configGenesisHash config)
+          `orThrowError` ChainValidationGenesisHashMismatch expected actual
+    (Right expected, Right actual) ->
+        (expected == actual)
+          `orThrowError` ChainValidationInvalidHash expected actual
+    _ -> Prelude.error $ "updateChainBoundary: " ++ show (cvsPreviousHash cvs, boundaryPrevHash bvd)
```
ruhatch: Yeah, so `panic` here

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -195,22 +208,22 @@ updateChainBoundary
...
-    prevHash =
-      fromMaybe (configGenesisHeaderHash config) (cvsPreviousHash cvs)
-
-  -- Validate the previous block hash of 'b'
-  (boundaryPrevHash bvd == prevHash)
-    `orThrowError` ChainValidationInvalidHash prevHash (boundaryPrevHash bvd)
-
+  case (cvsPreviousHash cvs, boundaryPrevHash bvd) of
+    (Left expected, Left actual) ->
+        (expected == actual && actual == configGenesisHash config)
```
ruhatch: I don't think we should be using `configGenesisHash` twice. I think it should either be in the initial rule or here, but probably not both.

src/Cardano/Chain/Genesis/Hash.hs
```diff
@@ -11,9 +14,11 @@ import Cardano.Prelude
 import Cardano.Binary.Class (Raw)
 import Cardano.Crypto.Hashing (Hash)
 
+-- import Control.DeepSeq (NFData)
```
ruhatch: Can delete this

test/test.hs
```diff
@@ -34,7 +34,10 @@ main :: IO ()
 main = do
   opts <- execParser optsParser
   let scenario = optsTestScenario opts
-  runTests
+  if True
```
ruhatch: I think this one's probably not needed anymore ;)

ruhatch (approved):
Looks good now :) I think we want to make sure that this is reflected in the specs somehow @dnadales @nc6

test/test.hs
dnadales: No, but solving [this](https://github.com/input-output-hk/cardano-ledger/issues/336) will make me happy :) :tulip:

dnadales:
> Looks good now :) I think we want to make sure that this is reflected in the specs somehow @dnadales @nc6

I wonder whether there's anything to reflect at all: the state of the chain rules just specifies that the previous block hash is just a hash, and we don't have an initial chain rule. Any ideas?

src/Cardano/Chain/Block/Validation.hs
erikd: The `error` was actually just while I was working on this. I had intended to provide a proper error.

test/test.hs
erikd: Bah, ended up with the wrong code in this PR.

scripts/nix/stack-shell.nix
```diff
@@ -4,5 +4,5 @@ with pkgs;
 haskell.lib.buildStackProject {
   name = "cardano-ledger-env";
   buildInputs = [ zlib openssl git ];
-  ghc = haskell.packages.ghc863.ghc;
+  ghc = haskell.packages.ghc864.ghc;
```
nc6, line 7: Is this file still in use any more, I wonder?

src/Cardano/Chain/Block/Block.hs
```diff
@@ -237,11 +238,12 @@ dropBoundaryBlock = do
     pure aHeaderHash
   pure $ BoundaryValidationData
     { boundaryBlockLength = end - start
-    , boundaryPrevHash    = hh
+    -- For the zeroth boundary block this field needs to be a 'GenesisHash'
+    -- and for all subsequent blocks its a 'HeaderHash'.
```
nc6:
```suggestion
    -- and for all subsequent blocks it's a 'HeaderHash'.
```

src/Cardano/Chain/Genesis/Hash.hs
```diff
@@ -11,9 +14,9 @@ import Cardano.Prelude
 import Cardano.Binary.Class (Raw)
 import Cardano.Crypto.Hashing (Hash)
 
-
 newtype GenesisHash = GenesisHash
   { unGenesisHash :: Hash Raw
-  }
+  } deriving (Eq, Generic)
+    deriving newtype NFData
```
nc6: I think `DerivingStrategies` isn't really needed here. `DeriveAnyClass` is not turned on, so GND would be used to derive `NFData` anyway. Maybe it's better to be explicit. But this is slightly confusing - `Eq` will also be derived using GND!

nc6 (approved):
Minor comments only, otherwise LGTM

scripts/nix/stack-shell.nix
erikd: Its Nix, I have no idea!

erikd:
\o/

PR #418, opened 2019-04-04
Enable full mainnet UTxO validation

ruhatch:


test/Test/Cardano/Chain/Txp/Validation.hs
```diff
@@ -61,7 +61,7 @@ tests scenario = do
   let
     takeFiles :: [FilePath] -> [FilePath]
     takeFiles = case scenario of
-      ContinuousIntegration -> take 10
+      ContinuousIntegration -> identity
```
dnadales, line 64: Uh? We were not running this on CI? When do we use the `QA` scenario BTW?

dnadales (approved):

test/Test/Cardano/Chain/Txp/Validation.hs
ruhatch: We disabled it because of memory issues, so I imagine we need to merge @intricate's PR to avoid those...

ruhatch: I added QA, for when QA want to run this manually :woman_shrugging:

ruhatch:
For reference, CI took 17 mins with the normal UTxO - I'm going to cherry-pick @intricate's commits here to get an idea of the overhead

test/Test/Cardano/Chain/Txp/Validation.hs
dnadales: > We disabled it because of memory issues, so I imagine we need to merge @intricate's PR to avoid those...

Right, but I thought we enabled it again since the memory issues were solved (I guess the fact that you're able to run CI means this is the case) :)

dnadales:
> For reference, CI took 17 mins with the normal UTxO - I'm going to cherry-pick @intricate's commits here to get an idea of the overhead

Thanks so much for bench-marking this. I'm really curious to see what the effect in the actual memory usage as well. Would it make sense to run a heap profile? If it does and you don't have time to do it I can help.

ruhatch:
The compact data types seem to have had little effect on the run time, which is pretty awesome!

dnadales:
> The compact data types seem to have had little effect on the run time, which is pretty awesome!

:tada:

PR #422, opened 2019-04-04
[#363] Make `withTests` counts configurable

### LINKED ISSUE #363: Make `withTests` counts configurable
## Background
While studying `cardano-chain` as part of my getting-up-to-speed prep, I came across [this comment](https://github.com/input-output-hk/cardano-chain/blob/03cf7207c4961fec6ca22a074759b9cd4014fa06/test/Test/Cardano/Chain/Block/Validation/Spec.hs#L59-L61):
```
  -- TODO: we might want to make this configurable, so that we run a smaller
  -- number of tests when developing, and use a higher number when on CI (or in
  -- nightly builds?).
```
This reminded me of test sluggishness we had back on `cardano-sl`, which I always meant to get around to making configurable by situation (CI / local dev / nightly) but never did.

I wanted to make an issue so we track this and prioritize it appropriately. It likely isn't super-high priority, given that I don't think PR times, or local dev test run times are a bottleneck (but then again, I haven't been developing actively, so I don't know).

## Proposed Solution

(This is very much conditional on if we stick with Hedgehog, versus switching to Quickcheck. I am only going to provide a proposal for Hedgehog now, but the Quickcheck version should look similar)

In most of our test modules we use:
```
tests :: IO Bool
tests = checkParallel $$discover
```
to find a test `Group` and execute it. We then import them all into the top-level `test/test.hs` executable which [runs all of them via `runTests`](https://github.com/input-output-hk/cardano-chain/blob/03cf7207c4961fec6ca22a074759b9cd4014fa06/test/test.hs#L30-L51).

My suggestion is to write a new version of `discover` which will find things of type `Int -> Property` (where the `Int` is the test scalar) and package them into something of type `Int -> Group` (where the argument is threaded to all respective properties). Properties declared like `prop_<...> testScalar` would then use `withTests (3*testScalar` to scale their test counts.

Thus we would have something like:
```
tests :: Int -> IO Bool
tests scalar = checkParallel ($$discoverScaled scalar)
```
and, in the top-level executable:
```
main :: IO ()
main = do
  opts <- execParser optsParser
  let scenario = optsTestScenario opts
  testScalar <- getEnv "TEST_SCALAR"
  runTests $ map ($ testScalar)
    [ Test.Cardano.Chain.Block.Bi.tests
    ...
    ]
```
to thread the test scalar into the test groups.

### Benefit

Developers can run tests locally with a low `TEST_SCALAR`, perhaps 10, to get speedy test results, with probabilistically lower guarantees of correctness (but still "enough", ideally).

CI can run tests with `TEST_SCALAR=100` or so (approximating the current values of test counts). Since CI should be run fewer times than local dev, we get better assurance with a longer wait which happens infrequently.

Nightly tests can be run with `TEST_SCALAR=500` or more, to get high assurance. We get even better assurance, with longer wait, but even more infrequent.

This allows us scale the assurance of correctness we want.

mhuesch:
Closes #363.

Uses functionality from https://github.com/input-output-hk/cardano-prelude/pull/45 to thread `TestScenario` into `Property`s. This then allows us to adjust the number of test runs according to Development, CI, and QA scenarios.

~I create new versions of the two main test helpers, `eachOfTS` and `withTestsTS`, which take a `Data.Ratio` scalar by which to multiply the default number of tests for each respective test scenario. This allows us to run more tests on certain Properties than others, while still having them all scale linearly up and down depending on scenario. With this PR in, I record the different runtimes (times don't include build time):~

~30 sec for a full test-suite execution is quite nice, and could mean around 1min for a full rebuild & rerun of the testsuite for local development!~

^ approach was modified - see discussion below.

- [x] Wait for https://github.com/input-output-hk/cardano-prelude/pull/45 to be merged and bump the hash accordingly in `stack.yaml`


mhuesch:
We can also consider making things even faster by dropping the number of epochs we validate in 2 test modules:
```
diff --git a/test/Test/Cardano/Chain/Block/Validation.hs b/test/Test/Cardano/Chain/Block/Validation.hs
index aebe7b0..e66a477 100644
--- a/test/Test/Cardano/Chain/Block/Validation.hs
+++ b/test/Test/Cardano/Chain/Block/Validation.hs
@@ -80,7 +80,7 @@ tests scenario = do
     takeFiles :: [FilePath] -> [FilePath]
     takeFiles = case scenario of
       ContinuousIntegration -> identity
-      Development           -> take 15
+      Development           -> take 8
       QualityAssurance      -> identity
 
   -- Get a list of epoch files to perform validation on
diff --git a/test/Test/Cardano/Chain/Txp/Validation.hs b/test/Test/Cardano/Chain/Txp/Validation.hs
index 3cee876..d7531e8 100644
--- a/test/Test/Cardano/Chain/Txp/Validation.hs
+++ b/test/Test/Cardano/Chain/Txp/Validation.hs
@@ -62,7 +62,7 @@ tests scenario = do
     takeFiles :: [FilePath] -> [FilePath]
     takeFiles = case scenario of
       ContinuousIntegration -> take 10
-      Development           -> take 15
+      Development           -> take 8
       QualityAssurance      -> identity
 
   -- Get a list of epoch files to perform validation on
```

That would get us down to 21.33sec for `Development` testsuite execution.

test/Test/Cardano/Chain/Block/Bi.hs
```diff
@@ -117,9 +118,9 @@ goldenHeader =
     "test/golden/bi/block/Header"
 
 -- | Round-trip test the backwards compatible header encoding/decoding functions
-roundTripHeaderCompat :: Property
+roundTripHeaderCompat :: TSProperty
 roundTripHeaderCompat =
-  eachOf 10 (feedPMEpochSlots $ genWithEpochSlots genHeader) roundTripsHeaderCompat
+  eachOfTS (1%5) (feedPMEpochSlots $ genWithEpochSlots genHeader) roundTripsHeaderCompat
```
dnadales: Question: I don't get why do we use a rational here, and what is the meaning of `1/5`. 

I'm wondering whether it would make sense to have an integer number here instead, which determines how many tests to run (independently of the scenario), and the adjust the percentage in the scenarios. The rationale for this is the following: consider a property `p` for which we know that we require `1000` tests to achieve a high degree of confidence that it holds, then if we must use a rational number here we must know what is the number of tests that will be run in the QA scenario, and this requires some non-local information, which might change and invalidate a bunch of other properties that rely on it.

So using an integer here and a rational in the scenarios, where QA means "run 100%" of the specified test cases, give us the guarantee that in QA we will be running all the tests that the properties requires to achieve a good level of confidence in their correctness.

dnadales (commented):
Thanks, this is good stuff! I'm just a bit concerned about using a rational instead of an integer (I would invert things), but I don't know what the rest thinks ...

test/Test/Cardano/Chain/Block/Bi.hs
ruhatch: I think I agree with Damian here - what we want to specify is the number of tests we want in CI to be confident and any scaling should be from that.

ruhatch (approved):
Looks neat, but I agree with Damian about what to pass to `eachOfTS` and `withTestsTS`

Jimbo4350 (commented):
I also agree with Damian here: https://github.com/input-output-hk/cardano-ledger/pull/422/files#r272477819
Other than that everything looks good!

test/Test/Cardano/Chain/Block/Bi.hs
mhuesch: Yeah, thanks for pointing this out! I think your proposed solution is much more intuitive.

I started down the `Ratio` path simply because it was my first idea, but it requires more mental "coupling" because the test writer has to remember what the default values are and multiply accordingly. And so does the reader of the tests...

I'll revamp this to use Damian's suggestion.

mhuesch: I am now wondering though - what scenario should we consider the original test numbers to "represent", so that I can scale them appropriately?

Say that a test read `withTests 1000 $ <...>` prior to this PR. Would we leave it at `withTestsTS 1000`, thus considering 1000 to be a good enough assurance level for the QualityAssurance scenario? Or would might we consider the original number ContinuousIntegration, and thus write `withTestsTS 5000` (I am using the crude relation that QA is 5x CI)?

It seems to me like we could argue either case, and I'd like to pin it down before I go around changing all the numbers.

dnadales: > I am now wondering though - what scenario should we consider the original test numbers to "represent", so that I can scale them appropriately?

That's a very good question :thinking: Like you say we could argue either case.

Personally I'd go with interpreting the number of tests in `withTests` as `QA` tests. The reason for this is that when developing properties we (should) try to determine the *minimum* number of tests that provides us with *high* enough confidence that the property holds. So if for a given property this number is `N`, `5*N` seems like an overkill. But also, on the other hand, if `N` is the number that provides us with enough confidence on the validity of the property, it seems `1/5 * N` might be a bit too low. So I'm wondering what's the proportion we should use for properties between `QA` and `CI` :man_shrugging:

mhuesch: @dnadales:

>  So I'm wondering what's the proportion we should use for properties between QA and CI 🤷‍♂️

I'm also unsure about this. I think it's difficult to determine what "high enough" confidence means.

Given that we have been running CI using a fixed number of test, I wonder if we ought to continue using that number, call it `N`, for CI. Then we can run `2*N` for QA (I don't believe we don't currently run QA, so we can set up a nightly job). We can run `N/2` for Dev.

Since we don't have QA set up, I don't want to drop our test numbers by 1/2 in CI with no corresponding nightly coverage. If we set up QA and get that into the workflow later, we can revise the test counts.

Does that sound reasonable? /cc @ruhatch

dnadales: > Does that sound reasonable? /cc @ruhatch

It does to me :) Thanks @mhuesch

mhuesch: Great, I just incorporated the changes!

mhuesch:
I went with Damian's suggestion conceptual (discussion [here](https://github.com/input-output-hk/cardano-ledger/pull/422#discussion_r272477819)), and my suggested solution for test counts - to let the fixed numbers represent CI, and use ratios `1/2`, `1`, and `2` for Dev, CI, QA, respectively.

The changeset is simpler this way, as all the test numbers stay the same, but the types and helper functions change.

Here are the runtimes I observe:

TestScenario | Testsuite runtime
------------ | -------------
Development | 126s
ContinuousIntegration | 369s
QualityAssurance | 515s

***

bors r+

mhuesch:
@ruhatch giving bors a break? I thought this one might succeed, since sometimes the build is cached from a previous timed-out execution of the same exact revision.

Just noticed the branch has gotten out of date as well... will rebase that.

PR #425, opened 2019-04-05
[#90] Move cardano-crypto tests to hedgehog and remove hspec

### LINKED ISSUE #90: Review the `cardano-crypto` tests
There are a number of issues with crypto testing:

- We should aim to replace `hspec` in all tests
- The `binaryTest`s in `Test.Cardano.Crypto.CborSpec` can be dropped as we already use `hedgehog`
- If possible we could replace all `QuickCheck` things with `hedgehog`
- Examine the use of `ArbitraryUnsafe`, maybe it can be removed

ruhatch:
Closes #90


dnadales (approved):
LGTM

PR #426, opened 2019-04-05
Goblins for state transition systems.

ruhatch:
Goblins are structured mutators for a type; structured in that the nature of
that mutation is goverened by their genome. GHC Generics provide a way to map
this genome to a Goblin.

The intention is to try to 'breed' goblins for a state transition system, such
that a goblin is bred to cause specific types of failure in that system.

Add goblin instance for Data.Ratio

Add Goblin for `Maybe`

Allow `Gen` actions for env and state.

Single generator for full context.

This is likely not what we want eventually, which will be to use `ProgressiveGen`.
But this will work for the time-being for testing purposes.

Fix up goblins!

Add Goblin explainer.

Appease the formatting tyrant.

Move goblins to a separate library.


mhuesch:
@ruhatch @nc6 After doing some looking, it seems like Nick's branches (`nc/goblins` & `nc/goblins-wip`) are stale and involve STS code from the `/specs` dir which has been moved to the https://github.com/input-output-hk/cardano-ledger-specs repo.

Is that why this issue was closed? And should I be migrating the work on the branches over to the `cardano-ledger-specs` repo?

mhuesch:
~I've moved the code changes from `nc/goblins` over the cardano-ledger-specs, and have a branch [here](https://github.com/input-output-hk/cardano-ledger-specs/commits/nc%2Bmhuesch/goblins). I am running into a build error, though, which is baffling me. @nc6 could you take a look when you have a chance? `stack build` on the branch should bring it up. Thanks.~

Please disregard ☝️, I figured it out.

PR #431, opened 2019-04-09
Add: Expose the validation state fields.

nc6:
We need these because the validation state is used to integrate with the
protocol.


ruhatch (approved):
LGTM!

Issue #124, opened 2018-11-12
Add validation rule to enforce a maximum size on `Attribute` values


ruhatch:
Right now `Attributes` can be used to bloat blockchain payloads with unnecessary data. People are actually using it to upload images and text, so this is a known exploit. We could add a protocol parameter to `BlockVersionData` (or whatever that gets named).

mdimjasevic:
Was there a discussion with a conclusion that a cap should be put? If so, please add a reference to it.

PR #434, opened 2019-04-09
Add UTxO and Update to mainnet tests

### LINKED ISSUE #164: Integrate chain validation with real update interface
Replace comparison with genesis BlockVersionData with calls through to the ledger's update interface.

ruhatch:
Closes #164 

Also imposes some limits on `Attributes` in `ExtraBlockData`, `ExtraHeaderData`, and `ProposalBody`. Only remaining `Attributes` are in `Tx` and `Address`, but these have been used on the chain. We must investigate what we can cap these at. This is part of #124.


mhuesch (dismissed):
LGTM, though I went fast.

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -271,58 +332,203 @@ updateChainBoundary cvs bvd = do
...
-    blockLength b <= maxBlockSize `orThrowError` ChainValidationBlockTooLarge
+  blockLength b <= maxBlockSize `orThrowError` ChainValidationBlockTooLarge
+
+  -- Validate the block attributes size
+  length attributes == 0 `orThrowError` ChainValidationBlockAttributesTooLarge
 
   -- Validate the delegation payload signature
-  proofDelegation (blockProof b) == hash (const () <$> blockDlgPayload b)
+  proofDelegation (blockProof b)
+    == hashDecoded (blockDlgPayload b)
```
mhuesch, line 371: Does this correspond to `verify vkd [[bCerts b]] (bhDlgSig bh)` in the spec (Figure 20 of blockchain-spec)? I'm wondering how to account for the lack of checking the signature of `vkd`.

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -271,58 +332,203 @@ updateChainBoundary cvs bvd = do
...
-
-  pure $ cvs
-    {cvsDelegationState = delegationState'}
+    DI.updateDelegation delegationEnv delegationState certificates
+      `wrapError` ChainValidationDelegationSchedulingError
+
+  -- Validate the transaction payload proof
+  proofTxp (blockProof b)
+    == recoverTxProof (blockTxPayload b)
+    `orThrowError` ChainValidationUTxOProofError
```
mhuesch, line 382: Same confusion, about where we actually check signatures, as above.

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -271,58 +332,203 @@ updateChainBoundary cvs bvd = do
...
+        protocolMagic
+        (UPI.adoptedProtocolParameters updateState)
+        utxo
+        txs
+      `wrapError` ChainValidationUTxOValidationError
+
+  -- Validate the update payload proof
+  proofUpdate (blockProof b)
+    == hashDecoded (blockUpdatePayload b)
+    `orThrowError` ChainValidationUpdateProofError
```
mhuesch, line 396: Same as above 2 signature check confusions.

src/Cardano/Chain/Delegation/Validation/Scheduling.hs
```diff
@@ -0,0 +1,133 @@
...
+  delegate        = mkStakeholderId $ pskDelegatePk cert
+
+  delegationEpoch = pskOmega cert
+
+  activationSlot  = addSlotNumber (kSlotSecurityParam k) currentSlot
+
+  delegatesThisSlot sd =
+    sdSlot sd == activationSlot && sdDelegator sd == delegator
+
+  delegation = ScheduledDelegation activationSlot delegator delegate
```
mhuesch, line 133: 🎉 nicely written

test/test.hs
```diff
@@ -50,9 +49,8 @@ main = do
     , Test.Cardano.Chain.Txp.CBOR.tests scenario
     , Test.Cardano.Chain.Txp.Compact.tests scenario
     , Test.Cardano.Chain.Txp.Json.tests scenario
-    , Test.Cardano.Chain.Txp.Validation.tests scenario
     , Test.Cardano.Chain.Update.CBOR.tests scenario
     , Test.Cardano.Chain.Update.Json.tests scenario
     , Test.Cardano.Chain.Update.Properties.tests scenario
-    , Test.Cardano.Chain.Block.Validation.Spec.tests scenario
+    -- , Test.Cardano.Chain.Block.Validation.Spec.tests scenario
```
mhuesch, line 55: Intentional?

mhuesch (approved):
LGTM - wrote a few comments and questions

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -337,37 +543,90 @@ updateBlock
...
+
+  BodyState { utxo, updateState, delegationState } <- updateBody bodyEnv bs b
+
+  pure $ cvs
+    { cvsLastSlot     = blockSlot b
+    , cvsPreviousHash = Right $ blockHashAnnotated b
+    , cvsUtxo         = utxo
+    , cvsUpdateState  = updateState
+    , cvsDelegationState = delegationState
+    }
```
mhuesch, line 573: Looks like a match with the spec, although the PBFT rule is omitted. That's because it's still in the works, right?

src/Cardano/Chain/Block/Validation.hs
```diff
@@ -271,58 +332,203 @@ updateChainBoundary cvs bvd = do
...
+  -- Update the update state
+  updateState' <-
+    UPI.registerUpdate updateEnv updateState updateSignal
+      `wrapError` ChainValidationUpdateError
+
+  pure $ BodyState
+    { utxo        = utxo'
+    , updateState = updateState'
+    , delegationState = delegationState'
+    }
```
mhuesch, line 407: Besides the above comments, this seems to match the blockchain-spec 👍

mhuesch (commented):
(repost comments which I edited from my first review, which made the order of events confusing)

src/Cardano/Chain/Block/Validation.hs
ruhatch: Yeah, I think that's a mistake in the spec and I added it to an issue in `cardano-ledger-specs`. It's always been a hash check, rather than a signature.

test/test.hs
ruhatch: Yeah, there are some issues surrounding these tests that I don't want to dive into now

PR #436, opened 2019-04-10
[#145] Split Bi class into ToCBOR and FromCBOR

### LINKED ISSUE #145: Split Bi class into ToCBOR and FromCBOR
There are a few cases when we want asymmetrical instances:

- With annotated values, we don't only want to decode to an annotated value and only want to encode unannotated values. We could implement `Bi` instances for both, but this wouldn't be as clear.

- With validity-tagged values, we only want to encode valid values, and only want to decode to unvalidated values. We could also do this with `Bi`, throwing errors in the other cases, but it makes sense to split `Bi`.

### LINKED ISSUE #435: `goldenTestBi` output is stifled
## Background

`goldenTestBi` has nice functionality for good error messages. However its output seems to be getting stifled somehow:
```
━━━ Test.Cardano.Chain.Slotting.Bi ━━━
  ✓ golden_EpochIndex passed 1 test.
  ✗ golden_FlatSlotId failed after 1 test.

    This failure can be reproduced by running:
    > recheck (Size 0) (Seed 2264516985870913048 455201420688529847) golden_FlatSlotId

  ✓ golden_LocalSlotIndex passed 1 test.
  ✓ golden_EpochSlots passed 1 test.
  ✓ golden_SlotId passed 1 test.
  ✗ 1 failed, 4 succeeded.
```
(test printouts from a run of `stack test cardano-ledger:test:cardano-ledger-test`)

## Proposed Solution

Unsure; requires investigation.

ruhatch:
Closes #145 
Closes #435


dnadales (approved):
Looks good. But I went really fast through it :)

PR #438, opened 2019-04-10
Convert bulk chain validation exe to a package

Jimbo4350:


bcv/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2018 IOHK
```
ruhatch: Different timezones have different years? :wink:

cabal.project
```diff
@@ -1,6 +1,7 @@
 packages:
   crypto crypto/test
   ./     ./test
+  bcv
```
ruhatch: Hmm, this is a bit cryptic - I'd say just go for `validate-mainnet` or something. Then you can have `live` and `bulk` be modes of the same application :man_shrugging:

ruhatch (approved):
Looks good, but slightly concerned about names

PR #441, opened 2019-04-10
adding RTS flag to view stats in EKG view on http://localhost:12788

CodiePP:
was trying out the block validation program. very nice!
adding the flag allows one to view the RTS stats in EKG on http://localhost:12788/


Jimbo4350 (approved):
This is really cool!

PR #442, opened 2019-04-11
[#363] Document TestScenario scaling

mhuesch:
Open to feedback on this - I wanted to capture it somewhere so there's a reference, but references & docs always go stale, so it's possible it's better to just rely on usage (in other test modules) to teach people how to use these helpers.

Another concern is that the links, which are pinned to a specific git revision, will go stale as development progresses and could become misleading. They would require periodic updates to stay fresh.

Relates to #363.


ruhatch (approved):
Leems legit and hopefully we won't mess around with it too much, so keeping docs up to date should be okay.

bors r+

PR #445, opened 2019-04-24
Fix nix

angerman:
Investigating the Win32 failures @ruhatch reported.


nc6 (approved):
@angerman Out of interest, what is the fix being applied here? The only thing added except to generated files is the date - I would have assumed this shouldn't change what's generated? But there are a bunch of resultant changes.

ruhatch (approved):
bors r+

angerman:
> @angerman Out of interest, what is the fix being applied here? The only thing added except to generated files is the date - I would have assumed this shouldn't change what's generated? But there are a bunch of resultant changes.

@nc6 yes, this has been confusing me as well. It seems like there were changes to the project, but the regenerate script was never run. And even more confusingly someone can `cabal2nix` to generate `nix/.stack.nix/bulk-chain-validation.nix`.

You can see that `cardano-ledger-validation-exe` was removed from `cardano-ledger` at some point as well, but the nix expressions were not regenerated.

PR #444, opened 2019-04-23
Log UTxO size directly from validation and update readme

ruhatch:

PR #446, opened 2019-04-24
[#228] T3.1.1 benchmarking - bulk block validation

CodiePP:
this is a first version of measuring the block validation routine with a bracket observing RTS and clock.

the log output is shown on 'stdout'; aggregated statistics in EKG (http://localhost:12788)

ran the program with: `cabal new-run validate-mainnet`


ruhatch (commented):
You're seeing a problem validating because of this code:
```
forM_ (take 10 files) $ \epochFile -> liftIO $ do
    -- Currently `validateEpochFiles` is restricted to the first 5 epochs.
        logDebug logTrace $ pack ("validating epoch file: " ++ epochFile)
    result <- liftIO $ validateEpochFiles (genesisConfig config) initialCVS (take 5 files)
        result <- Log.bracketObserveIO logConf
    liftIO $ putMVar cvsVar result
                                       logTrace
                                       Info
                                       "validate" $
                      validateEpochFile (genesisConfig config) initialCVS epochFile
        putMVar cvsVar result
```
You're using `forM_`, but the `validateEpochFile` action is not stateful. You need to use a `foldM` instead. I've actually changed this code in PR #444 and put logging into the `cardano-ledger` library. I'd suggest you look at that PR, then rebase on it and add the benchmarking inside the `foldM`.

CodiePP:
rebased on #444
the aggregated statistics of the measurements of RTS and clock are sent to EKG (http://localhost:12788)
(look for the calculated mean time between block validations: `cardano.epoch-validation.benchmark.diff.Time-interval.monoclock.basic.mean`; is about 3 seconds on my machine)

ruhatch (approved):
Looks great, thanks :)

bors r+

PR #448, opened 2019-04-24
Remove decision from Vote

ruhatch:


src/Cardano/Chain/Update/Vote.hs
```diff
@@ -292,21 +285,19 @@ type Vote = AVote ()
...
 data AVote a = UnsafeVote
-  { uvKey         :: !PublicKey
+  { voterPK       :: !PublicKey
   -- ^ Public key of stakeholder, who votes
-  , aUvProposalId :: !(Annotated UpId a)
+  , aProposalId   :: !(Annotated UpId a)
   -- ^ Proposal to which this vote applies
-  , aUvDecision   :: !(Annotated Bool a)
+  , decisionBytes :: !a
   -- ^ Approval/rejection bit
```
dnadales: Maybe we need to update this comment, since it seems this no longer conveys any semantic meaning.

src/Cardano/Chain/Update/Vote.hs
```diff
@@ -292,21 +285,19 @@ type Vote = AVote ()
...
 --   Invariant: The signature is valid.
 data AVote a = UnsafeVote
-  { uvKey         :: !PublicKey
+  { voterPK       :: !PublicKey
   -- ^ Public key of stakeholder, who votes
-  , aUvProposalId :: !(Annotated UpId a)
+  , aProposalId   :: !(Annotated UpId a)
   -- ^ Proposal to which this vote applies
-  , aUvDecision   :: !(Annotated Bool a)
+  , decisionBytes :: !a
```
dnadales: It seems we're not using this field anymore, so why do we keep it (I'm pretty sure I'm missing something anyway :))

dnadales: I mean, if this is always `bytesOf True`, why not just using this constant in `recoverSignedBytes`?

dnadales (approved):
LGTM!

Only some questions and requests for comments.

src/Cardano/Chain/Update/Vote.hs
ruhatch: Aha, I suppose we could do this, we'd just have to change the code here to use a `Raw "bytesOfTrue"`. The reason I did it this way was to allow non-canonical encodings of `True` to be accepted, but we can probably disregard that case, and just hardcode the `True` bytes.

dnadales: It was just a question, so I trust your judgement :) I'm fine either way

PR #449, opened 2019-04-25
Make katip accept Win32 > 2.6

angerman:


nc6 (approved):

nix/default.nix
```diff
@@ -33,6 +33,12 @@ let
       # list the packages that require template haskell
       # explicity here.
       iohk-module
+      {
+        # katip has an version bound of Win32 < 2.6; this however
+        # implies that it's incompatible with ghc-8.6 (on windows).
+        # Let's force it to accept out packageset.
+        packages.katip.doExactConfig = true;
```
ruhatch, line 40: Cool :D

ruhatch (approved):
Looks great!

bors r+

nix/default.nix
angerman: That's essentially what `jailbreakCabal` did. Just without the additional complexity of requiring a binary and mutilating the cabal file.

PR #451, opened 2019-04-25
Implement hedgehog state machine models for SDELEG, UTXOW, and CHAIN systems

ruhatch:
This is an initial PR to show a proof-of-concept implementation of the state machine testing of various state transition systems. In the end we would like to add tests for DELEG and various parts of the update system, as well as improving test coverage by including invalid chains in the generation.

### DIFF
--- a/src/Cardano/Chain/ProtocolConstants.hs
+++ b/src/Cardano/Chain/ProtocolConstants.hs
@@ -13,19 +13,20 @@ import Cardano.Prelude
 
 import Cardano.Chain.Common.BlockCount (BlockCount(..))
 import Cardano.Chain.Slotting.EpochSlots (EpochSlots(..))
+import Cardano.Chain.Slotting.SlotCount (SlotCount(..))
 
 
 -- | Security parameter expressed in number of slots. It uses chain quality
 --   property. It's basically @blkSecurityParam / chainQualityThreshold@.
-kSlotSecurityParam :: BlockCount -> EpochSlots
-kSlotSecurityParam = EpochSlots . (*) 2 . unBlockCount
+kSlotSecurityParam :: BlockCount -> SlotCount
+kSlotSecurityParam = SlotCount . (*) 2 . unBlockCount
 
 -- | Minimal chain quality (number of blocks divided by number of
 --   slots) necessary for security of the system.
 kChainQualityThreshold :: Fractional f => BlockCount -> f
-kChainQualityThreshold k = realToFrac (unBlockCount k) / fromIntegral (unEpochSlots $ kSlotSecurityParam k)
+kChainQualityThreshold k = realToFrac (unBlockCount k)
+  / fromIntegral (unSlotCount $ kSlotSecurityParam k)
 
 -- | Number of slots inside one epoch
 kEpochSlots :: BlockCount -> EpochSlots
 kEpochSlots = EpochSlots . (*) 10 . unBlockCount
-
--- a/src/Cardano/Chain/Txp/GenesisUTxO.hs
+++ b/src/Cardano/Chain/Txp/GenesisUTxO.hs
@@ -5,40 +5,30 @@ where
 
 import Cardano.Prelude
 
-import Data.Coerce (coerce)
 import qualified Data.Map.Strict as M
 
 import Cardano.Chain.Common (Address, Lovelace, makeRedeemAddress)
 import Cardano.Chain.Common.NetworkMagic (NetworkMagic, makeNetworkMagic)
-import Cardano.Chain.Genesis
-  (GenesisData(..), unGenesisAvvmBalances, unGenesisNonAvvmBalances)
+import Cardano.Chain.Genesis (unGenesisAvvmBalances, unGenesisNonAvvmBalances)
 import qualified Cardano.Chain.Genesis as Genesis
-import Cardano.Chain.Txp.Tx (TxIn(..), TxOut(..))
 import Cardano.Chain.Txp.UTxO (UTxO)
 import qualified Cardano.Chain.Txp.UTxO as UTxO
-import Cardano.Crypto (hash)
 
 
+-- | Create initial 'UTxO' from balances defined in the genesis config
 genesisUtxo :: Genesis.Config -> UTxO
-genesisUtxo genesisConfig = UTxO.fromList $ utxoEntry <$> preUtxo
+genesisUtxo config = UTxO.fromBalances balances
  where
-  genesisData :: GenesisData
-  genesisData = Genesis.configGenesisData genesisConfig
-
-  networkMagic :: NetworkMagic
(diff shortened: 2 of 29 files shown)


src/Cardano/Chain/Txp/GenesisUTxO.hs
```diff
@@ -5,40 +5,30 @@ where
...
-  (GenesisData(..), unGenesisAvvmBalances, unGenesisNonAvvmBalances)
+import Cardano.Chain.Genesis (unGenesisAvvmBalances, unGenesisNonAvvmBalances)
 import qualified Cardano.Chain.Genesis as Genesis
-import Cardano.Chain.Txp.Tx (TxIn(..), TxOut(..))
 import Cardano.Chain.Txp.UTxO (UTxO)
 import qualified Cardano.Chain.Txp.UTxO as UTxO
-import Cardano.Crypto (hash)
 
 
+-- | Create initial 'UTxO' from balances defined in the genesis block
```
nc6:
```suggestion
-- | Create initial 'UTxO' from balances defined in the genesis config
```

The term 'genesis block' has caused confusion multiple times before, so I suggest we avoid it entirely

test/Test/Cardano/Chain/Block/Model.hs
```diff
@@ -0,0 +1,147 @@
...
+import qualified Ledger.Delegation as Abstract
+
+import Test.Cardano.Chain.Elaboration.Block
+
+
+--------------------------------------------------------------------------------
+-- CHAIN
+--------------------------------------------------------------------------------
+
+prop_commandCHAIN :: Property
```
nc6, line 45: Maybe describe what this property is actually testing?

nc6 (approved):

src/Cardano/Chain/ProtocolConstants.hs
```diff
@@ -13,19 +13,20 @@ import Cardano.Prelude
...
 import Cardano.Chain.Common.BlockCount (BlockCount(..))
 import Cardano.Chain.Slotting.EpochSlots (EpochSlots(..))
+import Cardano.Chain.Slotting.SlotCount (SlotCount(..))
 
 
 -- | Security parameter expressed in number of slots. It uses chain quality
 --   property. It's basically @blkSecurityParam / chainQualityThreshold@.
-kSlotSecurityParam :: BlockCount -> EpochSlots
-kSlotSecurityParam = EpochSlots . (*) 2 . unBlockCount
+kSlotSecurityParam :: BlockCount -> SlotCount
```
dnadales, line 21: Uh, didn't we change `SlotCount` to `EpochSlots` ?

test/Test/Cardano/Chain/Block/Model.hs
dnadales: Isn't that described in the comment above?

test/Test/Cardano/Chain/Txp/Model.hs
```diff
@@ -0,0 +1,303 @@
...
+modelAddresses :: [Abstract.Addr]
+modelAddresses = Abstract.Addr . Abstract.VKey . Abstract.Owner <$> [0 .. 5]
+
+
+--------------------------------------------------------------------------------
+-- Main Test Export
+--------------------------------------------------------------------------------
+
+tests :: IO Bool
+tests = checkSequential $$discover
```
dnadales, line 303: Shouldn't we be able to check in parallel?

test/Test/Cardano/Chain/Txp/Model.hs
```diff
@@ -0,0 +1,303 @@
...
+      , crAppliedTx = Concrete.taTx concreteTxAux
+      }
+
+  callbacks :: [Callback SignalUTXOW ConcreteResult StateUTXOW]
+  callbacks =
+    [ Update update
+    , Ensure $ \_ StateUTXOW { lastResult } _ concreteResult -> do
+      annotateShow lastResult
+      isRight lastResult === isRight (crResult concreteResult)
+    , Require post
```
dnadales: Require is a pre-condition. Should `post` be renamed to `pre`?

test/Test/Cardano/Chain/Txp/Model.hs
```diff
@@ -0,0 +1,303 @@
...
+        }
+
+      result =
+        STS.applySTS @(UTXOW v) (STS.TRC (env, abstractState, txWits'))
+    in StateUTXOW (fromRight abstractState result) result
+
+  post :: StateUTXOW Symbolic -> SignalUTXOW Symbolic -> Bool
+  post StateUTXOW { abstractState } (SignalUTXOW (Abstract.TxWits tx _)) =
+    all (`elem` M.keys m) (Abstract.txins tx)
+      && Abstract.balance (Abstract.txouts tx)
```
dnadales, line 240: Isn't this being checked by the UTxO rule? Why do we need to check this here?

dnadales (approved):
Thanks for the great work Ru.

Some considerations:
- there are some repeating patterns regarding testing STS's with hedgehog-SM that we should be able to factor out later on.
- I'm a bit worried by the lack of symmetry between the testing of delegation and chain, and UTxO. In the former we do not make use of symbolic/concrete variables, however in the latter we had to parametrize `UTXOW`. I don't know whether we could reconcile the two approaches.
- I'm also worried (that's what I do apparently) by the additional complexity introduced (specially w.r.t. the use of traces). But since we do not have a better alternative at the moment I guess I'll have to live with it.

But all in all it looks good! Thanks again.

src/Cardano/Chain/ProtocolConstants.hs
ruhatch: Only where that makes sense. In this case this isn't a number of slots in an epoch, but instead just a number of slots.

test/Test/Cardano/Chain/Txp/Model.hs
ruhatch: There's only one property, so shouldn't make a difference!

test/Test/Cardano/Chain/Txp/Model.hs
ruhatch: Yep!

test/Test/Cardano/Chain/Txp/Model.hs
ruhatch: This is for shrinking. If you don't have this here then you can shrink to invalid transactions :(

mgudemann (approved):

test/Test/Cardano/Chain/Block/Model.hs
ruhatch: Hehe, I updated this before Damian saw and confused everyone

ruhatch:
Thanks for the reviews!

bors r+

PR #452, opened 2019-04-25
[#447] Perform checks for ApplicationName and SystemTag

### LINKED ISSUE #447: Perform checks for `ApplicationName` and `SystemTag` in update validation
## Background

The `checkApplicationName` and `checkSystemTag` functions perform some basic sanity checks on the respective types, but we're not currently using them in the update mechanism validation rules.

## Proposed Solution

Use those functions or inline their definitions into the update validation rules.

## Next Action
- [ ] Perform validation checks in the update rules

### LINKED ISSUE #453: Move ProposalBody from Vote to Proposal
## Background

`ProposalBody` is defined in `module Cardano.Chain.Update.Vote`. This does not look like the most suitable place for it.

## Proposed Solution

Move `ProposalBody` and related definitions to a module named `Cardano.Chain.Update.Proposal` or something similar.

ruhatch:
Closes #447
Closes #453


src/Cardano/Chain/Block/ExtraBodyData.hs
```diff
@@ -25,7 +25,7 @@ newtype ExtraBodyData = ExtraBodyData
 
 instance B.Buildable ExtraBodyData where
   build (ExtraBodyData attrs)
-    | areAttributesKnown attrs = "no extra data"
+    | attributesAreKnown attrs = "no extra data"
```
dnadales, line 28: Thanks! :)

src/Cardano/Chain/Update/Validation/Registration.hs
```diff
@@ -110,20 +111,20 @@ registerProposal
...
   verifySignatureDecoded
       protocolMagic
       SignUSProposal
-      proposerPK
-      (recoverProposalSignedBytes body)
-      sig
-    `orThrowError` RegistrationInvalidSignature
+      issuer
+      (recoverProposalSignedBytes aBody)
+      signature
```
dnadales, line 126: Minor thing, I saw that in some other PR's we changed `signature` to `sig`. We might wat to be consistent :man_shrugging:

src/Cardano/Chain/Update/Validation/Registration.hs
```diff
@@ -177,8 +180,8 @@ registerProposalComponents adoptedPV adoptedPP appVersions rs proposal = do
 
   pure $ State registeredPUPs' registeredSUPs'
  where
-  ProposalBody protocolVersion ppu softwareVersion _ _ =
-    proposalBody proposal
+  ProposalBody protocolVersion ppu softwareVersion metadata _ =
+    Proposal.body proposal
```
dnadales, line 184: Nice!

dnadales (approved):
:heavy_check_mark:

src/Cardano/Chain/Update/Validation/Registration.hs
ruhatch: So, this is to do with shadowing and `NamedFieldPuns`. When I pull something out of a record and it's called `signature` I keep it as `signature` and if it's a simple function parameter that might shadow one of those record accessors then I call it `sig`. Really in those cases we should be using the record accessors qualified to avoid such a clash.

PR #456, opened 2019-04-29
[#124] Add attribute size bounds, as per cardano-sl

### LINKED ISSUE #124: Add validation rule to enforce a maximum size on `Attribute` values

ruhatch:
Closes #124


mhuesch (approved):
LGTM!

mhuesch:
Looks like a merge conflict needs to be resolved though.

PR #463, opened 2019-04-30
Clarify UTxO size in validate-mainnet printouts

mhuesch:
Relates to #444.

A minor tweak I made while watching the demo video, which I thought would be helpful for people reproducing the demo in the future.


ruhatch (approved):
Looks great, thanks! Need to follow up with the logging team about how to display these with newlines preserved.

Issue #462, opened 2019-04-29
Need to use addressHashDecoded to remove canonicity requirement


ruhatch:
## Background

We have mostly removed the requirement for canonical CBOR encodings from the codebase, but when comparing hashes of `VerificationKey`s during validation, we're still using `addressHash`, which uses `serialize` under the hood.

## Proposed Solution

Add `addressHashDecoded` and replace uses of `addressHash` in validation code. This will also affect `mkStakeholderId`, so we should replace that as well.

## Next Action
- [ ] ~Add `addressHashDecoded` and replace uses of `addressHash` in validation code~

ruhatch:
It turns out that if we use `addressHashDecoded`, then two different nodes could hash the same `Address'` or `PublicKey` to different hashes. While this doesn't actually break the ledger rules, it would be a strange experience for the user. The same `Address'` could end up with two different `Address`es, with different `addressRoot`s and therefore a separate UTxO set. It therefore makes sense to enforce canonical encodings for these type that get hashed in this way.

## Next Action
- [ ] Enforce canonical encodings for `Address'` and `PublicKey`

PR #464, opened 2019-05-02
[#462] Enforce canonical encodings for types that are addressHashed

ruhatch:


dcoutts (approved):
I've not checked that this is comprehensive, that it catches every case, but the cases that are changed here look right.

PR #466, opened 2019-05-02
[#352] Delegation map should be injective

### LINKED ISSUE #352: Delegation map should be injective
Update the implementation of the delegation to restrict the delegation map to an injective one.

The current rule for delegation activation states:

image

whereas the check that there isn't a pre-image for `vk_d` seems to be absent from the current implementation:

```haskell

-- | Activate a 'ScheduledDelegation' if its activation slot is less than the
--   previous delegation slot for this delegate, otherwise discard it. This is
--   an implementation of the delegation activation rule in the ledger
--   specification.
activateDelegation :: ActivationState -> ScheduledDelegation -> ActivationState
activateDelegation as delegation
  | prevDelegationSlot < slot || slot == 0 = ActivationState
    { asDelegationMap   = M.insert delegator delegate delegationMap
    , asDelegationSlots = M.insert delegator slot delegationSlots
    }
  | otherwise = as
 where
  ActivationState delegationMap delegationSlots = as
  ScheduledDelegation slot delegator delegate   = delegation

  prevDelegationSlot =
    fromMaybe 0 $ M.lookup delegator (asDelegationSlots as)
```

We should also consider using a more efficient structure than `Map` since we need to lookup the pre-image  of a key. Maybe we can use a `Bimap` or something similar.

ruhatch:
Closes #352

Don't merge until I get `Bimap` PR included upstream


intricate:
I believe `Bimap` is implemented using `Data.Map` (lazy `Map`s). In an effort not to undermine the work done in PR #402, should we raise a PR upstream adding a strict `Bimap` or something?

ruhatch:
@intricate Looking at the implementation there are `!`s on the fields of `Bimap`, but I'm not sure if that would make sure they're strict. It might be worth making that explicit by using `Data.Map.Strict`, yeah.

dnadales (approved):
LGTM. However I wonder whether it makes sense to add an abstraction for `Map` inside delegation. I would have just exposed the use of `Bimap`.

ruhatch:
@dnadales, yeah I'm not entirely sure, but I'll leave it like this for now

PR #468, opened 2019-05-02
[#430] Unify calculation of the confirmation and adoption threshold

### LINKED ISSUE #430: Unify the calculation of the confirmation and adoption threshold
## Background

Currently the update proposal confirmation threshold is hardcoded:

```haskell
module Cardano.Chain.Update.Validation.Voting

-- | ...

    -- | This is the number of genesis keys that need to support a proposal
    threshold :: Int
    threshold = 5
```

Whereas the adoption threshold is calculated as follows:

```haskell
module Cardano.Chain.Update.Validation.Endorsement

-- | ...

  -- In Byron we do not have a @upAdptThd@ protocol parameter, so we have to
  -- use the existing ones.
  --
  -- @lovelacePortionToDouble . srMinThd . ppSoftforkRule@ will give us the
  -- ratio (in the interval @[0, 1]@) of the total stake that has to endorse a
  -- protocol version to become adopted. In genesis configuration, this ratio
  -- will evaluate to @0.6@, so if we have 7 genesis keys, @upAdptThd = 4@.
  upAdptThd :: Int
  upAdptThd = floor $ stakeRatio * fromIntegral ngk
   where
    stakeRatio = lovelacePortionToDouble . srMinThd . ppSoftforkRule $ pps
```

So when the number of genesis keys equals 7, `upAdptThd` will evaluate to `4`. So it seems we could take `ceiling` instead, and unify and abstract away the calculation of both thresholds.

## Proposed Solution

- Remove the hardcoded threshold in `Cardano.Chain.Update.Validation.Voting`.
- Abstract away `upAdptThd` and use `ceiling instead of `floor`
- Use the same threshold for endorsement and for adoption.

ruhatch:
Closes #430

### DIFF
--- a/src/Cardano/Chain/Update/ProtocolParameters.hs
+++ b/src/Cardano/Chain/Update/ProtocolParameters.hs
@@ -11,6 +11,7 @@
 
 module Cardano.Chain.Update.ProtocolParameters
   ( ProtocolParameters(..)
+  , upAdptThd
   , isBootstrapEraPP
   )
 where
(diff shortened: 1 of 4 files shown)


dnadales (approved):
Nice! Thanks :pray:

PR #469, opened 2019-05-03
Make it build with cabal new-build

erikd:


CodiePP:
thanks. works again for me, too. just ran `validate-mainnet`.

nc6 (approved):

PR #470, opened 2019-05-03
decoded boundary blocks have epoch index

avieth:
In the Byron proxy we store boundary blocks on disk. To open the `ImmutableDB` we need to give a parser which can figure out the slot number from genesis for a serialized block. In order to switch from the old `cardano-sl` block decoder to those from this package, I need to be able to decode the epoch index from a boundary block. I don't have a deep understanding of this repository but AFAICT this is a nice improvement: we don't need to give the `isEpochZero` parameter anymore.

### DIFF
--- a/src/Cardano/Chain/Block/Boundary.hs
+++ b/src/Cardano/Chain/Block/Boundary.hs
@@ -4,14 +4,18 @@
 
 module Cardano.Chain.Block.Boundary
   ( dropBoundaryConsensusData
+  , dropBoundaryConsensusDataRetainEpochIndex
   , dropBoundaryExtraHeaderData
   , dropBoundaryBody
   , dropBoundaryExtraBodyData
   )
 where
 
+import Control.Monad (return, void)
+import Data.Word (Word64)
+
 import Cardano.Binary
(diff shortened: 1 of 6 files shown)


src/Cardano/Chain/Block/Boundary.hs
```diff
@@ -4,14 +4,19 @@
...
   ( dropBoundaryConsensusData
+  , dropBoundaryConsensusDataRetainEpochIndex
   , dropBoundaryExtraHeaderData
   , dropBoundaryBody
   , dropBoundaryExtraBodyData
   )
 where
 
+import Control.Monad (return, void)
+import qualified Codec.CBOR.Decoding as D (Decoder, decodeWord64)
```
ruhatch: We export these types from `Cardano.Binary`, so you should be able to import them there and remove the explicit `cborg` dependency :)

ruhatch (approved):
Looks, great, just a small suggestion to remove the `cborg` dependency (which should also fix CI hopefully)

ruhatch:
Ah, looks like there are other issues in CI actually, but I'm happy to fix this up!

avieth:
Thanks @ruhatch

PR #472, opened 2019-05-06
Log UTxO-related stats at debug level during epoch validation

intricate:
Probably a good idea to only calculate and log `UTxO`-related stats at the `Debug` level.

### DIFF
--- a/configuration/log-configuration.yaml
+++ b/configuration/log-configuration.yaml
@@ -60,6 +60,10 @@ options:
         - - tag: Contains
             contents: diff.RTS.gcNum.timed.
       subtrace: FilterTrace
+    'cardano.epoch-validation.utxo-stats':
+      # Change the `subtrace` value to `Neutral` in order to log
+      # `UTxO`-related messages during epoch validation.
+      subtrace: NoTrace
   mapBackends:
     cardano.epoch-validation.benchmark:
       - AggregationBK
--- a/src/Cardano/Chain/Epoch/Validation.hs
+++ b/src/Cardano/Chain/Epoch/Validation.hs
@@ -20,7 +20,7 @@ import qualified Streaming.Prelude as S
 import qualified Cardano.BM.Configuration as Log
 import Cardano.BM.Data.Severity as Log
 import Cardano.BM.Observer.Monadic as BM
-import Cardano.BM.Trace (Trace, appendName, logNotice)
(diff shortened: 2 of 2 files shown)


configuration/log-configuration.yaml
```diff
@@ -1,5 +1,5 @@
 # global filter; messages must have at least this severity to pass:
-minSeverity: Debug
+minSeverity: Info
```
intricate: Instead of doing this, as @CodiePP has pointed out to me, I could also specifically set the `SubTrace` value for the `cardano.epoch-validation.utxo-stats` logger to `NoTrace`. 

Would this be preferable to changing the global `minSeverity`?

intricate: In the aforementioned case, we'd just have to change the `SubTrace` value for `cardano.epoch-validation.utxo-stats` to `Neutral` in order to resume the logging of `UTxO` -related values again. Can do this by either manually editing the log configuration file or via the logging configuration web editor (default URL: http://localhost:12787).

src/Cardano/Chain/Epoch/Validation.hs
```diff
@@ -61,33 +61,48 @@ validateEpochFile
...
-      utxoSize
-      heapSize
-    )
+  logResult
+    :: Trace m Text
+    -> Trace m Text
+    -> ChainValidationState
+    -> m ChainValidationState
+  logResult trace' utxoTrace cvs' =
+    cvs'
```
ruhatch: I think `do`-notation might be cleaner here

src/Cardano/Chain/Epoch/Validation.hs
```diff
@@ -61,33 +61,48 @@ validateEpochFile
...
+            )
+      <* logDebug
+            utxoTrace
+            (sformat
+              utxoStatsFormat
+              utxoSize
+              heapSize
+            )
    where
     sizePair = calcUTxOSize (cvsUtxo cvs')
```
ruhatch: You can just pattern-match, rather than using `fst` and `snd`!

configuration/log-configuration.yaml
ruhatch: Yeah, I think it would be better to leave this as is and put `NoTrace` on the UTxO value - you should add some documentation somewhere about how to change this value too!

ruhatch (changes requested):

ruhatch (approved):
Looks great :)

PR #473, opened 2019-05-07
Update cardano-mainnet-mirror

erikd:
This allows setting the location of the cardano-mainnet-mirror using
an environment variable.


ruhatch (approved):
LGTM!

bors r+

PR #474, opened 2019-05-07
Changes needed for consensus integration

nc6:
This PR incorporates the various changes needed for integration with the consensus layer. They can be divided into three basic ideas:

- Firstly, we expose a bunch of things which were previously hidden, because they're needed.
- Secondly, we change some of the serialisation code to take a more explicit serialiser, rather than using `ToCBOR` constraints which the consensus layer can typically not satisfy.
- Finally, we split some of the rule processing to fit with an explicit header/body split.

### DIFF
--- a/crypto/test/Test/Cardano/Crypto/Signing/Signing.hs
+++ b/crypto/test/Test/Cardano/Crypto/Signing/Signing.hs
@@ -9,6 +9,7 @@ import Hedgehog
 import qualified Hedgehog.Gen as Gen
 import qualified Hedgehog.Range as Range
 
+import Cardano.Binary (toCBOR)
 import Cardano.Crypto.Signing (SignTag(..), sign, toPublic, verifySignature)
 
 import qualified Test.Cardano.Crypto.Dummy as Dummy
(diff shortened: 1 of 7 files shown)


crypto/test/Test/Cardano/Crypto/Signing/Signing.hs
```diff
@@ -35,8 +36,8 @@ prop_sign = property $ do
   a        <- forAll genData
 
   assert
-    $ verifySignature Dummy.protocolMagicId SignForTestingOnly pk a
-    $ sign Dummy.protocolMagicId SignForTestingOnly sk a
+    $ verifySignature toCBOR Dummy.dummyProtocolMagicId SignForTestingOnly pk a
```
ruhatch, line 39: Should be just `protocolMagicId` now

crypto/test/Test/Cardano/Crypto/Signing/Signing.hs
```diff
@@ -35,8 +36,8 @@ prop_sign = property $ do
...
+    $ sign dummyProtocolMagicId SignForTestingOnly sk a
```
ruhatch, line 40: This should be `Dummy.protocolMagicId`

ruhatch (approved):
LGTM! Is the `HeaderState` stuff still good given the header/body split issues?

nc6:
@ruhatch To answer your comment, no! I'm probably going to drop the last commit. So this PR isn't ready for merge yet...

nc6:
Closing in favour of #500

PR #476, opened 2019-05-07
[#475] Use nix to cache cardano-mainnet-mirror and use env var

### LINKED ISSUE #475: Use nix to cache cardano-mainnet-mirror and use env var
## Background

`cardano-mainnet-mirror` being a Haskell package means that `stack` downloads it each time we run CI, even if it hasn't changed. 

## Proposed Solution

Move `cardano-mainnet-mirror` into a `nix` derivation that pulls the files, and put the output into an env var. We then move the Haskell code in `cardano-mainnet-mirror` into `cardano-ledger`.

## Next Action
- [ ] Create `cardano-mainnet-mirror.nix` to pull git repo into derivation
- [ ] Put output into env var accessible to CI (maybe stack-shell.nix?)
- [ ] Remove `cardano-mainnet-mirror` Haskell dependency

ruhatch:
Closes #475


erikd (approved):
I know nothing of Nix, but anyway ....

PR #482, opened 2019-05-09
Remove parseEpochFile/parseEpochFiles

erikd:
These functions parsed epochs ignoring the epoch boundary blocks,
but we should always parse the EBBs, so these can be dropped.


ruhatch (approved):
LGTM!

bors r+

PR #483, opened 2019-05-09
[#181] Change Public and SecretKey to Verification and SigningKey

ruhatch:
This renames `PublicKey` to `VerificationKey` and `SecretKey` to `SigningKey`. It also updates phrases like `public key`, variable names with `Pk` or `PK` to `VK`, and `toPublic` to `toVerification`.

I probably should have split this up into multiple PRs, but there isn't any meat to it at all, so hopefully not too bad to review!


kantp (approved):
Thanks, @ruhatch!

intricate (approved):

PR #487, opened 2019-05-09
Update logging

CodiePP:
this update brings
* textual logs on the console
* usage of logging functions only as exported via node-shell
* removed direct import of iohk-monitoring modules


ruhatch (approved):
Looks good, but there's some cross-compilation issue in `cardano-shell` that you might need to fix upstream.

PR #489, opened 2019-05-09
Assert that `ChainValidationState` is in normal form in `epochValid` prop

intricate:
I've added an option to `Test.Cardano.Chain.Block.Validation.tests` which either enables or disables an assertion to ensure that the `ChainValidationState` is always in normal form after validating each epoch. 

- In our existing `cardano-ledger-test` test-suite, we specify **not** to assert that the `ChainValidationState` is in normal form (`NoAssertNF`).
- Added a new test executable, `epoch-validation-normal-form-test`, in which we specify to assert that the `ChainValidationState` is in normal form (`AssertNF`). I've also added a new CI job which builds and executes this without `hpc` coverage.

The new test executable and CI job was added as a means of circumventing an "issue" which was encountered in our existing CI job. Because our existing CI job builds with `hpc` (`ghc -fhpc`/`stack --coverage`), thunks are introduced throughout our program for `hpc`'s program coverage measurement purposes and this prevents us from accurately determining whether a given expression is in normal form. 

So, following @erikd's advice, I've introduced this new test executable (which will be built without `hpc`) which basically just calls upon `Test.Cardano.Chain.Block.Validation.tests` and provides it with an argument of `AssertNF`. This way, we can still build and test our normal test-suite with code coverage and also test this new normal form assertion in a separate CI job.

### Notes

The function [`heap_view_closurePtrs`](https://gitlab.haskell.org/ghc/ghc/blob/3bdf0d01ff47977830ada30ce85f174098486e23/rts/Heap.c#L79) in [`ghc/rts/Heap.c`](https://gitlab.haskell.org/ghc/ghc/blob/3bdf0d01ff47977830ada30ce85f174098486e23/rts/Heap.c) currently does not handle closures of type `CONSTR_NOCAF` and, as a result, it `fprintf`s a message to `stderr` every time it encounters one: `"closurePtrs: Cannot handle type CONSTR_NOCAF yet"`. 

Since the `isNormalForm` function utilizes this function it could get a bit "spammy" if we run into `CONSTR_NOCAF`s when running the test-suite. However, I've only seen that `CompactTxInUtxo`s are allocated as `CONSTR_NOCAF`s when we compile with `-O2`. In our CI jobs, we tend to use `stack --fast` (`-O0`) so we don't see this sort of spammy output to `stderr` in the logs.

It's also worth noting that [Joachim Breitner has recently raised a PR with `ghc` for specifically handling `CONSTR_NOCAF` objects in `heap_view_closurePtrs`](https://gitlab.haskell.org/ghc/ghc/merge_requests/867)

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,9 +1,6 @@
 steps:
   - label: 'stack rebuild'
     env:
-      AWS_REGION: us-west-1
-      S3_BUCKET: appveyor-ci-cache
-      CACHE_S3_MAX_SIZE: 2500MB
       STACK_ROOT: "/build/cardano-ledger.stack"
     command:
       # cache-s3 needs a build directory that is the same across all buildkite agents.
@@ -15,13 +12,6 @@ steps:
       - "./stack-rebuild"
     agents:
       system: x86_64-linux
-
-  # - label: 'brittany'
-  #   command:
-  #     - "nix-build scripts/brittany -o check-brittany"
-  #     - "./check-brittany"
-  #   agents:
-  #     system: x86_64-linux
 
   - label: 'nix-tools'
     command: 'scripts/buildkite/check-nix-tools.sh'
--- a/test/NormalFormTest.hs
+++ b/test/NormalFormTest.hs
@@ -1,0 +1,31 @@
+{-|
+This module basically just runs the mainnet epoch validation tests from
+"Test.Cardano.Chain.Block.Validation" but provided a 'ShouldAssertNF' value of
+'AssertNF'.
+
+We've created a separate test executable for this as, in our typical CI jobs,
+we utilize @hpc@ (i.e. building with @ghc -fhpc@ or @stack --coverage@) for
+providing code coverage. @hpc@ appears to introduce thunks around our Haskell
+expressions for its program coverage measurement purposes which prevents us
+from accurately determining whether a given expression is in normal form. As a
+result, we have another CI job which will build and run this test executable
+without @hpc@.
+-}
+
+module Main
+  ( main
+  )
+where
+
+import Cardano.Prelude
+
+import Test.Options (ShouldAssertNF (..), mainWithTestScenario, tsGroupToTree)
+
+import qualified Test.Cardano.Chain.Block.Validation
+
+
+main :: IO ()
+main =
+  mainWithTestScenario
+    . tsGroupToTree
+    $ Test.Cardano.Chain.Block.Validation.tests AssertNF
--- a/.stack-to-nix.cache
+++ b/.stack-to-nix.cache
@@ -26,3 +26,6 @@ https://github.com/input-output-hk/cardano-ledger-specs ad86dcb64e8a5f3bd7670655
 https://github.com/input-output-hk/cardano-ledger-specs ad86dcb64e8a5f3bd7670655ab92b50921946479 byron/ledger/executable-spec 1p21vbqdpi3yxzvlmskcrlqlpkszmnif7i9rxgp7xnas6fyj079y cs-ledger cs-ledger.nix
 https://github.com/input-output-hk/cardano-ledger-specs ad86dcb64e8a5f3bd7670655ab92b50921946479 byron/chain/executable-spec 1p21vbqdpi3yxzvlmskcrlqlpkszmnif7i9rxgp7xnas6fyj079y cs-blockchain cs-blockchain.nix
 https://github.com/input-output-hk/cardano-shell de45bf135fc0c575753613ee157378a28f9261e2 . 0j3imc9c80904ybhbrl4mb0v6h9j3c7c9l6pij6idg3adh0rla9z cardano-shell cardano-shell.nix
+https://github.com/input-output-hk/cardano-prelude cca3c4f8da1de19321e96a5a7cca376bc6c37637 . 087yby0q18xk9i2f50a4gjz9rdn1jgwwp8yi3qv5b2nbwb2pzvn7 cardano-prelude cardano-prelude.nix
+https://github.com/input-output-hk/cardano-prelude cca3c4f8da1de19321e96a5a7cca376bc6c37637 test 087yby0q18xk9i2f50a4gjz9rdn1jgwwp8yi3qv5b2nbwb2pzvn7 cardano-prelude-test cardano-prelude-test.nix
+http://github.com/well-typed/canonical-json ddfe3593b80b5ceb88842bb7a6f2268df75d2c2f . 02fzn1xskis1lc1pkz0j92v6ipd89ww0k2p3dvwpm3yap5dpnm7k canonical-json canonical-json.nix
--- a/cabal.project
+++ b/cabal.project
@@ -18,12 +18,12 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-prelude
-  tag: a136c4242b9c9f6124b811329bc8ccdfd86c514e
+  tag: cca3c4f8da1de19321e96a5a7cca376bc6c37637
 
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-prelude
-  tag: a136c4242b9c9f6124b811329bc8ccdfd86c514e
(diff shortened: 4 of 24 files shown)


test/Test/Cardano/Chain/Block/Validation.hs
```diff
@@ -113,6 +113,8 @@ epochValid config cvsRef fp = withTests 1 . property $ do
   result <- (liftIO . runResourceT . runExceptT)
     (foldChainValidationState config cvs stream)
   newCvs <- evalEither result
+  newCvsIsNF <- liftIO $ isNormalForm $! newCvs
```
dnadales: Nice idea :) Let's hope the GHC bug gets resolved soon :pray:

ruhatch (dismissed):
Looks good, but the hydra build seems to have taken 2 hours and then timed out, which is a little odd

test/Test/Options.hs
```diff
@@ -129,3 +130,12 @@ withTestsTS
   -> Property
 withTestsTS count prop scenario =
   withTests (scenarioScaled count scenario) prop
+
+--------------------------------------------------------------------------------
+-- ShouldAssertNF
+--------------------------------------------------------------------------------
+
+data ShouldAssertNF
```
intricate: I wonder whether this is an appropriate place to define this data type.

cardano-ledger.cabal
```diff
@@ -260,3 +260,42 @@ test-suite cardano-ledger-test
 
   if (!flag(development))
     ghc-options:         -Werror
+
+executable epoch-validation-normal-form-test
```
intricate: Perhaps I should think of a better name for this `executable`.

.buildkite/pipeline.yml
```diff
@@ -16,6 +16,24 @@ steps:
     agents:
       system: x86_64-linux
 
+  - label: 'epoch validation normal form test'
```
intricate: Perhaps I should come up with a better label.

.buildkite/pipeline.yml
```diff
@@ -16,6 +16,24 @@ steps:
...
+      S3_BUCKET: appveyor-ci-cache
+      CACHE_S3_MAX_SIZE: 2500MB
+      STACK_ROOT: "/build/cardano-ledger.stack"
+    command:
+      # cache-s3 needs a build directory that is the same across all buildkite agents.
+      # so copy the source into /build/cardano-ledger
+      - "rm -rf /build/cardano-ledger"
+      - "cp -R . /build/cardano-ledger"
+      - "cd /build/cardano-ledger"
+      - "stack clean"
```
intricate: Hopefully this doesn't increase our CI build times too much.

cardano-ledger.cabal
```diff
@@ -260,3 +260,42 @@ test-suite cardano-ledger-test
...
+
+  default-language:    Haskell2010
+  default-extensions:  NoImplicitPrelude
+
+  ghc-options:         -Weverything
+                       -fno-warn-all-missed-specialisations
+                       -fno-warn-missing-import-lists
+                       -fno-warn-safe
+                       -fno-warn-unsafe
+                       "-with-rtsopts=-K450K -M500M"
```
intricate: Should probably play around with the heap mem capacity option a bit and see if I can set it a bit lower.

intricate (commented):
Some notes, I guess.

test/NormalFormTest.hs
```diff
@@ -0,0 +1,39 @@
+{-# LANGUAGE OverloadedStrings #-}
```
intricate: Perhaps this file could be better named as well.

test/Test/Options.hs
ruhatch: Seems legit ;)

cardano-ledger.cabal
ruhatch: Nah, it's decscriptive!

cardano-ledger.cabal
ruhatch: :+1:

ruhatch:
Looks like the CI times are quite high here - let's see whether hydra can get it done better

test/Test/Cardano/Chain/Block/Validation.hs
```diff
@@ -71,18 +81,42 @@ ts_prop_mainnetEpochsValid scenario = withTests 1 . property $ do
...
+  cvs' <- evalEither result
+
+  case shouldAssertNF of
+    AssertNF -> do
+      annotate ("Did you build with `ghc -fhpc` or `stack --coverage`?\n"
+        <> "If so, please be aware that hpc will introduce thunks around "
+        <> "expressions for its program coverage measurement purposes and "
+        <> "this assertion can fail as a result.\n"
+        <> "Otherwise, for some reason, the `ChainValidationState` is not in "
+        <> "normal form.")
```
mhuesch: 👍

test/test.hs
```diff
@@ -46,7 +41,7 @@ main =
     $   tsGroupToTree
     <$> [ Test.Cardano.Chain.Block.CBOR.tests
         , Test.Cardano.Chain.Block.Model.tests
-        , Test.Cardano.Chain.Block.Validation.tests
+        , Test.Cardano.Chain.Block.Validation.tests NoAssertNF
```
mhuesch: Is there a reason to actually include this into the test-suite, given that `NoAssertNF` essentially makes this evaluate to `pass`?

mhuesch (approved):
LGTM. I like the way you got around the `hpc` thunking issue in CI.

mhuesch:
Looks like this is exhausting the heap in CI:
```
Test.Cardano.Chain.Block.Validation
--
  | prop_signingHistoryUpdatesPreserveInvariants:   OK (0.62s)
  | ✓ prop_signingHistoryUpdatesPreserveInvariants passed 100 tests.
  | ts_prop_mainnetEpochsValid:                     cardano-ledger-test: Heap exhausted;
  | cardano-ledger-test: Current maximum heap size is 524288000 bytes (500 MB).
  | cardano-ledger-test: Relink with -rtsopts and use `+RTS -M<size>' to increase it.
```
https://buildkite.com/input-output-hk/cardano-ledger/builds/2072#3ee17651-55e0-4a5f-82af-a1dc5d61d448

test/test.hs
intricate: Yeah so this still runs the same mainnet mirror epoch validation tests, i.e. it's not a no-op if given a value of `NoAssertNF`. The `ShouldAssertNF` arg just indicates whether to assert that the `ChainValidationState` `isNormalForm` after each epoch is validated.

mhuesch: Oops, right. Somehow I glossed over that. All good then.

cardano-ledger.cabal
```diff
@@ -251,6 +256,52 @@ test-suite cardano-ledger-test
   default-language:    Haskell2010
   default-extensions:  NoImplicitPrelude
 
+  ghc-options:         -Weverything
+                       -fno-warn-all-missed-specialisations
+                       -fno-warn-missing-import-lists
+                       -fno-warn-safe
+                       -fno-warn-unsafe
+                       "-with-rtsopts=-K450K -M1G"
```
intricate: For some reason, I had to double the heap size to 1G in order for the test-suite to run successfully in our buildkite job. Need to determine why this is necessary since the test-suite shouldn't require any more heap mem than it does on the `master` branch.

ruhatch: Yeah, this is odd, we should profile and see what's going on:
```
stack --work-dir=.stack-work-profile test --profile \
   --ta="--pattern=mainnetEpochsValid --scenario=ContinuousIntegration"
```

ruhatch: I've tried reverting this after rebasing on latest `master`, so we'll see what happens

ruhatch:
It turns out that `foldM` was causing serious problems, so I've moved the normal form check into `foldChainValidationState`, which seems to work fine.

@intricate we should still explore while the epoch normal form tests are failing on Windows, but let's leave this for next iteration

intricate:
> It turns out that `foldM` was causing serious problems, so I've moved the normal form check into `foldChainValidationState`, which seems to work fine.
> 
> @intricate we should still explore while the epoch normal form tests are failing on Windows, but let's leave this for next iteration

Understood.

Issue #326, opened 2019-02-14
Redesign `Header` Haskell representation without changing binary format


ruhatch:
The current Haskell representation of the `Header` is confusing and could use some clean up. We should be able to do this without requiring any changes to the binary format by updating the `Bi` instance for `Header` to break it back into the old structure.

PR #493, opened 2019-05-13
[#326] Redesign Header Haskell representation

ruhatch:


dnadales (approved):
Nice. I like that the software and application version bubbled up :)

PR #500, opened 2019-05-15
Consensus Integration

nc6:
This PR incorporates the various changes needed for integration with the consensus layer. They can be divided into three basic ideas:

- Firstly, we expose a bunch of things which were previously hidden, because they're needed.
- Secondly, we change some of the serialisation code to take a more explicit serialiser, rather than using ToCBOR constraints which the consensus layer can typically not satisfy.

This replaces PR #474

### DIFF
--- a/crypto/src/Cardano/Crypto/Signing/Signature.hs
+++ b/crypto/src/Cardano/Crypto/Signing/Signature.hs
@@ -34,6 +34,7 @@ import Cardano.Prelude
 import qualified Cardano.Crypto.Wallet as CC
 import Data.Aeson (FromJSON(..), ToJSON(..))
 import Data.ByteArray (ScrubbedBytes)
+import qualified Data.ByteString.Lazy as BSL
 import Data.Coerce (coerce)
 import Formatting (Format, bprint, build, formatToString, later, sformat, stext)
 import qualified Formatting.Buildable as B
@@ -49,6 +50,7 @@ import Cardano.Binary
   , Raw
   , ToCBOR(..)
   , serialize'
+  , serializeEncoding
   )
 import Cardano.Crypto.ProtocolMagic (ProtocolMagicId)
 import Cardano.Crypto.Signing.VerificationKey (VerificationKey(..))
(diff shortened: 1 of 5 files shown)


dnadales:
> This PR incorporates the various changes needed for integration with the consensus layer. They can be divided into three basic ideas:
> 
I'm afraid to ask where is the third :laughing: 

> * Firstly, we expose a bunch of things which were previously hidden, because they're needed.
> * Secondly, we change some of the serialisation code to take a more explicit serialiser, rather than using ToCBOR constraints which the consensus layer can typically not satisfy.
> 
> This replaces PR #474

crypto/src/Cardano/Crypto/Signing/Signature.hs
```diff
@@ -135,12 +137,12 @@ sign
...
   -> Signature a
-sign pm tag sk = signEncoded pm tag sk . serialize'
+sign pm tag sk = signEncoded pm tag sk . toCBOR
 
 -- | Like 'sign' but without the 'ToCBOR' constraint
 signEncoded
-  :: ProtocolMagicId -> SignTag -> SigningKey -> ByteString -> Signature a
-signEncoded pm tag sk = coerce . signRaw pm (Just tag) sk
+  :: ProtocolMagicId -> SignTag -> SigningKey -> Encoding -> Signature a
+signEncoded pm tag sk = coerce . signRaw pm (Just tag) sk . BSL.toStrict . serializeEncoding
```
dnadales, line 145: If we don't pass the _encoded_ `ByteString` anymore, shouldn't we rename this function to something like `signWithEncoding` or something like that?

dnadales (approved):

crypto/src/Cardano/Crypto/Signing/Signature.hs
nc6: Well, we pass it encoded into an `Encoding`. An encoding is surely something that it's been encoded to?

nc6:
Third change got reverted as it wasn't needed, but I clearly didn't review the comment well enough!

PR #505, opened 2019-05-17
Initialise genesis delegation from the first block

### LINKED ISSUE #504: Initial delegation certificates are applied incorrectly
## Background

In `Cardano.Chain.Delegation.Validation.Interface`, initial delegation is specified by:
- Firstly, constructing an initial delegation map which delegates from stakeholders to themselves.
- Secondly, applying the certificates from the genesis data in the normal way.

The problem is that normally, delegation certificates do not come into effect from the moment they are posted - they are only scheduled, and will come into effect at some future slot. For the genesis configuration, this is incorrect - the initial delegation should be in effect from the first block.

## Proposed Solution

Change `initialState` to create the delegation map directly using the heavyweight delegation certificates, rather than calling `updateDelegation`.

## Next Action
- [x] Confirm this diagnosis (@ruhatch?)
- [ ] Fix all the things

nc6:
I can confirm this addresses the issue we were seeing in ouroboros-consensus

Fixes #504

### DIFF
--- a/src/Cardano/Chain/Delegation/Validation/Interface.hs
+++ b/src/Cardano/Chain/Delegation/Validation/Interface.hs
@@ -23,7 +23,7 @@ import qualified Data.Sequence as Seq
 import qualified Data.Set as Set
 
 import Cardano.Binary (Annotated(..), serialize')
-import Cardano.Chain.Common (BlockCount, StakeholderId, mkStakeholderId)
+import Cardano.Chain.Common (BlockCount(..), StakeholderId, mkStakeholderId)
 import qualified Cardano.Chain.Delegation as Delegation
 import Cardano.Chain.Delegation.Certificate (ACertificate, Certificate)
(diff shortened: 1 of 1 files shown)


ruhatch (approved):
Love it, haha

src/Cardano/Chain/Delegation/Validation/Interface.hs
```diff
@@ -71,9 +71,13 @@ initialState
   => Environment
   -> GenesisDelegation
   -> m State
-initialState env genesisDelegation = updateDelegation env is certificates
+initialState env genesisDelegation = updateDelegation env' is certificates
  where
   Environment { allowedDelegators } = env
+  -- We modify the environment here to allow the delegation certificates to
```
dnadales, line 77: :heart:

dnadales (approved):

PR #507, opened 2019-05-18
Fix Fee Validate Note

felixYing:


ruhatch (commented):
Sorry, this is actually correct - the formal spec is wrong here I believe. @dnadales can you confirm this?

dnadales:
> Sorry, this is actually correct - the formal spec is wrong here I believe. @dnadales can you confirm this?

Yes, this was addressed in https://github.com/input-output-hk/cardano-ledger-specs/pull/475#issuecomment-493541180

Sorry for the confusion @felixYing and thanks!

PR #512, opened 2019-05-20
Relax requirement for proposal to not be empty

ruhatch:


dnadales (approved):
@ruhatch @disassembler let me know if this works so that I can modify the specs.

PR #513, opened 2019-05-20
Update logging shell

CodiePP:
* simplification in the calling of the bracket funcitons: no more access to the logging configuration needed.
* only shows time between epoch validations in EKG


stack.yaml
```diff
@@ -36,16 +37,20 @@ extra-deps:
...
   - git: https://github.com/input-output-hk/iohk-monitoring-framework
-    commit: 00413dcad916a41cf392437199cfba7a2e2c76d7
+    commit: f1c4ceef7d7ea6fb4425484c2b19b84048a3549d
     subdirs:
       - contra-tracer
       - iohk-monitoring
 
+  - ekg-prometheus-adapter-0.1.0.4
+  - prometheus-2.1.1
+  - containers-0.5.11.0
```
ruhatch, line 52: What's this for? Just curious, not a big issue!

ruhatch (approved):
LGTM

stack.yaml
CodiePP: that's a new output which was required by devops. they use prometheus for monitoring/alerting.

PR #521, opened 2019-05-24
[#461] Change StakeholderId to KeyHash

### LINKED ISSUE #461: Change StakeholderId to HashKey
## Background

In the Shelley specs we use `HashKey` to refer to the `Hash` of a `VerificationKey`, which in the Byron implementation corresponds to a `StakeholderId`. I have always found `StakeholderId` to be a confusing term and it would be good to bring the two versions into alignment.

## Proposed Solution

Change the name of `StakeholderId` to `HashKey`.

## Next Actions
- [ ] Change name of `StakeholderId` to `HashKey`
- [ ] Change `mkStakeholderId` function to `hashKey` or similar

ruhatch:
Closes #461


intricate:
_n.b. We can disregard this comment if it isn't really a big deal._

Not opposed to this as I understand why we'd like to do this (in order to remain consistent with the naming in the Shelley specs) but I'd like to raise a couple of concerns with regard to the name, `HashKey`, and get your opinion on them @ruhatch :

1. In my opinion, based on the name, it's not entirely clear whether this data type represents the hash of a symmetric or some asymmetric key. i.e. Maybe we should include the term `VerificationKey` in the name.

2. Perhaps I'm being too picky but `HashKey` _could_ sound like it represents a key which is used in some sort of cryptographic hash function rather than the hash of a key. For example, it could sound like the secret key which is utilized in constructing an HMAC (although, `HmacKey` would be a better name for that). i.e. Maybe we should swap the words `Hash` and `Key` such that we have a `KeyHash`.

So maybe something like `VerificationKeyHash` or `KeyHash` might be better?

ruhatch:
My main concern here was getting inline with the spec terminology, so I'd ask @dnadales and @JaredCorduan to explain the name `HashKey` from the spec

dnadales:
> My main concern here was getting inline with the spec terminology, so I'd ask @dnadales and @JaredCorduan to explain the name `HashKey` from the spec

See https://github.com/input-output-hk/cardano-ledger-specs/pull/438#discussion_r281140393

I don't know if `HashKey` is the best name for the Byron spec, so I'm open to suggestions.

dnadales (approved):
:+1: :key:

PR #522, opened 2019-05-27
Removed obsolete cardano-mainnet-mirror submodule.

jbgi:
Now provided through a nix import since #476.
 Side note: using git submodule makes nix caching
 less efficient, especially for large submodules.


ruhatch (approved):
LGTM!

bors r+

PR #527, opened 2019-05-29
[#515] Replace hedgehog state machine model with Trace-based one

### LINKED ISSUE #515: Incorporate New UTxO Generators in UTXOW State Machine
## Background

The initial generators that we used for the UTXOW state machine testing relied heavily on the `Monad` instance of `Gen` and therefore have relatively poor shrinking properties. input-output-hk/cardano-ledger-specs#489 replaces these generators to use the applicative instance and manual shrinking. We should update the state machine model to use these new generators.

## Proposed Solution

Update the state machine generators to use the new ones from `cardano-ledger-specs`.

## Next Action
- [ ] Update the state machine generators to use the new ones from `cardano-ledger-specs`
- [ ] Make sure that the UTXOW model tests still pass

ruhatch:
Closes #515

### DIFF
--- a/test/Test/Cardano/Chain/Block/Validation/Spec.hs
+++ b/test/Test/Cardano/Chain/Block/Validation/Spec.hs
@@ -1,7 +1,8 @@
 {-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE OverloadedLists #-}
-{-# LANGUAGE TemplateHaskell #-}
-{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE OverloadedLists  #-}
+{-# LANGUAGE TemplateHaskell  #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies     #-}
 
 -- | Test module where we check that the block validation implementation
 -- matches the formal specification. To this end, the strategy is:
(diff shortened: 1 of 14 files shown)


test/Test/Cardano/Chain/Block/Validation/Spec.hs
```diff
@@ -51,7 +52,7 @@ tests = $$discoverPropArg
 ts_prop_generatedChainsAreValidated :: TSProperty
 ts_prop_generatedChainsAreValidated =
   withTestsTS 300 $ property $ do
-    tr <- forAll trace
+    tr <- forAll $ trace @CHAIN 100
```
dnadales, line 55: Would it make more sense to test `100` times with traces of length `300`?

What happens if we extend this to `200`?

test/Test/Cardano/Chain/UTxO/Model.hs
```diff
@@ -1,303 +1,118 @@
...
-{-# LANGUAGE RankNTypes          #-}
-{-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TemplateHaskell     #-}
-{-# LANGUAGE TypeApplications    #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TemplateHaskell   #-}
+{-# LANGUAGE TupleSections     #-}
+{-# LANGUAGE TypeApplications  #-}
 
-{-# OPTIONS_GHC -fno-warn-orphans #-}
```
dnadales, line 13: :tada:

dnadales (approved):
Looks great! Let's see how far traces can get us :four_leaf_clover:

test/Test/Cardano/Chain/Block/Validation/Spec.hs
ruhatch: Not sure yet, as this one's not hooked up :P I'm going to revive it in the next PR

ruhatch: But yeah, not sure what the best parameters are for these

PR #529, opened 2019-05-30
Add nightly buildkite pipeline

ruhatch:
As well as adding the nightly tests, I also moved the `cardano-ledger` files into a `cardano-ledger` directory, which improves nix caching while testing changes to `nix` infrastructure.


.buildkite/nightly.yml
```diff
@@ -0,0 +1,7 @@
+steps:
+  - label: 'nightly tests'
+    command:
```
rvl, line 3: I prefer to have the buildkite command call a shell script (e.g. `.buildkite/nightly-tests.sh`). That way it's easier to test locally.

rvl (approved):
I have a feeling that running `makeWrapper` over a haskell.nix test-suite will be painful.

Try adding this to the test script instead. It's not that much duplication.
```
export CARDANO_MAINNET_MIRROR=$(nix-build nix/cardano-mainnet-mirror.nix --arg pkgs 'import <nixpkgs> {}' --no-out-link)/epochs
```
(edited to add `/epochs` to the variable)

.buildkite/nightly.yml
deepfire: :+1:

.buildkite/nightly.yml
```diff
@@ -0,0 +1,6 @@
...
+      - "./nightly-tests.sh"
```
rvl: Cool, and Buildkite command paths are relative to the repo root, not relative to the pipeline.yaml.
```suggestion
      - ".buildkite/nightly-tests.sh"
```

.buildkite/nightly-tests.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+nix build -f ../default.nix nix-tools.tests.cardano-ledger.cardano-ledger-test -o cardano-ledger-test
```
rvl: This will probably need something like a
```
cd `dirname $0`/..
```
So that the script can be run from any working directory.

deepfire (approved):
LGTM!

PR #532, opened 2019-05-31
[#516] Add UTxO Generators to CHAIN State Machine Tests

### LINKED ISSUE #516: Add UTxO Generators to CHAIN State Machine Tests
## Background

The CHAIN state machine tests currently only use generators for the delegation certificates. We should extend them to incorporate UTxO generators. This will require using the `Var` types from `hedgehog` to capture the transaction IDs and there may be further complications, as multiple transactions are generated at each step of the state machine.

## Proposed Solution

We should try to incorporate the UTxO generators and then see whether there is an obvious way to get the `Var`s working. If this turns out to be quite complex we should think carefully about whether generating `Trace`s directly might be a better solution.

## Next Action
- [x] Try to incorporate the UTxO generators in the naive way
- [x] Take stock of whether the approach seems like a good one

ruhatch:
The amount of tests run is quite low at the moment, because of the performance of the generators, but hopefully we'll see some improvements after input-output-hk/cardano-ledger-specs#518

Closes #516


src/Cardano/Chain/Block/Validation.hs
```diff
@@ -224,7 +223,7 @@ data ChainValidationError
   | ChainValidationHeaderAttributesTooLarge
   -- ^ The size of a block header's attributes is non-zero
 
-  | ChainValidationHeaderTooLarge
+  | ChainValidationHeaderTooLarge Natural Natural
```
dnadales, line 226: Maybe here we could use the `TooLarge` trick of `cardano-ledger`. You don't have to do it here, just a thought ...

test/Test/Cardano/Chain/Block/Model.hs
```diff
@@ -1,155 +1,104 @@
...
-{-# LANGUAGE TypeApplications    #-}
-
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE OverloadedLists  #-}
+{-# LANGUAGE TemplateHaskell  #-}
+{-# LANGUAGE TupleSections    #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies     #-}
+
+-- | Test module where we check that the block validation implementation
```
dnadales, line 8: :pray: :100:

test/Test/Cardano/Chain/Block/Model.hs
```diff
@@ -1,155 +1,104 @@
...
-            certs' = if isRight result
-              then block ^. Abstract.bBody . Abstract.bDCerts ++ certs
-              else certs
-          in StateCHAIN (fromRight abstractState result) certs' result
-    , Ensure $ \_ StateCHAIN { lastAbstractResult } _ result -> do
-      annotateShow lastAbstractResult
-      annotateShow result
-      isRight lastAbstractResult === isRight result
-    ]
+-- classifyTransactions :: Trace CHAIN -> PropertyT IO ()
```
dnadales: Any reason why we're leaving this commented out?

dnadales:
image

Zen!

test/Test/Cardano/Chain/Elaboration/Block.hs
```diff
@@ -212,9 +216,9 @@ abEnvToCfg (_, _, dsEnv, pps) =
   gPps        = Update.ProtocolParameters
     { Update.ppScriptVersion    = 0
     , Update.ppSlotDuration     = 0
-    , Update.ppMaxBlockSize     = pps ^. maxBkSz
-    , Update.ppMaxHeaderSize    = pps ^. maxHdrSz
-    , Update.ppMaxTxSize        = 0
+    , Update.ppMaxBlockSize     = 100 * pps ^. maxBkSz
```
dnadales, line 219: We might want to comment where this `100` comes from.

test/Test/Cardano/Chain/UTxO/Model.hs
```diff
@@ -44,7 +46,7 @@ tests = $$discoverPropArg
 --   UTxO validator.
 ts_prop_generatedChainsAreValidated :: TSProperty
 ts_prop_generatedChainsAreValidated =
-  withTestsTS 200 $ property $ do
+  withTestsTS 2 $ property $ do
```
dnadales: :)

test/Test/Cardano/Chain/UTxO/Model.hs
```diff
@@ -54,7 +56,7 @@ passConcreteValidation :: MonadTest m => Trace (UTXOW Abstract.TxId) -> m ()
 passConcreteValidation tr = void $ evalEither res
  where
   res = foldM (elaborateAndUpdate abstractEnv) initSt
-    $ preStatesAndSignals OldestFirst tr
+    $ traceSignals OldestFirst tr
```
dnadales, line 59: Cool that you could get rid of the pre-states!

dnadales (approved):
LGTM!

src/Cardano/Chain/Block/Validation.hs
ruhatch: Yeah, I'm going to add an issue for a general purpose way of having expected values

test/Test/Cardano/Chain/UTxO/Model.hs
ruhatch: Woops, that's not meant to be so low, haha!

test/Test/Cardano/Chain/UTxO/Model.hs
ruhatch: Yeah, the elaboration got nice and simple with the `TxId` map :)

PR #533, opened 2019-05-31
[#429] Configuration conversion function 

### LINKED ISSUE #429: `CardanoConfiguration` and `Config` reconciliation
## Background

Currently `cardano-shell` inherits a [`CardanoConfiguration`](https://github.com/input-output-hk/cardano-shell/blob/develop/src/Cardano/Shell/Constants/Types.hs#L36) which is meant to pass values such as `RequiresNetworkMagic` to the bulk chain validator in `cardano-ledger`. 

The issue is, for example, [`RequiresNetworkMagic`](https://github.com/input-output-hk/cardano-ledger/blob/08a8f128ae55c5f8de873404e3ea57537b16ada9/crypto/src/Cardano/Crypto/ProtocolMagic.hs#L82) exists as a sum type in `cardano-ledger` whilst it exists as a [`Text`](https://github.com/input-output-hk/cardano-shell/blob/develop/src/Cardano/Shell/Constants/Types.hs#L61) value in `cardano-shell`. 

The correct type can still be generated via `mkConfigFromFile` in `cardano-ledger` which will produce a `Config` with the correct `RequiresNetworkMagic` data type, however this seems to add to the confusion if we also have a field in `CardanoConfiguration` indicating that `RequiresNetworkMagic` is a `Text` value.

Relevant issue: https://github.com/input-output-hk/cardano-shell/issues/87

## Proposed Solution

After chatting with @mhuesch, we decided a pure conversion function would be suitable for now. Something like the following:

`convertConfig :: Monad m => Shell.Config -> ExceptT ConfigError m Ledger.Config`

## Next Action
- [x] Implement configuration conversion function

Jimbo4350:
Closes #429


src/Cardano/Chain/Conversion.hs
```diff
@@ -0,0 +1,227 @@
...
+  (GenesisInitializer(..), FakeAvvmOptions(..), TestnetBalanceOptions(..))
+import Cardano.Chain.Genesis.Spec (mkGenesisSpec)
+import Cardano.Chain.Slotting (EpochIndex(..), FlatSlotId(..))
+import Cardano.Chain.Update (ProtocolParameters(..), SoftforkRule(..))
+import Cardano.Crypto.Hashing (Hash, decodeAbstractHash)
+import Cardano.Crypto.ProtocolMagic
+  (AProtocolMagic(..), ProtocolMagicId(..), RequiresNetworkMagic(..))
+import Cardano.Shell.Constants.Types as Shell
+
+-- | This module converts `CardanoConfiguration` to `Genesis.Config`
```
ruhatch: You can move this to the line about `module ... where` to get this as an actual module header in Haddock

src/Cardano/Chain/Conversion.hs
```diff
@@ -0,0 +1,227 @@
...
+  -- ^ Convert configuration based on 'geSpec'/`GCSpec`
+  --   If we provide a seed ('Maybe Integer') this will override
+  --   the seed in `GenesisSpec` and we will also need to provide
+  --   a start time (Maybe UTCTime)
+
+-- | Converts 'CardanoConfiguration' (from 'cardano-shell') into a `Config` (from 'cardano-ledger')
+convertConfig
+  :: SpecifyGenesis
+  -> Shell.CardanoConfiguration
+  -> EitherT ConfigurationError IO Config
```
ruhatch: I think I'd prefer if we used `MonadError` to remain consistent with the rest of the codebase

ruhatch: Then you can also make use of `wrapError`, which we use in the validation rules a lot

src/Cardano/Chain/Conversion.hs
```diff
@@ -0,0 +1,227 @@
...
+    let pmId            = prProtocolMagic $ spProtocolConstants shellSpec
+    let pm = AProtocolMagic (Annotated (ProtocolMagicId pmId) ()) reqNM
+
+     -- GenesisInitializer
+    let init            = spInitializer shellSpec
+
+    -- TestNetBalance
+    let shellTestNetBal = inTestBalance init
+    rMenShare <- firstEitherT ConfigPortionConvErr
+      $ hoistEither (lovelacePortionFromDouble $ teRichmenShare shellTestNetBal)
```
ruhatch: So this would become:
```haskell
    rMenShare <- lovelacePortionFromDouble (teRichmenShare shellTestNetBal)
      `wrapError` ConfigPortionConvErr
```

src/Cardano/Chain/Genesis/Config.hs
```diff
@@ -230,7 +227,16 @@ mkConfigFromStaticConfig rnm confDir mSystemStart mSeed = \case
 
     isNothing mSeed `orThrowError` MeaninglessSeed
 
-    mkConfigFromFile rnm (confDir </> fp) (Just expectedHash)
+    -- Check expectedHash is equivalent to hash of `mainnet-genesis.json` at fp
+    (_, genesisHash) <-
+      liftEither . first ConfigurationGenesisDataError =<< runExceptT
```
ruhatch: Can use `wrapError` here too :)

ruhatch: You would have to `<$>` it over the `IO` though

src/Cardano/Chain/Genesis/Config.hs
```diff
@@ -230,7 +227,16 @@ mkConfigFromStaticConfig rnm confDir mSystemStart mSeed = \case
...
+        (readGenesisData fp)
+
+    unGenesisHash genesisHash
```
ruhatch: Looks like you're already checking this in `mkConfigFromFile`, so you can remove this

stack.yaml
```diff
@@ -29,14 +29,15 @@ extra-deps:
       - byron/chain/executable-spec
 
   # Currently Hackage only
-  - tasty-hedgehog-1.0.0.0
+  - tasty-hedgehog-1.0.0.1
+  - transformers-either-0.1.1
```
ruhatch: Should be able to remove this if you switch to `MonadError` :)

stack.yaml
```diff
@@ -29,14 +29,15 @@ extra-deps:
       - byron/chain/executable-spec
 
   # Currently Hackage only
-  - tasty-hedgehog-1.0.0.0
+  - tasty-hedgehog-1.0.0.1
```
ruhatch: I guess you should rebase on master as this is already updated

test/Test/Cardano/Chain/Config.hs
```diff
@@ -19,10 +21,16 @@ import Cardano.Crypto.ProtocolMagic (RequiresNetworkMagic(..))
...
 -- We use `RequiresNoMagic`, as it indicates mainnet
 readMainetCfg :: MonadIO m => m Genesis.Config
-readMainetCfg =
-  either
-      (panic . sformat build)
-      identity
-    <$> runExceptT
-          (Genesis.mkConfigFromFile RequiresNoMagic "mainnet-genesis.json" Nothing)
+readMainetCfg = do
+  genHash <-
```
ruhatch: I think this is just
```
let genHash = either (panic . sformat build . Genesis.GenesisHashDecodeError) identity (decodeAbstractHash ...)
```

ruhatch (changes requested):
Looks good, but a few changes to bring it in line with the rest of the codebase

ruhatch (approved):
LGTM!

src/Cardano/Chain/Common/NetworkMagic.hs
```diff
@@ -38,7 +37,3 @@ makeNetworkMagic :: AProtocolMagic a -> NetworkMagic
 makeNetworkMagic pm = case getRequiresNetworkMagic pm of
     RequiresNoMagic -> NetworkMainOrStage
     RequiresMagic   -> NetworkTestnet (getProtocolMagic pm)
-
-parseReqNetworkMag :: Text -> RequiresNetworkMagic
-parseReqNetworkMag "NetworkMainOrStage" = RequiresNoMagic
-parseReqNetworkMag _                    = RequiresMagic
```
mhuesch: 👍

src/Cardano/Chain/Conversion.hs
```diff
@@ -0,0 +1,64 @@
...
+--------------------------------------------------------------------------------
+-- Parsers
+--------------------------------------------------------------------------------
+
+pRequiresNetworkMagic :: Parsec Void Text RequiresNetworkMagic
+pRequiresNetworkMagic = choice
+  [ RequiresNoMagic <$ string "RequiresNoMagic"
+  , RequiresMagic <$ string "RequiresMagic"
+  ]
+
```
mhuesch: This looks great 🙌

src/Cardano/Chain/Genesis/Config.hs
```diff
@@ -230,7 +227,7 @@ mkConfigFromStaticConfig rnm confDir mSystemStart mSeed = \case
 
     isNothing mSeed `orThrowError` MeaninglessSeed
 
-    mkConfigFromFile rnm (confDir </> fp) (Just expectedHash)
+    mkConfigFromFile rnm fp expectedHash
```
mhuesch: Is the removal of the `confDir` argument possible because `fp` is a full path instead of a filename?

src/Cardano/Chain/Genesis/Config.hs
```diff
@@ -308,6 +303,12 @@ data ConfigurationError
   | MeaninglessSeed
   -- ^ Custom seed was provided, but it doesn't make sense
   | ConfigurationGenerationError GenesisDataGenerationError
+  | GenesisHashDecodeError Text
+  | ConfigParsingError (ParseErrorBundle Text Void)
+  -- ^ An error occured while parsing 'CardanoConfiguration'
+  | ConfigPortionConvErr LovelacePortionError
+  | ConfigLovelaceConvErr LovelaceError
+  | ConfigGenSpecConvErr Text
```
mhuesch: Given that all the other constructors have Haddock fields, it would be nice if these all did too.

src/Cardano/Chain/Genesis/NonAvvmBalances.hs
```diff
@@ -90,10 +90,9 @@ convertNonAvvmDataToBalances balances = fmap GenesisNonAvvmBalances $ do
  where
   mkBalances :: [(Address, Lovelace)] -> m (Map Address Lovelace)
   mkBalances =
-    liftEither
       -- Pull 'LovelaceError's out of the 'Map' and lift them to
       -- 'NonAvvmBalancesError's
-      . first NonAvvmBalancesLovelaceError
+      (`wrapError` NonAvvmBalancesLovelaceError)
```
mhuesch: This is a little pedantic, but this block is indented by 4 spaces and could be outdented to just 2 spaces, for consistency with the surrounding code.

src/Cardano/Chain/Update/ProtocolParameters.hs
```diff
@@ -45,15 +45,15 @@ data ProtocolParameters = ProtocolParameters
...
   -- ^ Time to live for a protocol update proposal. This used to be the number
   -- of slots after which the system made a decision regarding an update
   -- proposal confirmation, when a majority of votes was not reached in the
   -- given number of slots. If there were more positive than negative votes the
   -- proposal became confirmed, otherwise it was rejected. Since in the
   -- Byron-Shelley bridge we do not have negative votes, and we aim at
-  -- simplifying the update mechanism, 'ppUpdateImplicit' is re-interpreted as
-  -- the number of slots a proposal has to gather a majority of votes. If a
+  -- simplifying the update mechanism, 'ppUpdateProposalTTL' is re-interpreted as
+  -- the slot up to which a proposal has to gather a majority of votes. If a
```
mhuesch: This seems unrelated to the Config conversion function, and significant. Are we indeed changing the interpretation of this field, to be an absolute `FlatSlotId`? Why is it bundled into this PR?

This seems to be at odds with the [main source-code usage of `ppUpdateProposalTTL`](https://github.com/input-output-hk/cardano-ledger/blob/e5d6844bc64bb6e720044511d74f6541964fe510/src/Cardano/Chain/Update/Validation/Interface.hs#L305) (param `u` comes from [here](https://github.com/input-output-hk/cardano-ledger/blob/e5d6844bc64bb6e720044511d74f6541964fe510/src/Cardano/Chain/Update/Validation/Interface.hs#L360)) which does filtering relative to the current slot. That should change if we're changing from relative to absolute.

This change also seems inconsistent all the other uses of `ppUpdateProposalTTL`:
https://github.com/input-output-hk/cardano-ledger/blob/e5d6844bc64bb6e720044511d74f6541964fe510/src/Cardano/Chain/Update/ProtocolParametersUpdate.hs#L60
https://github.com/input-output-hk/cardano-ledger/blob/e5d6844bc64bb6e720044511d74f6541964fe510/test/Test/Cardano/Chain/Elaboration/Block.hs#L227
https://github.com/input-output-hk/cardano-ledger/blob/e5d6844bc64bb6e720044511d74f6541964fe510/test/Test/Cardano/Chain/Genesis/Dummy.hs#L126

mhuesch: It also seems at odds with [the definition in `cardano-shell`](https://github.com/input-output-hk/cardano-shell/blob/44fa2129e07668ebc620708c3a83913cff6fff7c/src/Cardano/Shell/Presets.hs#L256) (documented [here](https://github.com/input-output-hk/cardano-shell/blame/44fa2129e07668ebc620708c3a83913cff6fff7c/configuration.md#L44)), so I'm wondering what motivated the change, or if it was an accident.

src/Cardano/Chain/Genesis/Config.hs
```diff
@@ -256,18 +253,17 @@ mkConfigFromFile
...
+  -> Hash Raw
+  -- ^ This hash comes from 'CardanoConfiguration'
+  -- which lives in cardano-shell
   -> m Config
-mkConfigFromFile rnm fp mGenesisHash = do
+mkConfigFromFile rnm fp expectedHash = do
   (genesisData, genesisHash) <-
-    liftEither . first ConfigurationGenesisDataError =<< runExceptT
+    (`wrapError` ConfigurationGenesisDataError) =<< runExceptT
       (readGenesisData fp)
```
mhuesch: I didn't think much of this on the first read-through, but now I am wondering -

My understanding of the point of `cardano-shell` was that once it was done, we wouldn't need JSON decoders in `cardano-ledger` anymore. We would do something more like pure datatype-to-datatype conversion, with no parsing and e.g. the possibility of `read` failing on a numeric field.

The way things are set up in this PR, it seems like all that `cardano-shell` has done is introduced another layer of parsing and serialization. We still have the old JSON decoding logic, but we now have a new layer between us and the on-disk configuration. It is unclear to me that this is at all better - it actually seems worse.

mhuesch (changes requested):
2 issues that I see (noted in line-comments):

1. `ppUpdateProposalTTL` has been changed in a breaking way
2. I'm unclear on the story of `cardano-shell`

@Jimbo4350 it might be best to talk this over synchronously - perhaps in the core team meeting.

src/Cardano/Chain/Conversion.hs
erikd: `MonadError` and `ExceptT` are basically isomorphic, but `MonadError` uses names like`throwError` and `catchError` that are easily confused with `throw`/`catch` used with exceptions.

I prefer the `EitherT` or `ExceptT` approach because the types are explicit, and using the combinator functions in `transformers-either`  handling errors in a principled way is super convenient.

However since the rest of the code base uses `MonadError` (with its poorly chosen method names) we should use that.

src/Cardano/Chain/Update/ProtocolParameters.hs
mhuesch: Ah. I was thrown off by the change of "number of slots" to "slot up to which". The above comments can be ignored.

I suggest we change that back, and maybe clarify that this `FlatSlotId` is a delta (to be added to / subtracted from an absolute slot), rather than an absolute index of a slot.

```suggestion
  -- the number of slots a proposal has to gather a majority of votes. If a
```

src/Cardano/Chain/Genesis/Config.hs
mhuesch: After talking this over in our meeting, and then revisiting & rereading the code, I think the current situation is fine. The benefit of `cardano-shell` seems to be less obvious from `cardano-ledger`'s point of view, because most of our config is stuck in the genesis block (thus stuck as JSON) and so we have to keep the `FromJSON` instances around to extract things.

I imagine (but I haven't checked yet) that the other parts of the project (e.g. `ouroboros-consensus` and `ouroboros-network` have greater apparent benefit from `cardano-shell` having factored out the configuration (thus insulating them from it).

src/Cardano/Chain/Conversion.hs
```diff
@@ -0,0 +1,64 @@
...
+  , mkConfigFromFile
+  )
+import Cardano.Crypto.Hashing (Hash, decodeAbstractHash)
+import Cardano.Crypto.ProtocolMagic
+  (RequiresNetworkMagic(..))
+import Cardano.Shell.Constants.Types as Shell
+
+
+-- | Converts 'CardanoConfiguration' (from 'cardano-shell') into a `Config` (from 'cardano-ledger')
+convertConfig
```
mhuesch:
```suggestion
--   This function mostly relies on FromJSON parsing, because much of `Config` originates from
--   data locked in the genesis block as JSON. Sadly, this means that `cardano-ledger` can't
--   benefit as much from `cardano-shell`'s function as an insulating layer around config &
--   parsing.
convertConfig
```

In keeping with my comment below [here](https://github.com/input-output-hk/cardano-ledger/pull/533/files#r290500402), I'd suggest adding something vaguely like this as documentation.

mhuesch (commented):
Added additional notes - the major issues are resolved, but the minor tweaks still stand.

src/Cardano/Chain/Update/ProtocolParameters.hs
Jimbo4350: Agreed, I think there was a miscommunication on slack. Good catch!
https://input-output-rnd.slack.com/archives/CCZSDE005/p1559143010050200

src/Cardano/Chain/Conversion.hs
Jimbo4350: I think as it stands now we generate the mainnet config this way, but I think in the future for generating testnet configs we would use `CardanoConfiguration`.

mhuesch: Mainnet will always depend on the JSON locked in the genesis block, though, so the point still stands. But we could add a caveat that it only applies to mainnet.

src/Cardano/Chain/Genesis/Config.hs
mhuesch: Jordan says:

> I think as it is now, you would have to specify the full path in `CardanoConfiguration`
> 
> This simplifies thing imo

Jimbo4350: Currently you would have to specify the full path in `CardanoConfiguration` and this simplifies things imo.

src/Cardano/Chain/Conversion.hs
mhuesch: 👍

mhuesch (approved):
LGTM :shipit:

PR #539, opened 2019-06-05
Update conformance tests with transaction fees

### LINKED ISSUE #535: Update conformance tests with transaction fees
## Background

The transaction generators that @ruhatch wrote to incorporate Edsko's shrinking concerns didn't include transaction fees. After those generators have been updated we should update our dependency on `cardano-ledger-specs` and get the conformance tests passing with transaction fees included. This shouldn't require any changes to the elaboration code.

## Proposed Solution

Update the dependency on `cardano-ledger-specs` once the generators are complete and fix any resulting errors in the conformance tests.

ruhatch:
Closes #535


dnadales (approved):
:+1:

intricate (approved):

test/Test/Cardano/Chain/Elaboration/Delegation.hs
```diff
@@ -73,6 +72,7 @@ ts_prop_elaboratedCertsValid =
       $ [0 .. 6]
     , _dSEnvEpoch       = Epoch 0
     , _dSEnvSlot        = Slot 0
+    , _dSEnvK           = BlockCount 2160
```
intricate: Any issue with hardcoding this? We were previously using `Ledger.GlobalParams.k` which was also equal to `BlockCount 2160` [but this has been removed in `cardano-ledger-specs`](https://github.com/input-output-hk/cardano-ledger-specs/pull/557).

ruhatch: That's fine for now, yeah

Issue #348, opened 2019-03-05
Simplify proxy signing


ruhatch:
After #347 the `ProxySecretKey` will be called `ProxyVerificationKey`. This is being used as a delegation certificate and nothing else, but is designed as a more general scheme. I think we should remove this premature abstraction, so I propose to:
- Remove the `omega` parameter and concretise to `EpochIndex`
- Inline the definitions from `ProxySecretKey` to `Delegation.Certificate`
- Perhaps change the name to `Delegation.OperationalCertificate` for forwards compatibility with Shelley

AskAlexSharov:
Looks like done by https://github.com/input-output-hk/cardano-ledger/issues/321

dnadales:
Hi @AskAlexSharov! Thanks for checking this out.

If I'm not mistaken, these are two different issues: #321 was closed by #347, but [currently](https://github.com/input-output-hk/cardano-ledger/blob/b6bb4f9d288820c2eb26059a4d70b69cd3d26eec/crypto/src/Cardano/Crypto/Signing/Proxy/VerificationKey.hs#L56) we still have an omega (`aPskOmega`) parameter lying around in `ProxyVerificationKey`:

```haskell
-- | Convenient wrapper for secret key, that's basically ω plus
-- certificate.
data AProxyVerificationKey w a = UnsafeAProxyVerificationKey
  { aPskOmega     :: Annotated w a
  , pskIssuerPk   :: PublicKey
  , pskDelegatePk :: PublicKey
  , pskCert       :: ProxyCert w
  } deriving (Eq, Ord, Show, Generic, Functor)
    deriving anyclass NFData
```

And the other two tasks mentioned in this issue still need to be done.

ruhatch:
In order to make progress on this I need to decide what to do with `ProxySignature` first. My initial plan is to remove `ProxySignature` entirely, unifying `proxySign` with the regular `sign` function. This requires a modification to the `SignTag` type that will allow the `sign` function to append the `VerificationKey` in the prefix. So we need to:

## Next Actions
- [x] Modify the `SignTag` type to use `VerificationKey` in the `SignBlock` constructor
- [x] Replace calls to `proxySign` and `proxyVerify` with `sign` and `verify` calls
- [x] Split `BlockSignature` into two parts in `Header` and preserve binary format
- [x] Remove `ProxySignature` entirely

ruhatch:
[...]

PR #541, opened 2019-06-06
[#348] Remove ProxySignature and modify SignBlock to include key

ruhatch:
The ProxySignature was only used for the BlockSignature, so we move the body to
there. We also remove the proxySign function and replace it with regular sign.
This requires adding a value to the SignBlock constructor of SignTag, that gets
prepended to the original tag.


nc6 (approved):
LGTM. I'll make corresponding changes in ouroboros-consensus

src/Cardano/Chain/Block/Header.hs
```diff
@@ -462,31 +460,47 @@ wrapBoundaryBytes = mappend "\130\NUL"
 -- BlockSignature
 --------------------------------------------------------------------------------
 
+type BlockSignature = ABlockSignature ()
+
 -- | Signature of the 'Block'
 --
 --   We use a heavyweight delegation scheme, so the signature has two parts:
```
dnadales, line 472: This might be confusing: if we're using a heavyweight delegation scheme then we wouldn't need to present the delegation certificate. I don't know whether it'd make sense to clarify why do we do this...

dnadales: Related to https://github.com/input-output-hk/cardano-ledger-specs/issues/244

src/Cardano/Chain/Block/Header.hs
```diff
@@ -462,31 +460,47 @@ wrapBoundaryBytes = mappend "\130\NUL"
...
 --
 --   1. A delegation certificate from a genesis key to the block signer
 --   2. The actual signature over `ToSign`
-newtype BlockSignature = BlockSignature
-  { unBlockSignature :: ProxySignature EpochIndex ToSign
-  } deriving (Show, Eq, Generic)
+data ABlockSignature a = ABlockSignature
+  { delegationCertificate :: Delegation.ACertificate a
```
dnadales, line 477: :tada:

dnadales (approved):

src/Cardano/Chain/Block/Header.hs
ruhatch: I'm going to split this again later so that we just have `Delegation.Certificate` and `Signature` directly in the header. I'll add a comment to the `Delegation.Certificate` there once that's done.

PR #547, opened 2019-06-11
[#348] Move ProxyVerificationKey to Delegation.Certificate

### LINKED ISSUE #348: Simplify proxy signing

ruhatch:
This moves the `ProxyVerificationKey` from `cardano-crypto-wrapper` into `cardano-ledger`, as it is only ever used as a `Delegation.Certificate`.

Closes #348


intricate (approved):
This definitely makes more sense being referred to as a `Certificate`. ~However, shouldn't we also remove `AProxyVerificationKey` from `cardano-crypto-wrapper`?~

EDIT: Oops, didn't notice that you actually did remove it :+1:

cardano-ledger/src/Cardano/Chain/Delegation/Certificate.hs
```diff
@@ -1,45 +1,204 @@
...
+--------------------------------------------------------------------------------
+-- Certificate
+--------------------------------------------------------------------------------
 
--- | A delegation certificate is a `ProxyVerificationKey` tagged with an `EpochIndex`
 type Certificate = ACertificate ()
 
-type ACertificate a = AProxyVerificationKey EpochIndex a
+-- | Delegation certificate allowing the @delegateVK@ to sign blocks on behalf
+--   of @issuerVK@, from epoch @aEpoch@ until a new delegation is activated
```
dnadales: BTW, in the Byron rules we don't check that the activation epoch corresponds with the epoch in the certificate. Epochs are used for replay protection and protection against resource exhaustion only. Is this ok?

dnadales (approved):

cardano-ledger/src/Cardano/Chain/Delegation/Certificate.hs
ruhatch: Yeah, good point, I'll adjust this comment

PR #548, opened 2019-06-11
[#526] Implement multiple validation modes

### LINKED ISSUE #526: Implement multiple validation modes
### Multiple Validation Modes
_Related to [input-output-hk/ouroboros-network#440](https://github.com/input-output-hk/ouroboros-network/issues/440)_

- We need to be able to run validation with a signal that may have been previously applied
- There should be three scenarios: NewSignal/NewState, OldSignal/NewState, OldSignal/OldState
- We should customise which predicates should be run depending on the situation

intricate:
Closes #526


cardano-ledger/src/Cardano/Chain/Block/Validation.hs
```diff
@@ -304,6 +310,7 @@ updateChainBoundary
   -> BoundaryValidationData ByteString
   -> m ChainValidationState
 updateChainBoundary cvs bvd = do
+  -- @intricate: Should we only check these cases depending on `BlockValidationMode`?
   case (cvsPreviousHash cvs, boundaryPrevHash bvd) of
```
erikd: I think those are important enough and cheap enough to check in all validation modes.

cardano-ledger/src/Cardano/Chain/Block/Validation.hs
```diff
@@ -354,39 +361,49 @@ data BodyState = BodyState
...
-
+  whenBlockValidate bvmode $
+    proofDelegation (blockProof b)
+      == hashDecoded (blockDlgPayload b)
+      `orThrowError` ChainValidationDelegationProofError
+
+  -- @intricate: How should we deal with the `BlockValidationMode` here?
+  --             When in the 'BlockRevalidate' mode, it shouldn't be possible
+  --             to return an error. Perhaps there should also be a
+  --             'DelegationValidationMode'?
```
erikd: Surely there are some things that are serious enough to warrant being an error in all block validation modes.

cardano-ledger/src/Cardano/Chain/Block/Validation.hs
```diff
@@ -504,6 +523,10 @@ epochTransition
   -> m UPI.State
 epochTransition env st slot = if nextEpoch > currentEpoch
   then
+    -- @intricate: It doesn't seem as if 'UPI.registerEpoch' actually
+    --             returns any errors at this time. Should I still pass a
+    --             'BlockValidationMode' through? Perhaps it would serve in
+    --             remininding us to utilize it in the future, if necessary.
```
erikd: Probably worth asking about that. If that validation does not return any errors its not complete.

cardano-ledger/src/Cardano/Chain/Block/Validation.hs
intricate: Perhaps so but I'm only questioning this because of Edsko's explanation in this issue: https://github.com/input-output-hk/ouroboros-network/issues/440

>```
>      -- | Apply a previously applied block
>      --
>      -- Since a block can only be applied to a single, specific, ledger state,
>      -- if we apply a previously applied block again it will be applied in the
>      -- very same ledger state, and therefore can't possibly fail.
>      --
>      -- NOTE: During testing it might be useful to implement 'ledgerDbReapply'
>      -- using the code that /does/ do all checks, and throw an exception if
>      -- any of them fail (as this would indicate a bug in the code somewhere).
>      -- We do not have sufficient context in the ledger DB to /verify/ that
>      -- when we reapply a block we are applying it to the correct ledger.
>    , ldbConfReapply :: b -> l -> l
>```

Specifically,
>it will be applied in the very same ledger state, and therefore can't possibly fail.

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Block.hs
```diff
@@ -216,8 +216,8 @@ abEnvToCfg (_, _, dsEnv, pps) =
   gPps        = Update.ProtocolParameters
     { Update.ppScriptVersion    = 0
     , Update.ppSlotDuration     = 0
-    , Update.ppMaxBlockSize     = 748 * pps ^. maxBkSz
-    , Update.ppMaxHeaderSize    = 95 * pps ^. maxHdrSz
+    , Update.ppMaxBlockSize     = 832 * pps ^. maxBkSz
+    , Update.ppMaxHeaderSize    = 569 * pps ^. maxHdrSz
```
erikd, line 220: What are these changes about?

cardano-ledger/src/Cardano/Chain/Block/Validation.hs
erikd: Ok, but after applying a whole bunch of blocks without any validation, it should still check that at the end the chain the hash should still match a fully validated chain.

Probably worth adding a comment about that (if there isn't one already).

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Block.hs
intricate: I'll have to add a comment explaining these. These seem to be the minimum constant multipliers that I can use for size limits such that all tests still pass without returning "block too large" or "block header too large" errors.

intricate: I'll have to add a comment explaining these. These seem to be the minimum constant multipliers that I can use for elaborating abstract size limits such that all tests still pass without returning "block too large" or "block header too large" errors.

In `cardano-ledger-specs`, `Size`s are represented differently than they are in the concrete implementation (where they're represented in actual bytes) and I'm unsure as to how we can accurately convert between the two.

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Update.hs
```diff
@@ -0,0 +1,45 @@
...
+import qualified Cardano.Chain.Common as Concrete
+import qualified Cardano.Chain.Update as Concrete
+
+import qualified Ledger.Update as Abstract
+
+import Test.Cardano.Chain.Genesis.Dummy (dummyProtocolParameters)
+
+elaboratePParams :: Abstract.PParams -> Concrete.ProtocolParameters
+elaboratePParams pps = Concrete.ProtocolParameters
+  -- FIXME @intricate: Fields that I'm not sure about, I'm just using
```
intricate: Actually, I'm not sure that any of the fields I'm copying from `dummyProtocolParameters` are even relevant to the tests that I've added. I can probably just initialize them all to `0` or something.

cardano-ledger/test/Test/Cardano/Chain/UTxO/ValidationMode.hs
```diff
@@ -0,0 +1,359 @@
...
+          (utxo, txIdMap) = elaborateInitialUTxO abstractUtxo
+
+      -- Generate abstract transaction and elaborate.
+      abstractTxWits <- forAll $ genValidTxWits ppau txIdMap
+      let tx = elaborateTxWitsBS
+            (elaborateTxId txIdMap)
+            abstractTxWits
+
+      -- Validate the generated concrete transaction
+      let pm = Dummy.aProtocolMagic
```
intricate, line 81: Will address these `Dummy.aProtocolMagic`s tomorrow.

cardano-ledger/test/Test/Cardano/Chain/Block/Model.hs
```diff
@@ -72,7 +74,8 @@ passConcreteValidation :: MonadTest m => Trace CHAIN -> m ()
 passConcreteValidation tr = void $ evalEither res
  where
   res =
-    foldM (elaborateAndUpdate config) (initialState, txIdMap)
+    -- @intricate: Hardcoded BlockValidate
```
intricate: I think hardcoding `BlockValidate` here is fine. Will remove comments such as this as well.

cardano-ledger/test/Test/Cardano/Chain/UTxO/ValidationMode.hs
```diff
@@ -0,0 +1,359 @@
...
+  initialBalances <- genInitialBalances nm
+  pure $ UTxO.fromBalances initialBalances
+
+-- | Each address should only appear once in the starting balances
+genInitialBalances :: NetworkMagic -> Gen [(Address, Lovelace)]
+genInitialBalances nm = do
+  as <- Gen.list (Range.constant 1 5) (genAddressWithNM nm)
+  addrs <- Gen.filter (not . null) $ Gen.subsequence as
+  let lovelaceRange =
+        Range.constant 156789 -- @intricate
```
intricate: Tagged myself on lines containing this `156789` "magic number" to remind myself to come back and rethink this. `156789` isn't really significant other than the fact that it "just worked" as a minimum bound. Will take a look and improve this generator tomorrow.

dnadales:
I'm a bit confused by this, so sorry for the silly question:

> There should be three scenarios: NewSignal/NewState, OldSignal/NewState, OldSignal/OldState

Do we care about these scenarios? I thought this was about checking whether we would perform validation or not when applying a block.

cardano-ledger/test/Test/Cardano/Chain/UTxO/ValidationMode.hs
dnadales: Great. Whatever constant we choose, we should document why we chose it :D

cardano-ledger/src/Cardano/Chain/Block/ValidationMode.hs
```diff
@@ -0,0 +1,42 @@
...
+
+-- | Indicates what sort of block validation should be performed.
+data BlockValidationMode
+  = BlockValidate
+  -- ^ Perform all block validations.
+  | BlockRevalidate
+  -- ^ Perform no block validations.
+  deriving (Eq, Show)
+
+-- | Perform an action only when in the 'BlockValidate' mode. Otherwise, do
```
dnadales: Why not using `Control.Monad.when` with the predicate `mode == BlockValidate`?

But regardless of this, we are seeing a lot of patterns of the form

```haskell
whenBlockValidate mode (cond `orThrowError` err)
```

Isn't it better to define a function like `check mode cond err`?

Something a bit more controversial could be making the block validation functions and `check` a reader monad, so that we don't have to pass the mode everywhere :man_shrugging:

cardano-ledger/src/Cardano/Chain/Block/ValidationMode.hs
```diff
@@ -0,0 +1,42 @@
...
+
+--------------------------------------------------------------------------------
+-- BlockValidationMode
+--------------------------------------------------------------------------------
+
+-- | Indicates what sort of block validation should be performed.
+data BlockValidationMode
+  = BlockValidate
+  -- ^ Perform all block validations.
+  | BlockRevalidate
```
dnadales: Revalidate here seems to hint at the fact that we're performing the validations again :)

cardano-ledger/src/Cardano/Chain/Block/Validation.hs
ruhatch: :heavy_plus_sign:

cardano-ledger/src/Cardano/Chain/Block/Validation.hs
ruhatch: Just pass the `BlockValidationMode` here I think

cardano-ledger/src/Cardano/Chain/Block/Validation.hs
ruhatch: The validation in that one cannot fail, so I'd say take it out of `MonadError` and just make it return `State`. Then here, it's just `pure $ UPI.registerEpoch ...`

cardano-ledger/src/Cardano/Chain/Block/ValidationMode.hs
ruhatch: Indeed, I think `BlockValidation | NoBlockValidation` might work here

cardano-ledger/src/Cardano/Chain/Block/ValidationMode.hs
ruhatch: In this case, as the priority is readability of the predicates, I think a `MonadReader BlockValidationMode m` would actually work pretty well. You could add an:
```
orThrowErrorBVM :: (MonadError e m, MonadReader BlockValidationMode m) => Bool -> e -> m ()
orThrowErrorBVM condition e = do
  bvm <- ask
  unless (bvm == NoBlockValidation || condition) (throwError e)

infix 1 `orThrowErrorBVM`
```
resulting in e.g.
```
  blockProtocolMagicId b == configProtocolMagicId config
    `orThrowErrorBVM` ChainValidationProtocolMagicMismatch
                        (blockProtocolMagicId b)
                        (configProtocolMagicId config)
```

ruhatch: @dnadales is that what you had in mind too?

cardano-ledger/src/Cardano/Chain/UTxO/ValidationMode.hs
```diff
@@ -0,0 +1,45 @@
...
+
+import Cardano.Prelude
+
+--------------------------------------------------------------------------------
+-- TxValidationMode
+--------------------------------------------------------------------------------
+
+-- | Indicates what sort of transaction validation should be performed.
+data TxValidationMode
+  = TxValidate
```
ruhatch: I think `TxValidation | TxValidationNoCrypto | NoTxValidation` might be clearer

cardano-ledger/src/Cardano/Chain/UTxO/Validation.hs
```diff
@@ -215,25 +235,27 @@ updateUTxOTx env utxo aTx@(Annotated tx _) = do
...
+        `wrapError` UTxOValidationUTxOError
 
-  -- Validate witnesses and their signing addresses
-  mapM_
-      (uncurry $ validateWitness pmi sigData)
-      (zip addresses (V.toList witness))
-    `wrapError` UTxOValidationTxValidationError
+    -- Validate witnesses and their signing addresses
+    mapM_
+        (uncurry $ validateWitness tvmode pmi sigData)
```
ruhatch: I don't think you need to pass the mode here, because you're already using it above

cardano-ledger/src/Cardano/Chain/UTxO/Validation.hs
```diff
@@ -143,16 +149,27 @@ validateTx env utxo (Annotated tx txBytes) = do
...
 -- | Validate that 'TxIn' is in the domain of 'UTxO'
-validateTxIn :: MonadError TxValidationError m => UTxO -> TxIn -> m ()
-validateTxIn utxo txIn
+validateTxIn
+  :: MonadError TxValidationError m
+  => TxValidationMode
+  -> UTxO
+  -> TxIn
+  -> m ()
+validateTxIn TxRevalidateSameState _ _ = pure ()
```
ruhatch: If you're using `TxRevalidateSameState`, you'll never call this function anyway, so you should be able to drop the arg here

cardano-ledger/src/Cardano/Chain/UTxO/Validation.hs
```diff
@@ -143,16 +149,27 @@ validateTx env utxo (Annotated tx txBytes) = do
...
   | txIn `UTxO.member` utxo = pure ()
   | otherwise = throwError $ TxValidationMissingInput txIn
 
 
 -- | Validate the NetworkMagic of a TxOut
 validateTxOutNM
-  :: MonadError TxValidationError m => NetworkMagic -> TxOut -> m ()
-validateTxOutNM nm txOut = do
+  :: MonadError TxValidationError m
+  => TxValidationMode
```
ruhatch: You should be able to drop the arg here too

cardano-ledger/src/Cardano/Chain/UTxO/Validation.hs
ruhatch: In `whenTxValidate` that is

cardano-ledger/test/Test/Cardano/Chain/Block/Model.hs
ruhatch: Yeah, definitely what we want here

ruhatch: You can even push it into `elaborateAndUpdate` considering you're hardcoding it anyway

cardano-ledger/test/Test/Cardano/Chain/Block/Validation.hs
```diff
@@ -146,7 +146,7 @@ foldChainValidationState shouldAssertNF config cvs blocks = S.foldM_
                   )
               NoAssertNF -> pure ()
             updateChainBoundary c bvd
-          ABOBBlock block -> updateBlock config c block
+          ABOBBlock block -> updateBlock bvmode config c block
```
ruhatch: Again, I'd say do the hardcoding here at the `updateBlock` call site and avoid polluting the rest of the logic

cardano-ledger/test/Test/Cardano/Chain/Block/ValidationMode.hs
```diff
@@ -0,0 +1,274 @@
...
+  :: ABlock ByteString
+  -> Gen (ABlock ByteString)
+invalidateABlockProof ab =
+  -- 'Gen.filter' to ensure we don't generate a valid proof
+  Gen.filter (\x -> blockProof x /= blockProof ab) $ do
+    txProof  <- Gen.choice
+      [ pure $ (proofUTxO . blockProof) ab
+      , feedPM genTxProof
+      ]
+    sscProof <- Gen.choice
```
ruhatch: I'd say drop this if you're not using it and change below to be an update, rather than construction

cardano-ledger/test/Test/Cardano/Chain/Block/ValidationMode.hs
```diff
@@ -0,0 +1,274 @@
...
+    dlgProof <- Gen.choice
+      [ pure $ (proofDelegation . blockProof) ab
+      , genAbstractHash (feedPM Delegation.genPayload)
+      ]
+    updProof <- Gen.choice
+      [ pure $ proofUpdate (blockProof ab)
+      , feedPM Update.genProof
+      ]
+    pure $ modifyAProof
+      (\(Annotated _ bs) -> Annotated
```
ruhatch: So `\(Annotated p bs) -> Annotated (p { record update}) bs`

cardano-ledger/test/Test/Cardano/Chain/UTxO/ValidationMode.hs
```diff
@@ -0,0 +1,359 @@
...
+  PParamsAddrsAndUTxO
+    { ppauPParams
+    , ppauAddrs
+    , ppauUTxO
+    } = ppau
+
+  pparams = elaboratePParams ppauPParams
+
+--------------------------------------------------------------------------------
+-- Concrete Generators
```
ruhatch: So as a general comment these generators are going to have really poor shrinking. This is exactly how I defined them the first time round in the abstract generators, but they suffer from using the `Monad Gen` instance, as outlined in Edsko's blog post. I think it would be better to use the abstract generators that I made in `cardano-ledger-specs` as those have good manual shrinking. Please come to me if you're having difficulties adapting those.

cardano-ledger/src/Cardano/Chain/Epoch/Validation.hs
```diff
@@ -105,25 +108,27 @@ validateEpochFile config ll cvs fp = do
...
+  -> Genesis.Config
   -> ChainValidationState
   -> Stream (Of (ABlockOrBoundary ByteString)) (ExceptT ParseError ResIO) ()
   -> ExceptT EpochError ResIO ChainValidationState
-foldChainValidationState config chainValState blocks = S.foldM_
+foldChainValidationState bvmode config chainValState blocks = S.foldM_
   (\cvs block ->
     withExceptT (EpochChainValidationError (blockOrBoundarySlot block))
-      $ updateChainBlockOrBoundary config cvs block
+      $ updateChainBlockOrBoundary bvmode config cvs block
```
ruhatch: I think this is a case where we're only ever going to run it in `BlockValidate` mode, so you can probably hardcode it right here, and remove the parameter passing from the other functions here.

cardano-ledger/src/Cardano/Chain/UTxO/ValidationMode.hs
```diff
@@ -0,0 +1,45 @@
...
+  :: MonadError err m
+  => TxValidationMode
+  -> m ()
+  -> m ()
+whenTxValidate TxValidate action = action
+whenTxValidate _ _ = pure ()
+
+-- | Perform an action only when in either the 'TxValidate' or
+-- 'TxRevalidate' modes. Otherwise, do nothing.
+whenTxValidateOrRevalidate
```
ruhatch: I'm not sure you ever use this one, so maybe get rid of it if you don't need it. Someone else can always define it if we need it later on.

ruhatch (changes requested):
Okay, so quite a few changes here, but the general structure is good.

cardano-ledger/src/Cardano/Chain/Block/ValidationMode.hs
dnadales: @ruhatch that was exactly what I had in mind :man: :zap:  :man:

cardano-ledger/src/Cardano/Chain/UTxO/ValidationMode.hs
intricate: Oops

cardano-ledger/src/Cardano/Chain/UTxO/Validation.hs
intricate: @ruhatch You're right but I wasn't sure if, for some reason, in the future, someone may want to test some of these individual functions with different `ValidationMode`s. Does that make sense or am I perhaps overthinking it?

cardano-ledger/test/Test/Cardano/Chain/Block/ValidationMode.hs
intricate: :man_facepalming: oops.

cardano-ledger/src/Cardano/Chain/Block/ValidationMode.hs
ruhatch: @intricate To make it clearer I think `orThrowErrorInBlockValidationMode` is better. Even though it seems crazy long, it's just more descriptive, so:
```
  blockProtocolMagicId b == configProtocolMagicId config
    `orThrowErrorInBlockValidationMode`
      ChainValidationProtocolMagicMismatch
        (blockProtocolMagicId b)
        (configProtocolMagicId config)

File: cardano-ledger/src/Cardano/Chain/Block/Validation.hs [hunk 21]
```diff
@@ -605,25 +609,26 @@ foldUTxO
   :: UTxO.Environment
   -> UTxO
   -> Stream (Of (ABlock ByteString)) (ExceptT ParseError ResIO) ()
-  -> ExceptT Error ResIO UTxO
+  -> ValidationModeT Error ValidationModeEnv ResIO UTxO
```
erikd: ~Wow, is that type for real???~

I am pretty certain the `ValidationModeEnv` can be dropped from that if you drop the `r` from the `ValidationModeT` type definition.

cardano-ledger/src/Cardano/Chain/Block/Validation.hs
```diff
@@ -605,25 +609,26 @@ foldUTxO
...
 foldUTxO env utxo blocks = S.foldM_
   (foldUTxOBlock env)
   (pure utxo)
   pure
-  (hoist (withExceptT ErrorParseError) blocks)
+  (pure (hoist (withExceptT ErrorParseError) blocks))
```
erikd, line 636: Something about `pure` and `hoist` being used like that strikes me as odd. Its almost as if you need a transformers version of `Control.Monad.join`.

cardano-ledger/src/Cardano/Chain/Epoch/Validation.hs
```diff
@@ -119,14 +122,14 @@ foldChainValidationState
...
-  -> ExceptT EpochError ResIO ChainValidationState
+  -> ValidationModeT EpochError ValidationModeEnv ResIO ChainValidationState
 foldChainValidationState config chainValState blocks = S.foldM_
   (\cvs block ->
     withExceptT (EpochChainValidationError (blockOrBoundarySlot block))
       $ updateChainBlockOrBoundary config cvs block
   )
   (pure chainValState)
-  pure $ hoist (withExceptT EpochParseError) blocks
+  pure (pure (hoist (withExceptT EpochParseError) blocks))
```
erikd, line 131: Like above, `pure . pure . hoist` seems weird.

cardano-ledger/src/Cardano/Chain/UTxO/Validation.hs
```diff
@@ -198,39 +207,45 @@ data UTxOValidationError
...
 
   UTxO.union (S.fromList (NE.toList (txInputs tx)) </| utxo) (txOutputUTxO tx)
     `wrapError` UTxOValidationUTxOError
 
 
 -- | Validate a transaction with a witness and use it to update the 'UTxO'
 updateUTxOTxWitness
-  :: MonadError UTxOValidationError m
+  :: (MonadError UTxOValidationError m, MonadReader ValidationModeEnv m)
   => Environment
```
erikd, line 229: @ruhatch @dnadales Here we have `Environment` being passed in as a parameter and `ValidationModeEnv` being passed as a `MonadReader`. Was this suggested by either of you? If so why? I don't seem the sense of having two bits of data with essentially the same purpose being passed into the function in two such different ways.

cardano-ledger/src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -368,12 +368,11 @@ registerEndorsement env st endorsement = do
...
 -- This corresponds to the @UPIEC@ rules in the Byron ledger specification.
 registerEpoch
-  :: MonadError Error m
-  => Environment
+  :: Environment
   -> State
   -> EpochIndex
   -- ^ Epoch seen on the block.
-  -> m State
+  -> State
```
erikd, line 375: Here, the code used to be able to return an error and now can't? Why is that?

cardano-ledger/src/Cardano/Chain/ValidationMode.hs
```diff
@@ -0,0 +1,70 @@
...
+fromBlockValidationMode bvm = ValidationModeEnv
+  { blockValidationMode = bvm
+  , txValidationMode    = toTxValidationMode bvm
+  }
+
+orThrowErrorBVM
+  :: (MonadError e m, MonadReader ValidationModeEnv m)
+  => Bool
+  -> e
+  -> m ()
```
erikd, line 46: This is beginning to look like the old code base.

cardano-ledger/src/Cardano/Chain/ValidationMode.hs
```diff
@@ -0,0 +1,70 @@
...
+  bvm <- askBVM
+  unless (bvm == NoBlockValidation || condition) (throwError err)
+
+infix 1 `orThrowErrorBVM`
+
+--------------------------------------------------------------------------------
+-- ValidationModeT
+--------------------------------------------------------------------------------
+
+type ValidationModeT e r m a = ExceptT e (ReaderT ValidationModeEnv m) a
```
erikd: If you drop the `r` from that, then I think `ValidationModeT e ValidationModeEnv m a` can be changed to `ValidationModeT e m a` which is a definite improvement.

cardano-ledger/src/Cardano/Chain/Update/Validation/Interface.hs
intricate: https://github.com/input-output-hk/cardano-ledger/pull/548#discussion_r292792803

This function, `registerEpoch`, was never actually returning an error in any case.

cardano-ledger/src/Cardano/Chain/ValidationMode.hs
intricate: That `r` must have been a mistake. Oops.

cardano-ledger/src/Cardano/Chain/UTxO/Validation.hs
intricate: Personally, I think I'd rather drop these `MonadReader` changes and pass the appropriate `*ValidationMode` either as a separate parameter or via an appropriate field within any of the relevant `Environment` data types.

However, I suppose we could also pass the appropriate `Environment` via the `MonadReader`. Not sure what @ruhatch and @dnadales think about this.

dnadales: @erikd see https://github.com/input-output-hk/cardano-ledger/pull/548#discussion_r292799340 for context. Basically the idea was so preserve the readability of the predicates. 

The Environment argument refers to the environment of the operational rules. 

>  I don't seem the sense of having two bits of data with essentially the same purpose being passed into the function in two such different ways.

Is a consequence of this the fact that monad readers should not take any arguments?

But in any case if you don't want to have a monad reader feel free to revert it back. But as far as I understand the validation mode *should not* be in the environment since this is the environment in the spec, and we want it to be as close as possible to the latter.

@ruhatch : thoughts?

intricate: > But as far as I understand the validation mode _should not_ be in the environment since this is the environment in the spec, and we want it to be as close as possible to the latter.

This makes sense. Didn't realize this.

cardano-ledger/src/Cardano/Chain/Block/Validation.hs
```diff
@@ -376,17 +382,18 @@ updateBody env bs b = do
...
   proofUTxO (blockProof b)
     == recoverTxProof (blockTxPayload b)
-    `orThrowError` ChainValidationUTxOProofError
+    `orThrowErrorBVM` ChainValidationUTxOProofError
 
   -- Update the UTxO
+  vModeEnv <- ask
   utxo' <-
-    UTxO.updateUTxO utxoEnv utxo txs
+    runReaderT (UTxO.updateUTxO utxoEnv utxo txs) vModeEnv
```
dnadales: Remember to check this.

cardano-ledger/src/Cardano/Chain/UTxO/Validation.hs
```diff
@@ -198,39 +207,45 @@ data UTxOValidationError
...
   -> m UTxO
 updateUTxOTx env utxo aTx@(Annotated tx _) = do
 
-  validateTx env utxo aTx
-    `wrapError` UTxOValidationTxValidationError
+  tvmode <- askTVM
+
+  whenTxValidationOrNoCrypto tvmode $
+    validateTx env utxo aTx
+      `wrapError` UTxOValidationTxValidationError
```
dnadales: What about:

```haskell
  whenTxValidationOrNoCrypto (validateTx env utxo aTx)
     `wrapError` UTxOValidationTxValidationError

whenTxValidationOrNoCrypto
  :: (MonadError TxValidationError m, MonadReader ValidationModeEnv m)
  => m ()
  -> m ()


```

intricate: Turns out that `wrapError` doesn't play well with the results of our `when*Validation*` functions due to type constraints (see this [comment](https://github.com/input-output-hk/cardano-ledger/pull/548#discussion_r294111601)).

Here's the type sig for `wrapError`:
```
wrapError :: MonadError e' m => Either e a -> (e -> e') -> m a
```

I ended up having to write another version of `wrapError`, `wrapExceptTError` (open to suggestions for a better name) which I'm thinking should be in `cardano-prelude`:
```
wrapExceptTError :: MonadError e' m => ExceptT e m a -> (e -> e') ->  m a
```

I spent some time struggling to find some way of converting the error type in a `MonadError` but fell short. If we could find a nice way to do this, then we could rewrite `wrapError` in a more abstract way.

cardano-ledger/src/Cardano/Chain/Block/Validation.hs
intricate: I now remember why I used `runReaderT` here. `wrapError` accepts an `Either e a` parameter but the type of `updateUTxO` is `(MonadError UTxOValidationError m, MonadReader ValidationModeEnv m) => ... -> m UTxO` (notice the `MonadReader` constraint). There's no `MonadReader` instance for `Either` so, therefore, the result of a call to `updateUTxO` can't be passed to `wrapError`.

I think I was temporarily using `runReaderT` here while I worked out how I wanted to approach this issue.

cardano-ledger/test/Test/Cardano/Chain/UTxO/ValidationMode.hs
intricate: Followed @ruhatch's [advice](https://github.com/input-output-hk/cardano-ledger/pull/548#discussion_r292808302) and removed these concrete generators as well as the property for which they were written, `ts_prop_updateUTxO_Invalid`.

cardano-ledger/src/Cardano/Chain/ValidationMode.hs
```diff
@@ -0,0 +1,134 @@
...
+-- TODO @intricate: This needs to be moved to `cardano-prelude` or something.
+--
+-- 'wrapError' from `cardano-prelude` does not work in our case since it
+-- specifically accepts an 'Either' parameter and there is no 'MonadReader'
+-- instance for 'Either'. Therefore we can't pass it resulting actions from
+-- functions such as 'whenBlockValidation', 'whenTxValidationOrNoCrypto', etc.
+--
+-- If we could find some way to convert the error type in a 'MonadError' then
+-- we could re-write 'wrapError' in a more abstract way.
+wrapExceptTError :: MonadError e' m => ExceptT e m a -> (e -> e') ->  m a
```
intricate: @ruhatch @dnadales: Is it cool if I move this into `cardano-prelude` or do you guys perhaps have a more clever way of doing this (see comment above this function definition)?

cardano-ledger/src/Cardano/Chain/UTxO/Validation.hs
dnadales: I don't want to keep on adding more noise, so ignore my nonsense in favor of what you or @ruhatch think is more sensible.

I'd rather not have a `wrapExceptTError`, maybe `wrapError` should have a more general type in that case.

But if the problem is the type mismatch that you're getting, we could simply write:

```haskell
whenTxValidationOrNoCrypto $ 
    validateTx env utxo aTx  `wrapError` UTxOValidationTxValidationError
```

And regarding 

```haskell
whenTxValidationOrNoCrypto action = do
  tvmode <- askTVM
  case tvmode of
    NoTxValidation -> pure ()
    _              -> action

```

we might want to use `Control.Monad.when` instead of the `case` statement.

cardano-ledger/src/Cardano/Chain/ValidationMode.hs
dnadales: See https://github.com/input-output-hk/cardano-ledger/pull/548#discussion_r294220544

cardano-ledger/src/Cardano/Chain/UTxO/Validation.hs
ruhatch: Like @dnadales said these are two very different kinds of 'environment'. We want one to be very clearly in the types so it looks just like the spec and the other to be implicit in the `Reader`, because it's an optimisation.

cardano-ledger/src/Cardano/Chain/ValidationMode.hs
```diff
@@ -0,0 +1,134 @@
...
+
+import Cardano.Chain.Block.ValidationMode
+  (BlockValidationMode (..), toTxValidationMode)
+import Cardano.Chain.UTxO.ValidationMode (TxValidationMode (..))
+
+--------------------------------------------------------------------------------
+-- ValidationModeEnv
+--------------------------------------------------------------------------------
+
+data ValidationModeEnv = ValidationModeEnv
```
ruhatch: You can definitely just call this `ValidationMode` :)

cardano-ledger/src/Cardano/Chain/ValidationMode.hs
ruhatch: Not all `mtl` is evil ;)

cardano-ledger/src/Cardano/Chain/ValidationMode.hs
```diff
@@ -0,0 +1,134 @@
...
+--------------------------------------------------------------------------------
+
+-- | Perform an action only when in the 'BlockValidation' mode. Otherwise, do
+-- nothing.
+whenBlockValidation
+  :: (MonadError err m, MonadReader ValidationModeEnv m)
+  => m ()
+  -> m ()
+whenBlockValidation action = do
+  bvmode <- askBVM
```
ruhatch: Yeah, this can become:
```
bvmode <- askBVM
when (bvmode == BlockValidation) action
```

cardano-ledger/src/Cardano/Chain/ValidationMode.hs
```diff
@@ -0,0 +1,134 @@
...
+
+-- | Perform an action only when in the 'TxValidation' mode. Otherwise, do
+-- nothing.
+whenTxValidationOrNoCrypto
+  :: (MonadError err m, MonadReader ValidationModeEnv m)
+  => m ()
+  -> m ()
+whenTxValidationOrNoCrypto action = do
+  tvmode <- askTVM
+  case tvmode of
```
ruhatch: And this could be
```
tvmode <- askTVM
unless (tvmode == NoTxValidation) action
```

ruhatch: You could even rename the function to `unlessNoTxValidation`?

cardano-ledger/src/Cardano/Chain/ValidationMode.hs
```diff
@@ -0,0 +1,134 @@
...
+type ValidationModeT e m a = ExceptT e (ReaderT ValidationModeEnv m) a
+
+runValidationModeT
+  :: ValidationModeEnv
+  -> ValidationModeT e m a
+  -> m (Either e a)
+runValidationModeT vme vmt =
+  ((flip runReaderT) vme) . runExceptT $ vmt
+
+askBVM
```
ruhatch: I think I would prefer to put these in full `askBlockValidationMode`, at which point you might say, don't even bother with the helper function!

cardano-ledger/src/Cardano/Chain/ValidationMode.hs
```diff
@@ -0,0 +1,134 @@
...
+  bvm <- askBVM
+  unless (bvm == NoBlockValidation || condition) (throwError err)
+
+infix 1 `orThrowErrorInBlockValidationMode`
+
+--------------------------------------------------------------------------------
+-- ValidationModeT
+--------------------------------------------------------------------------------
+
+type ValidationModeT e m a = ExceptT e (ReaderT ValidationModeEnv m) a
```
ruhatch: I don't think we should define an extra transformer to be honest. It's masking the fact that there is an `ExceptT` inside and it's more obvious to have
```
(`runReaderT` validationMode) . runExceptT
```
as a reader (haha)

cardano-ledger/src/Cardano/Chain/ValidationMode.hs
ruhatch: @intricate what about:
```
wrapErrorWithValidationMode
  :: (MonadError e' m, MonadReader ValidationMode m) 
  => ReaderT ValidationMode (Either e a) 
  -> (e -> e') 
  -> m a
wrapErrorWithValidationMode readerAction = do
  validationMode <- ask
  runReaderT readerAction validationMode >>= \case
    Left err -> throwError $ wrapper err
    Right x -> pure x
```

ruhatch: Let me know if there are any type errors with that, but I think it should work

ruhatch (changes requested):

cardano-ledger/src/Cardano/Chain/ValidationMode.hs
intricate: Alright, dropped the `ValidationModeT` type.

cardano-ledger/src/Cardano/Chain/ValidationMode.hs
intricate: Thanks!

Only had to make a slight change to type sig

```
wrapErrorWithValidationMode
  :: (MonadError e' m, MonadReader ValidationMode m) 
  => ReaderT ValidationMode (Either e) a 
  -> (e -> e') 
  -> m a
```

ruhatch (approved):
Looks really good I think - worth several rounds of review :)

PR #549, opened 2019-06-12
[#530] Introduce function to verify whether a header matches a block body

### LINKED ISSUE #530: Function to verify whether a header matches a block (or block body)
In the consensus layer we download headers separately from blocks, and then need to verify that the two line up. Ideally, we'd like a function that takes a header and a block (or block body) and verifies that the two do in fact line up. For Byron, this should verify all the "proofs" that are in the header (delegation payload, transaction payload, etc.) For Shelley I guess we should just have a single "body hash" in the header.

intricate:
Closes #530.

### DIFF
--- a/cardano-ledger/src/Cardano/Chain/Block/Proof.hs
+++ b/cardano-ledger/src/Cardano/Chain/Block/Proof.hs
@@ -5,6 +5,7 @@
 
 module Cardano.Chain.Block.Proof
   ( Proof(..)
+  , ProofValidationError (..)
   , mkProof
   , recoverProof
   )
@@ -70,3 +71,14 @@ recoverProof body = Proof
   , proofDelegation = hashDecoded $ bodyDlgPayload body
   , proofUpdate     = Update.recoverProof $ bodyUpdatePayload body
   }
+
+-- | Error which can result from attempting to validate an invalid payload
+-- proof.
+data ProofValidationError
+  = DelegationProofValidationError
+  -- ^ The delegation payload proof did not match
+  | UTxOProofValidationError
+  -- ^ The UTxO payload proof did not match
+  | UpdateProofValidationError
+  -- ^ The update payload proof did not match
+  deriving (Eq, Show)
--- a/cardano-ledger/src/Cardano/Chain/Block/Validation.hs
+++ b/cardano-ledger/src/Cardano/Chain/Block/Validation.hs
@@ -72,7 +72,7 @@ import Cardano.Chain.Block.Header
   , headerSlot
   , wrapBoundaryBytes
   )
-import Cardano.Chain.Block.Proof (Proof(..))
+import Cardano.Chain.Block.Proof (Proof(..), ProofValidationError (..))
 import Cardano.Chain.Common
   ( BlockCount(..)
   , KeyHash
@@ -253,9 +253,6 @@ data ChainValidationError
   | ChainValidationInvalidSignature BlockSignature
   -- ^ The signature of the block is invalid
 
-  | ChainValidationDelegationProofError
-  -- ^ The delegation proof does not correspond to the delegation payload
-
   | ChainValidationDelegationSchedulingError Scheduling.Error
   -- ^ A delegation certificate failed validation in the ledger layer
 
@@ -271,14 +268,11 @@ data ChainValidationError
   | ChainValidationUpdateError UPI.Error
   -- ^ Something failed to register in the update interface
 
-  | ChainValidationUpdateProofError
-  -- ^ The update payload proof did not match
-
   | ChainValidationUTxOValidationError UTxO.UTxOValidationError
   -- ^ A transaction failed validation in the ledger layer
 
-  | ChainValidationUTxOProofError
-  -- ^ The UTxO payload proof did not match
+  | ChainValidationProofValidationError ProofValidationError
+  -- ^ A payload proof did not match.
 
   deriving (Eq, Show)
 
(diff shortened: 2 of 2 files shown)


cardano-ledger/src/Cardano/Chain/Block/Validation.hs
```diff
@@ -333,6 +328,42 @@ updateChainBoundary cvs bvd = do
     }
 
 
+validateHeaderBodyProofs
```
intricate: Any better ideas for a name here?

ruhatch: At a high level we're checking that the header matches the body somehow - so we could say `validateHeaderMatchesBody`?

ruhatch: That would probably capture what Edsko was looking for :)

ruhatch (approved):
Looks great, thanks :)

cardano-ledger/src/Cardano/Chain/Block/Proof.hs
```diff
@@ -70,3 +71,14 @@ recoverProof body = Proof
   , proofDelegation = hashDecoded $ bodyDlgPayload body
   , proofUpdate     = Update.recoverProof $ bodyUpdatePayload body
   }
+
+-- | Error which can result from attempting to validate an invalid payload
```
dnadales, line 75: I wonder whether we could omit the `Proof` prefix and use it qualified. I don't know what's the approach we took in the other modules. 

I also wonder whether we can drop the `ValidationError` suffix from the error types.

cardano-ledger/src/Cardano/Chain/Block/Validation.hs
```diff
@@ -271,14 +269,11 @@ data ChainValidationError
...
 
-  | ChainValidationUpdateProofError
-  -- ^ The update payload proof did not match
-
   | ChainValidationUTxOValidationError UTxO.UTxOValidationError
   -- ^ A transaction failed validation in the ledger layer
 
-  | ChainValidationUTxOProofError
-  -- ^ The UTxO payload proof did not match
+  | ChainValidationProofValidationError ProofValidationError
```
dnadales, line 275: Probably not something that we want to change (neither in this PR nor ever), but I wonder whether this would have looked better:

```haskell
| Proof Proof.ValidationError
```

cardano-ledger/src/Cardano/Chain/Block/Validation.hs
```diff
@@ -333,6 +328,42 @@ updateChainBoundary cvs bvd = do
...
+validateHeaderMatchesBody
+  :: MonadError ProofValidationError m
+  => AHeader ByteString
+  -> ABody ByteString
+  -> m ()
+validateHeaderMatchesBody hdr body = do
+  let hdrProof = headerProof hdr
+
+  -- Validate the delegation payload signature
+  proofDelegation hdrProof
```
dnadales: Just curious about why the line breaks? Didn't the two sides of the equality fit in 80 lines?

dnadales (approved):

cardano-ledger/src/Cardano/Chain/Block/Validation.hs
intricate: Perhaps so. I was originally considering going with something such as `Proof.ValidationError` or `Proof.Error` :man_shrugging:

cardano-ledger/src/Cardano/Chain/Block/Validation.hs
intricate: Ah, I just copy and pasted this from within the function, `updateBody`, below. However, you're right, it does fit within 80 chars. Can make this change.

cardano-ledger/src/Cardano/Chain/Block/Proof.hs
intricate: It would be cool if we could decide on how we'd generally prefer to name our errors. It seems to be done in a few different ways throughout the repo. 

For example, this is from `Cardano.Chain.Delegation.Validation.Scheduling`:
```
data Error

  = InvalidCertificate
  -- ^ The delegation certificate has an invalid signature

  | MultipleDelegationsForEpoch EpochIndex KeyHash
  -- ^ This delegator has already delegated for the given epoch

  | MultipleDelegationsForSlot FlatSlotId KeyHash
  -- ^ This delegator has already delgated in this slot

  | NonGenesisDelegator KeyHash
  -- ^ This delegator is not one of the allowed genesis keys

  | PastEpoch EpochIndex EpochIndex
  -- ^ This delegation is for a past epoch

  deriving (Eq, Show)
```

and this is from `Cardano.Chain.UTxO.Validation`:
```
-- | A representation of all the ways a transaction might be invalid
data TxValidationError
  = TxValidationLovelaceError Text LovelaceError
  | TxValidationFeeTooSmall Tx Lovelace Lovelace
  | TxValidationInvalidWitness TxInWitness
  | TxValidationMissingInput TxIn
  | TxValidationNetworkMagicMismatch NetworkMagic NetworkMagic
  -- ^ Fields are <expected> <actual>
  | TxValidationTxTooLarge Natural Natural
  | TxValidationUnknownAddressAttributes
  | TxValidationUnknownAttributes
  deriving (Eq, Show)
```

cardano-ledger/src/Cardano/Chain/Block/Validation.hs
dnadales: Yeah. But like I said, we might want to address this in the future (or just leave it be :) )

cardano-ledger/src/Cardano/Chain/Block/Proof.hs
dnadales: Yeah. We might want to make an issue out of this. CC: @ruhatch

PR #550, opened 2019-06-12
[#170] Rename FlatSlotId and EpochIndex

ruhatch:
This is the first PR of #170 that does a couple of simple renamings and should be fairly uncontroversial


cardano-ledger/cardano-ledger.cabal
```diff
@@ -84,7 +84,7 @@ library
                        Cardano.Chain.Genesis.NonAvvmBalances
                        Cardano.Chain.Genesis.Spec
 
-                       Cardano.Chain.Slotting.EpochIndex
+                       Cardano.Chain.Slotting.EpochNo
```
dnadales: Probably a bit late, and I think there are other packages that also use `EpochNo` and `SlotNo`, but to me these read as "epoch no" and "slot no" :) I just wonder why didn't we choose something like `EpochNumber` or `EpochNr` :man_shrugging:

dnadales (approved):
:wave: `FlatSlotId` :tada:

LFUTM (looks fairly uncontroversial to me) ;)

cardano-ledger/cardano-ledger.cabal
ruhatch: No, you're right, and [here's](https://chrisdone.com/posts/german-naming-convention/) the blog post to prove it. I will update this :)

PR #551, opened 2019-06-12
Update cabal.project

erikd:


rvl (commented):
The commit IDs for cardano-prelude and cardano-base don't match what's in stack.yaml.

Do they need to be in sync?

If they need to be in sync, what options do we have for keeping the information in a single place? Could we delete stack.yaml? Or could we generate cabal.project from stack.yaml? (This would be possible by modifying stack-to-nix and generating cabal.project from the Nix files which are generated from stack.yaml).

erikd:
They don't necessarily *have* to be in sync, but its probably a good idea if they were.

Unfortunately some people prefer `stack` and some `cabal` so we need to keep both. Yes it would be good to have a tool to keep these in sync.

rvl (approved):

PR #552, opened 2019-06-13
[#170] Rename SlotId and remove LocalSlotIndex

### LINKED ISSUE #170: Give `SlotId` a more specific name, and `FlatSlotId` a better name.

ruhatch:
Instead of using `LocalSlotIndex` to represent the number of slots into an epoch a slot is, we use `SlotCount`. This has two benefits, the first, expanding the size to a `Word64`, which is the same size as the `EpochSlots` type, and second removing an extra unnecessary `newtype`.

Closes #170


cardano-ledger/src/Cardano/Chain/Block/Validation.hs
```diff
@@ -511,7 +511,7 @@ epochTransition env st slot = if nextEpoch > currentEpoch
   EpochEnvironment { protocolMagic, k, numGenKeys, delegationMap, currentEpoch }
     = env
 
-  nextEpoch = siEpoch $ unflattenSlotId (kEpochSlots k) slot
+  nextEpoch = slotNumberEpoch (kEpochSlots k) slot
```
dnadales, line 514: Nice. This looks more readable!

cardano-ledger/src/Cardano/Chain/Slotting/SlotNumber.hs
```diff
@@ -0,0 +1,70 @@
...
+
+import Cardano.Binary (FromCBOR(..), ToCBOR(..))
+import Cardano.Chain.Common.BlockCount (BlockCount, unBlockCount)
+import Cardano.Chain.Slotting.SlotCount (SlotCount(..))
+
+
+-- | 'SlotNumber' is an absolute slot number from the beginning of time
+--
+--   'SlotNumber' is held in a 'Word64'. Assuming a slot every 20 seconds, 'Word64'
+--   is sufficient for slot indices for 10^13 years.
```
dnadales, line 32: > is sufficient for slot indices for 10^13 years.

Right, let people in the future fix this ... o wait, the sun will die earlier than that. Oh, well, I guess we'll have bigger problems by then ...

But seriously, it's nice that you put that comment :)

cardano-ledger/src/Cardano/Chain/Slotting/SlotNumber.hs
```diff
@@ -0,0 +1,70 @@
...
+
+instance MonadError SchemaError m => FromJSON m SlotNumber where
+  fromJSON val = do
+    number <- fromJSON val
+    pure $ SlotNumber number
+
+instance B.Buildable SlotNumber where
+  build s = bprint int (unSlotNumber s)
+
+-- | Increase a 'SlotNumber' by 'SlotCount'
```
dnadales, line 56: Just some musings...

I'm just wondering why don't we call this `addSlotCount` (and reversing the arguments), since we're adding a `SlotCount` to a `SlotNumber` and returning a `SlotNumber`.

I'm also wondering whether it'd make sense to define a type-class with an operator, say `+.`, for types in which you can add a `Count` to a `Number`. Otherwise we'll have similar functions for `EpochCount` and `EpochNumber` ...

cardano-ledger/src/Cardano/Chain/Slotting/SlotNumber.hs
```diff
@@ -0,0 +1,70 @@
...
+addSlotNumber :: SlotCount -> SlotNumber -> SlotNumber
+addSlotNumber (SlotCount a) (SlotNumber b) = SlotNumber $ a + b
+
+-- | Decrease a 'SlotNumber' by 'SlotCount', going no lower than 0
```
dnadales, line 60: Are we ok with ignoring the fact we're subtracting a higher slot count? My guess is that we are: returning a maybe here might be a pain in the posterior for the code that uses this.

cardano-ledger/src/Cardano/Chain/Slotting/SlotNumber.hs
```diff
@@ -0,0 +1,70 @@
...
+subSlotNumber :: SlotCount -> SlotNumber -> SlotNumber
+subSlotNumber (SlotCount a) (SlotNumber b) =
+  if a > b then SlotNumber 0 else SlotNumber (b - a)
+
+-- | Compute the number of slots after which a block becomes stable as @2 * k@,
```
dnadales, line 65: I think I might be the responsible for this `twice` function. I'm wondering whether it'd be better to have a function `blockCountToSlotCount:: BlockCount -> SlotCount` (or something like that), and then convert the `SlotCount` to a `SlotNumber`.

dnadales: This would mean that instead of `twice x` we would have to write 

```haskell
slotCountToSlotNumber . blockCountToSlotCount $ x`
```

or define `blockCountToSlotNumber = slotCountToSlotNumber . blockCountToSlotCount` and replace `twice x` by

```haskell
blockCountToSlotNumber x 
```

I don't know what's better though ... :rabbit2: :hole:

cardano-ledger/test/Test/Cardano/Chain/Slotting/Properties.hs
```diff
@@ -8,214 +8,60 @@ where
...
-  _    <- pure $ unflattenSlotId sc fsId
+  _    <- pure $ fromSlotNumber sc fsId
   success
 
--- Check that `unflattenSlotId . flattenSlotId == id`.
-ts_prop_unflattenFlattenSlotId :: TSProperty
-ts_prop_unflattenFlattenSlotId = withTestsTS 100 . property $ do
-  (sId, sc) <- forAll genConsistentSlotIdEpochSlots
-  sId === unflattenSlotId sc (flattenSlotId sc sId)
+-- Check that `fromSlotNumber . toSlotNumber == id`.
```
dnadales, line 44: Maybe we need a different name for this test.

dnadales (approved):
Nice that we're getting rid of these types :smile:

cardano-ledger/src/Cardano/Chain/Slotting/SlotNumber.hs
ruhatch: It was already there, but it's definitely one of the most informative comments I've found

cardano-ledger/src/Cardano/Chain/Slotting/SlotNumber.hs
ruhatch: That seems reasonable, yeah.

I don't think it's worth the abstraction, because we don't actually have or use `EpochCount`!

cardano-ledger/src/Cardano/Chain/Slotting/SlotNumber.hs
ruhatch: I think so!

cardano-ledger/src/Cardano/Chain/Slotting/SlotNumber.hs
ruhatch: When you implemented, did you see `kSlotSecurityParam`? That's what is supposed to do the translation. It's not a general `BlockCount` to `SlotCount` conversion.

dnadales: No, I didn't. So maybe we can remove `twice` :/ Sorry for that.

PR #555, opened 2019-06-18
[#503] Audit generators for use of Monad instance

### LINKED ISSUE #503: Audit Generators for use of `Monad` instance
## Background

@edsko published a [blog post](http://www.well-typed.com/blog/2019/05/integrated-shrinking/) discussing integrated shrinking and specifically why a `Monad` instance for `Gen` is a bad idea.

## Proposed Solution

Audit the generators for uses of the `Monad` interface and either replace them with applicative, if possible, or manual shrinking where necessary.

## Next Actions
- [ ] Audit codebase for generators using the Monad interface
- [ ] Update generators to use applicative or manual shrinking

ruhatch:
Closes #503

### DIFF
--- a/cardano-ledger/test/Test/Cardano/Chain/UTxO/Gen.hs
+++ b/cardano-ledger/test/Test/Cardano/Chain/UTxO/Gen.hs
@@ -123,10 +123,10 @@ genTxOutList :: Gen (NonEmpty TxOut)
 genTxOutList = Gen.nonEmpty (Range.linear 1 100) genTxOut
 
 genUTxOConfiguration :: Gen UTxOConfiguration
-genUTxOConfiguration = do
-  limit <- Gen.int (Range.constant 0 200)
-  addrs <- Gen.list (Range.linear 0 50) genAddress
-  return (UTxOConfiguration limit (S.fromList addrs))
+genUTxOConfiguration =
+  UTxOConfiguration
+    <$> Gen.int (Range.constant 0 200)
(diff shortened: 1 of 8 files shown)


cardano-ledger/test/Test/Cardano/Chain/UTxO/Gen.hs
```diff
@@ -123,10 +123,10 @@ genTxOutList :: Gen (NonEmpty TxOut)
 genTxOutList = Gen.nonEmpty (Range.linear 1 100) genTxOut
 
 genUTxOConfiguration :: Gen UTxOConfiguration
-genUTxOConfiguration = do
-  limit <- Gen.int (Range.constant 0 200)
-  addrs <- Gen.list (Range.linear 0 50) genAddress
-  return (UTxOConfiguration limit (S.fromList addrs))
+genUTxOConfiguration =
+  UTxOConfiguration
+    <$> Gen.int (Range.constant 0 200)
```
dnadales, line 128: Why did we go from `linear` to `constant`?

dnadales (approved):

cardano-ledger/test/Test/Cardano/Chain/UTxO/Gen.hs
ruhatch: We didn't that's a different `Range` ;)

dnadales: :man_facepalming:

PR #560, opened 2019-06-21
move unused functions from Cardano.Crypto.Limits.hs to Test.Cardano.Crypto.Limits.hs

DeepakKapiswe:
Drop `Cardano.Crypto.Limits.hs`
Addresses https://github.com/input-output-hk/cardano-base/pull/25


ruhatch (approved):
LGTM!

crypto/test/Test/Cardano/Crypto/Limits.hs
```diff
@@ -30,6 +36,38 @@ import Test.Cardano.Crypto.Gen
 tests :: IO Bool
 tests = checkParallel $$discover
 
+---------------------------------------------------------------------------
+-- Limit
+---------------------------------------------------------------------------
+-- | A limit on the length of something (in bytes).
```
dnadales, line 42: Maybe "a limit on the length of a data value"?

crypto/test/Test/Cardano/Crypto/Limits.hs
```diff
@@ -30,6 +36,38 @@ import Test.Cardano.Crypto.Gen
...
+-- Helper Functions
+--------------------------------------------------------------------------------
+
+mlAbstractHash
+  :: forall algo a . HashAlgorithm algo => Limit (AbstractHash algo a)
+mlAbstractHash =
+  fromIntegral (hashDigestSize (panic "AbstractHash limit" :: algo) + 4)
+
+mlVerificationKey :: Limit VerificationKey
+mlVerificationKey = 66
```
dnadales, line 64: We might want to explain why did we chose these constants.

crypto/test/Test/Cardano/Crypto/Limits.hs
```diff
@@ -30,6 +36,38 @@ import Test.Cardano.Crypto.Gen
...
+  fmap _ (Limit x) = Limit x
+
+--------------------------------------------------------------------------------
+-- Helper Functions
+--------------------------------------------------------------------------------
+
+mlAbstractHash
+  :: forall algo a . HashAlgorithm algo => Limit (AbstractHash algo a)
+mlAbstractHash =
+  fromIntegral (hashDigestSize (panic "AbstractHash limit" :: algo) + 4)
```
dnadales, line 61: Might be worth explaining why did we add a `panic` here. And why not just using `undefined`?

crypto/test/Test/Cardano/Crypto/Limits.hs
```diff
@@ -30,6 +36,38 @@ import Test.Cardano.Crypto.Gen
 tests :: IO Bool
 tests = checkParallel $$discover
 
+---------------------------------------------------------------------------
+-- Limit
+---------------------------------------------------------------------------
+-- | A limit on the length of something (in bytes).
+--   TODO should check for overflow in the Num instance.
```
dnadales, line 43: Hmmm, I'm not sure about this comment. I'd rather say that we support maximum sizes equal to `maxBound :: Word32`. Maybe we would like to clarify that `maxBound` bytes is about 4GB.

dnadales (commented):
Looks OK to me, but I think we would want to add more comments, specially around the magic constants and `panic` usages.

Since I don't have ownership on this codebase I'd feel more comfortable if @erikd or @ruhatch could take a quick look.

ruhatch (commented):
bors r+

PR #566, opened 2019-07-03
Changes for epoch boundary integration

nc6:
Unfortunately we do need to serialise boundary blocks, because we need to create the initial one for testing.


dnadales (approved):

cardano-ledger/src/Cardano/Chain/Block/Boundary.hs
```diff
@@ -57,6 +52,4 @@ dropBoundaryBody = dropList dropBytes
 --------------------------------------------------------------------------------
 
 dropBoundaryExtraBodyData :: Dropper s
-dropBoundaryExtraBodyData = do
-  enforceSize "BoundaryExtraBodyData" 1
-  dropAttributes
+dropBoundaryExtraBodyData = dropAttributes
```
intricate: **Related to `ts_prop_mainnetEpochsValid` and `prop_deserializeEpochs` failures:**

We need to `enforceSize 1` prior to calling `dropAttributes` here.

In `cardano-sl`, `deriveSimpleBi` was used to define the `Bi` instance for the `GenesisExtraBodyData` (EBB extra body data): https://github.com/input-output-hk/cardano-sl/blob/34337f53764bc558396be8bc146699ba9d6e6f4e/chain/src/Pos/Chain/Block/Genesis.hs#L203

`deriveSimpleBi` automatically `encodeListLen numFields` for each constructor before encoding the actual values of each field: https://github.com/input-output-hk/cardano-sl/blob/e74bd61d6e497533b0ac9ea5272211bbf4665bd3/binary/src/Pos/Binary/Class/TH.hs#L493

cardano-ledger/src/Cardano/Chain/Block/Header.hs
```diff
@@ -438,17 +438,17 @@ headerHashAnnotated = hashDecoded . fmap wrapHeaderBytes
 -- BoundaryHeader
 --------------------------------------------------------------------------------
 
-dropBoundaryHeader :: Decoder s (HeaderHash, Word64)
+dropBoundaryHeader :: Decoder s (HeaderHash, Word64, ChainDifficulty)
 dropBoundaryHeader = do
   enforceSize "BoundaryHeader" 5
   dropInt32
```
intricate, line 444: **Related to `ts_roundTripBoundaryBlock` failure:**

It seems that decoding and dropping the `ProtocolMagicId :: Word32` here as an `Int32` can fail in certain circumstances. 

Apparently, decoding will fail when attempting to `dropInt32` (to discard the encoded `ProtocolMagicId :: Word32`) if the encoded `ProtocolMagicId :: Word32` is greater than the `Int32` max bound.

The issue is that we're trying to decode and drop an `Int32` (what `ProtocolMagicId` used to be in `cardano-sl`), but the `ProtocolMagicId` is a `Word32` in `cardano-ledger`. Therefore, our round-trip will end up failing as, in our `ProtocolMagicId` generator, we're generating values within the `Word32` range which can potentially exceed that of the `Int32` range.

Initially, I would have expected that the decoder would have just interpreted the bytes as an `Int32` and there'd be no problem, but apparently not.

intricate (changes requested):

intricate:
I pushed a commit that addresses https://github.com/input-output-hk/cardano-ledger/pull/566#discussion_r302318552. 

However, I didn't address the error in which `ts_roundTripBoundaryBlock` fails. I wasn't too sure how we might want to approach this. If we want to appropriately test compatibility with `cardano-sl`, then there could likely be negative `ProtocolMagicId :: Int32`s encoded in blocks (not on mainnet but perhaps on some testnet?). 

If we don't really care too much about handling this negative case, we could just ensure that our `ProtocolMagicId` generator only generates `Word32` values between 0 and the max `Int32` and this would allow the round-trip test to pass.

avieth:
> Unfortunately we do need to serialise boundary blocks, because we need to create the initial one for testing.

And also because we need to store-and-forward them from/to the Byron net.

intricate (approved):

PR #569, opened 2019-07-08
Add ToCBOR and FromCBOR instances for ChainValidationState

mrBliss:
This requires `ToCBOR` and `FromCBOR` instances for a whole bunch of other types
too.

The consensus storage layer needs these instances in order to write/read snapshots of the ledger state to/from disk.

Roundtrip property tests and golden tests still need to be written. Can somebody of the ledger team do this? Feel free to push more commits to this PR.

 I tried to adhere to the existing style as much as I could, but I might have made some mistakes.

### DIFF
--- a/cardano-ledger/src/Cardano/Chain/Block/Validation.hs
+++ b/cardano-ledger/src/Cardano/Chain/Block/Validation.hs
@@ -42,11 +42,19 @@ import qualified Data.ByteString.Lazy as BSL
 import Data.Coerce (coerce)
 import qualified Data.Map.Strict as M
 import Data.Sequence (Seq(..), (<|))
+import qualified Data.Sequence as Seq
 import Formatting.Buildable (Buildable)
 import Streaming (Of(..), Stream, hoist)
 import qualified Streaming.Prelude as S
 
-import Cardano.Binary (Annotated(..), serialize')
+import Cardano.Binary
+  ( Annotated(..)
+  , FromCBOR(..)
+  , ToCBOR(..)
+  , encodeListLen
+  , enforceSize
+  , serialize'
+  )
 import Cardano.Chain.Block.Body (ABody (..))
 import Cardano.Chain.Block.Block
   ( ABlock(..)
@@ -130,6 +138,21 @@ data SigningHistory = SigningHistory
   , shKeyHashCounts :: !(Map KeyHash BlockCount)
   } deriving (Eq, Show, Generic, NFData)
 
+instance FromCBOR SigningHistory where
+  fromCBOR = do
+    enforceSize "SigningHistory" 3
+    SigningHistory
+      <$> fromCBOR
+      <*> (Seq.fromList <$> fromCBOR)
+      <*> fromCBOR
+
+instance ToCBOR SigningHistory where
+  toCBOR sh =
+    encodeListLen 3
+      <> toCBOR (shK sh)
+      <> toCBOR (toList (shSigningQueue sh))
+      <> toCBOR (shKeyHashCounts sh)
+
 -- | Update the `SigningHistory` with a new signer, removing the oldest value if
 --   the sequence is @K@ blocks long
 updateSigningHistory :: VerificationKey -> SigningHistory -> SigningHistory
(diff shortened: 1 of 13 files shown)


cardano-ledger/src/Cardano/Chain/Block/Validation.hs
```diff
@@ -130,6 +138,21 @@ data SigningHistory = SigningHistory
   , shKeyHashCounts :: !(Map KeyHash BlockCount)
   } deriving (Eq, Show, Generic, NFData)
 
+instance FromCBOR SigningHistory where
+  fromCBOR = do
+    enforceSize "SigningHistory" 3
+    SigningHistory
+      <$> fromCBOR
+      <*> (Seq.fromList <$> fromCBOR)
```
mrBliss, line 146: I go through a list here because there are no `ToCBOR` and `FromCBOR` instances for `Seq`.

cardano-ledger/src/Cardano/Chain/Common/Compact.hs
```diff
@@ -32,6 +32,12 @@ newtype CompactAddress = CompactAddress ShortByteString
   deriving newtype HeapWords
   deriving anyclass NFData
 
+instance FromCBOR CompactAddress where
+  fromCBOR = CompactAddress . BSS.toShort <$> fromCBOR
```
mrBliss, line 36: I go through a `ByteString` here because there are no `ToCBOR` and `FromCBOR` instances for `ShortByteString`.

cardano-ledger/src/Cardano/Chain/Delegation/Map.hs
```diff
@@ -33,6 +34,11 @@ newtype Map = Map
   } deriving (Eq, Show, Generic)
     deriving anyclass NFData
 
+instance FromCBOR Map where
+  fromCBOR = Map . Bimap.fromList <$> fromCBOR
```
mrBliss, line 38: I go through a `Map` here because there are no `ToCBOR` and `FromCBOR` instances for `Bimap`.

cardano-ledger/src/Cardano/Chain/Delegation/Validation/Scheduling.hs
```diff
@@ -50,12 +57,40 @@ data State = State
   , keyEpochDelegations  :: !(Set (EpochNumber, KeyHash))
   } deriving (Eq, Show, Generic, NFData)
 
+instance FromCBOR State where
+  fromCBOR = do
+    enforceSize "State" 2
+    State
+      <$> (Seq.fromList <$> fromCBOR)
```
mrBliss, line 64: I go through a list here because there are no `ToCBOR` and `FromCBOR` instances for `Seq`.

dnadales (approved):
LGTM

dnadales:
I've created [this issue](https://github.com/input-output-hk/cardano-ledger/issues/570) to add the tests.

cardano-ledger/src/Cardano/Chain/Common/Compact.hs
intricate: Seems sensible. I imagine it should be encoded/decoded in the same way as a regular `ByteString` anyways.

cardano-ledger/src/Cardano/Chain/Block/Validation.hs
```diff
@@ -175,6 +198,27 @@ data ChainValidationState = ChainValidationState
...
+      <$> fromCBOR
+      <*> fromCBOR
+      <*> fromCBOR
+      <*> fromCBOR
+      <*> fromCBOR
+      <*> fromCBOR
+
+instance ToCBOR ChainValidationState where
+  toCBOR c =
+    encodeListLen 7
```
intricate: Why `encodeListLen 7` when we're only encoding 6 values? Same question goes for the `enforceSize "ChainValidationState" 7` call above.

cardano-ledger/src/Cardano/Chain/Delegation/Validation/Scheduling.hs
```diff
@@ -50,12 +57,40 @@ data State = State
...
+  fromCBOR = do
+    enforceSize "ScheduledDelegation" 3
+    ScheduledDelegation
+      <$> fromCBOR
+      <*> fromCBOR
+      <*> fromCBOR
+
+instance ToCBOR ScheduledDelegation where
+  toCBOR sd =
+    encodeListLen 2
```
intricate: Did you mean, `encodeListLen 3`?

intricate (changes requested):
Looks good. Just a couple of concerns with what seem to be incorrect parameter values for some `enforceSize` and `encodeListLen` calls.

cardano-ledger/src/Cardano/Chain/Block/Validation.hs
mrBliss: > Why `encodeListLen 7` when we're only encoding 6 values? Same question goes for the `enforceSize "ChainValidationState" 7` call above.

Counting is hard :man_facepalming:

cardano-ledger/src/Cardano/Chain/Delegation/Validation/Scheduling.hs
mrBliss: > Did you mean, `encodeListLen 3`?

Indeed, this is a copy-paste error. Another argument for using generics for this :slightly_smiling_face:

mrBliss: About `ShortByteString`: https://hackage.haskell.org/package/serialise-0.2.1.0/docs/src/Codec.Serialise.Class.html#line-358 made me think that maybe a more efficient implementation is possible.

mrBliss:
> Looks good. Just a couple of concerns with what seem to be incorrect parameter values for some `enforceSize` and `encodeListLen` calls.

Thanks for going over this. Those bugs are what you get when you don't test your code. I wouldn't wait too long with #570  :wink:

mrBliss:
Pushed my fixes for the bugs @intricate noticed.

intricate:
Note: I pushed a commit that includes a CBOR round-trip test for `SigningHistory`. There are still some more round-trips and goldens to implement.

intricate (approved):
LGTM. Just need to ensure we add round-trip and golden tests (issue #570) ASAP.

PR #571, opened 2019-07-12
Update `cardano-shell`, plus small Nix changes

deepfire:


erikd (changes requested):
Can we squash all the Nix related commits into a single one?

deepfire:
> Can we squash all the Nix related commits into a single one?

@erikd, done.

erikd (approved):

erikd:
Would still be nice to clean up the `RecordWildCard` related commit (ie squash it into one one the others)

deepfire:
Done as well!

PR #573, opened 2019-07-15
Add a specific tag for genesis block

nc6:
This is a little ugly. The ouroboros-consensus demo does not necessarily start in epoch 0 (it does this in order to easily synchronise time). This means we have a "genesis" EBB in a non-0 epoch. In order to round-trip the serialisation here correctly, we add an extra tag into the attributes which encodes the presence of a genesis hash.

This only applies on non-0 epochs, so should not be invoked by any existing code - these will only be produced in the demo, and everything else should carry on regardless.

### DIFF
--- a/cardano-ledger/src/Cardano/Chain/Block/Block.hs
+++ b/cardano-ledger/src/Cardano/Chain/Block/Block.hs
@@ -8,7 +8,6 @@
 {-# LANGUAGE TypeApplications     #-}
 {-# LANGUAGE TypeFamilies         #-}
 {-# LANGUAGE TypeOperators        #-}
-{-# LANGUAGE TypeSynonymInstances #-}
 
 module Cardano.Chain.Block.Block
   (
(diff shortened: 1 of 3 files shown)


cardano-ledger/src/Cardano/Chain/Block/Block.hs
```diff
@@ -429,6 +429,9 @@ toCBORABOBBoundary pm bvd =
 toCBORBoundaryBlock :: ProtocolMagicId -> BoundaryValidationData a -> Encoding
 toCBORBoundaryBlock pm bvd = let
     bodyProof = hash (mempty :: LByteString)
+    genesisTag = case (boundaryPrevHash bvd, boundaryEpoch bvd) of
```
dnadales, line 435: Maybe some comments on the magic number would help :man_shrugging:

cardano-ledger/src/Cardano/Chain/Block/Boundary.hs
```diff
@@ -15,7 +16,10 @@ import Data.Word (Word64)
 
 import Cardano.Binary
   (Decoder, Dropper, decodeWord64, dropBytes, dropList, enforceSize, fromCBOR)
-import Cardano.Chain.Common (ChainDifficulty, dropAttributes)
+import Cardano.Chain.Common 
+  (ChainDifficulty, attrData, dropAttributes, fromCBORAttributes)
+import Cardano.Prelude
```
dnadales, line 21: The symbols in `Cardano.Prelude` are usually imported implicitly.

dnadales (approved):

Issue #536, opened 2019-06-03
Add Update mechanism to conformance tests


ruhatch:
## Background

The conformance tests currently include Delegation and UTxO payload, but not the update mechanism. We need to wait for reasonable valid generators for the update mechanism to be added before we can elaborate and tests chains including them.

## Proposed Solution

Once generators for update mechanism are defined, update the elaboration code to do full update mechanism elaboration, and then add the update mechanism to the chain tests. Fix up any errors that result from this.

PR #574, opened 2019-07-15
Elaborate update payload

dnadales:
- Add elaborators for the update payload
- Fix the errors caused by updates in the executable specs

Part of #536

### DIFF
--- a/cardano-ledger/src/Cardano/Chain/Update/ProtocolParametersUpdate.hs
+++ b/cardano-ledger/src/Cardano/Chain/Update/ProtocolParametersUpdate.hs
@@ -38,7 +38,7 @@ data ProtocolParametersUpdate = ProtocolParametersUpdate
   , ppuHeavyDelThd       :: !(Maybe LovelacePortion)
   , ppuUpdateVoteThd     :: !(Maybe LovelacePortion)
   , ppuUpdateProposalThd :: !(Maybe LovelacePortion)
-  , ppuUpdateImplicit    :: !(Maybe SlotNumber)
+  , ppuUpdateProposalTTL :: !(Maybe SlotNumber)
   , ppuSoftforkRule      :: !(Maybe SoftforkRule)
   , ppuTxFeePolicy       :: !(Maybe TxFeePolicy)
   , ppuUnlockStakeEpoch  :: !(Maybe EpochNumber)
@@ -73,7 +73,7 @@ instance B.Buildable ProtocolParametersUpdate where
     (ppuHeavyDelThd ppu)
     (ppuUpdateVoteThd ppu)
     (ppuUpdateProposalThd ppu)
-    (ppuUpdateImplicit ppu)
+    (ppuUpdateProposalTTL ppu)
     (ppuSoftforkRule ppu)
     (ppuTxFeePolicy ppu)
     (ppuUnlockStakeEpoch ppu)
@@ -97,7 +97,7 @@ instance ToCBOR ProtocolParametersUpdate where
       <> toCBOR (ppuHeavyDelThd ppu)
       <> toCBOR (ppuUpdateVoteThd ppu)
       <> toCBOR (ppuUpdateProposalThd ppu)
-      <> toCBOR (ppuUpdateImplicit ppu)
+      <> toCBOR (ppuUpdateProposalTTL ppu)
       <> toCBOR (ppuSoftforkRule ppu)
       <> toCBOR (ppuTxFeePolicy ppu)
       <> toCBOR (ppuUnlockStakeEpoch ppu)
@@ -133,7 +133,7 @@ empty = ProtocolParametersUpdate
   , ppuHeavyDelThd      = Nothing
   , ppuUpdateVoteThd    = Nothing
   , ppuUpdateProposalThd = Nothing
-  , ppuUpdateImplicit   = Nothing
+  , ppuUpdateProposalTTL   = Nothing
   , ppuSoftforkRule     = Nothing
   , ppuTxFeePolicy      = Nothing
   , ppuUnlockStakeEpoch = Nothing
@@ -159,7 +159,7 @@ apply ppu pp = ProtocolParameters
   , ppUpdateProposalThd = fromMaybe
     (ppUpdateProposalThd pp)
     (ppuUpdateProposalThd ppu)
-  , ppUpdateProposalTTL = fromMaybe (ppUpdateProposalTTL pp) (ppuUpdateImplicit ppu)
+  , ppUpdateProposalTTL = fromMaybe (ppUpdateProposalTTL pp) (ppuUpdateProposalTTL ppu)
   , ppSoftforkRule      = fromMaybe (ppSoftforkRule pp) (ppuSoftforkRule ppu)
   , ppTxFeePolicy       = fromMaybe (ppTxFeePolicy pp) (ppuTxFeePolicy ppu)
   , ppUnlockStakeEpoch  = fromMaybe
--- a/cardano-ledger/test/Test/Cardano/Chain/Block/Model.hs
+++ b/cardano-ledger/test/Test/Cardano/Chain/Block/Model.hs
@@ -27,7 +27,17 @@ import Cardano.Prelude hiding (trace, State)
 import Test.Cardano.Prelude
 
 import Control.Lens ((^.))
-import Hedgehog (MonadTest, PropertyT, collect, evalEither, forAll, property)
+import qualified Data.Set as Set
+import Data.Word (Word64)
+import Hedgehog
+  ( MonadTest
+  , PropertyT
+  , collect
+  , evalEither
+  , footnote
+  , forAll
+  , property
+  )
 
 import Cardano.Chain.Block
   ( BlockValidationMode (..)
@@ -36,8 +46,8 @@ import Cardano.Chain.Block
   , initialChainValidationState
   , updateBlock
   )
+import Cardano.Chain.Common (unBlockCount)
 import qualified Cardano.Chain.Genesis as Genesis
-import qualified Cardano.Chain.UTxO as Concrete
 import Cardano.Chain.ValidationMode
   (ValidationMode, fromBlockValidationMode)
 import Cardano.Spec.Chain.STS.Rule.Chain (CHAIN)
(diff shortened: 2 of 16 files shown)


cardano-ledger/test/Test/Cardano/Chain/Elaboration/Block.hs
```diff
@@ -112,28 +145,56 @@ elaborate txIdMap config dCert st ab =
...
+    = fmap (identity &&& elaborateUpdateProposal pm)
+    $ Abstract._bUpdProp
+    $ Abstract._bBody abstractBlock
+
+  proposalsIdMap' :: Map Abstract.Update.UpId Update.UpId
+  proposalsIdMap' = maybe proposalsIdMap addUpdateProposalId maybeProposals
+    where
+      addUpdateProposalId (abstractProposal, concreteProposal) =
+        -- TODO: check whether this is the right way of getting a proposal id.
+        -- UTxO uses 'hashDecoded' instead
```
dnadales: @ruhatch ? :arrow_up:

cardano-ledger/src/Cardano/Chain/Update/ProtocolParametersUpdate.hs
```diff
@@ -38,7 +38,7 @@ data ProtocolParametersUpdate = ProtocolParametersUpdate
   , ppuHeavyDelThd       :: !(Maybe LovelacePortion)
   , ppuUpdateVoteThd     :: !(Maybe LovelacePortion)
   , ppuUpdateProposalThd :: !(Maybe LovelacePortion)
-  , ppuUpdateImplicit    :: !(Maybe SlotNumber)
+  , ppuUpdateImplicit    :: !(Maybe SlotNumber) -- TODO: this should be named 'ppUpdateImplicit'
```
ruhatch: I think TODO is wrong, the name looks fine to me (while there's still prefixes here anyway)

cardano-ledger/test/Test/Cardano/Chain/Block/Model.hs
```diff
@@ -67,43 +85,60 @@ tests = $$discoverPropArg
 ts_prop_generatedChainsAreValidated :: TSProperty
 ts_prop_generatedChainsAreValidated =
   withTestsTS 100 $ property $ do
-    tr <- forAll $ trace @CHAIN 100
-    classifyTraceLength tr 100 50
+    let (traceLength, step) = (100 :: Word64, 10 :: Word64)
```
ruhatch, line 88: I find this slightly confusing - why not use two lines for two bindings?

cardano-ledger/test/Test/Cardano/Chain/Delegation/Model.hs
```diff
@@ -12,13 +12,17 @@ where
 
 import Cardano.Prelude
```
ruhatch, line 13: Maybe we should be chucking this file now?

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Block.hs
```diff
@@ -57,30 +65,54 @@ import Cardano.Chain.Common
...
 import Test.Cardano.Chain.UTxO.Model (elaborateTxWitnesses)
 import qualified Test.Cardano.Crypto.Dummy as Dummy
 
 
+data AbstractToConcreteIdMaps = AbstractToConcreteIdMaps
+  { transactionIds :: !(Map Abstract.TxId UTxO.TxId)
+  , proposalIds :: !(Map Abstract.Update.UpId Update.UpId)
+  } deriving (Eq, Show)
+
+instance Semigroup AbstractToConcreteIdMaps where
```
ruhatch, line 85: Is this something you could derive?

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Block.hs
```diff
@@ -57,30 +65,54 @@ import Cardano.Chain.Common
...
+  (AbstractToConcreteIdMaps tids0 pids0) <> (AbstractToConcreteIdMaps tids1 pids1) =
+    AbstractToConcreteIdMaps (tids0 <> tids1) (pids0 <> pids1)
+
+instance Monoid AbstractToConcreteIdMaps where
```
ruhatch, line 89: Same with this, looks pretty derivable

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Block.hs
```diff
@@ -112,28 +145,56 @@ elaborate txIdMap config dCert st ab =
...
-    ab
+    abstractBlock
       ^.. (Abstract.bBody . Abstract.bDCerts . traverse . to
             (elaborateDCert pm)
           )
 
   (txPayload, txIdMap') = first (fmap void) $ elaborateTxWitnesses
     txIdMap
-    (reverse $ ab ^. Abstract.bBody . Abstract.bUtxo)
+    (reverse $ abstractBlock ^. Abstract.bBody . Abstract.bUtxo)
```
ruhatch, line 159: Didn't someone fix this so we don't have to reverse anymore?

ruhatch: Maybe we need to update the ledger specs version

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Block.hs
```diff
@@ -112,28 +145,56 @@ elaborate txIdMap config dCert st ab =
...
+
+  updatePayload :: Update.APayload ()
+  updatePayload =
+    Update.APayload
+      ( fmap snd maybeProposals )
```
ruhatch: This indentation and spacing is kind of confusing

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Block.hs
ruhatch: You could use `hashDecoded`, but you'd have to change the `elaborateUpdateProposal` return type to an `AProposal`. Either would be fine really.

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Update.hs
```diff
@@ -47,3 +61,110 @@ elaboratePParams pps = Concrete.ProtocolParameters
...
+    (fromIntegral major)
+    (fromIntegral minor)
+    (fromIntegral alternative)
+
+elaborateSoftwareVersion
+  :: Abstract.SwVer
+  -> Concrete.SoftwareVersion
+elaborateSoftwareVersion _ = Concrete.SoftwareVersion applicationName applicationVersion
+  where
+    -- TODO: there is no software version in the abstract block!
```
ruhatch: You could still implement this elaborator and then pass in `Abstract.SwVer` that's hardcoded

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Update.hs
```diff
@@ -47,3 +61,110 @@ elaboratePParams pps = Concrete.ProtocolParameters
...
+  , Proposal.softwareVersion =
+      elaborateSoftwareVersion $ Abstract._upSwVer proposal
+  , Proposal.metadata =
+      Map.fromList $ zip systemTags systemHashes
+  }
+  where
+    systemTags = fmap elaborateSystemTag $ Set.toList $ Abstract._upSTags proposal
+    -- TODO: we might need different hashes here, which means that either the
+    -- elaborators should be able to generate random data, or the abstract
+    -- update payload should include (an abstract version of) these hashes.
```
ruhatch, line 119: Yeah, you could just include `ByteString`s in the abstract block and then we can hash those here

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Update.hs
```diff
@@ -47,3 +61,110 @@ elaboratePParams pps = Concrete.ProtocolParameters
...
+    systemHashes = repeat $ Concrete.InstallerHash $ coerce $ H.hash ("" :: ByteString)
+
+-- | Convert a 'ProtocolParameters' value to a 'ProtocolParametersUpdate'
+--
+-- TODO: in the spec we do not use maybe fields, so we might want to remove
```
ruhatch: In the formal specs we use a map `Ppm |-> Value`, which is isomorphic to this record with `Maybe` values. It's only in the executable spec that we use a whole `ProtocolParameters`-esque record for the update. I think we should think more about how we're doing it there, rather than removing anything here.

ruhatch (approved):
LGTM, just a few small things

cardano-ledger/test/Test/Cardano/Chain/Block/Model.hs
dnadales: I guess to me it made sense to have them in a tuple since trace length and step are related (if you change the trace length you might want to change the step as well, since otherwise you might end up with too many or too few intervals). Should I make two bindings?

cardano-ledger/test/Test/Cardano/Chain/Delegation/Model.hs
dnadales: Hmmm, but don't we want a test that checks that generated valid delegation chains are valid according to our concrete validation?

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Block.hs
dnadales: No, this isn't fixed yet.

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Block.hs
dnadales: Hmmm, how could you derive this? This isn't a `newtype`. Are there any strategies I could use? I guess `DertiveAnyClass` is not what I want...

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Block.hs
dnadales: This is a task in the next iterations. Maybe I should convert it to an issue and put it in the backlog ...

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Block.hs
dnadales: Ups, there was a space too many. What about:

```haskell
    Update.APayload
      (fmap snd maybeProposals)
      (fmap (elaborateVote pm proposalsIdMap')
      $ Abstract._bUpdVotes
      $ Abstract._bBody abstractBlock
      )
      ()
```

Otherwise I don't know how can I make this less confusing. If we avoid multiline parameters then I don't know how can we avoid long lines :scream:

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Update.hs
dnadales: Yup. I just did that. Thanks.

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Update.hs
dnadales: Right. So we should generate the hashes in the abstract blocks right?

cardano-ledger/test/Test/Cardano/Chain/Delegation/Model.hs
ruhatch: Oh, right, I thought we reactivated the Trace version of this?

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Block.hs
ruhatch: Yeah, `DeriveGeneric` and `DeriveAnyClass` are exactly what you want ;)

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Block.hs
ruhatch: Yeah, that's better :)

PR #577, opened 2019-07-17
BoundaryValidationData annotation for whole block

avieth:
byron-proxy needs to de-serialize and re-serialize boundary blocks.
By retaining the bytes of the whole boundary block, this is now
possible. The CBOR list length and tag discriminator bytes can be
pre-pended to this byte string annotation.


ruhatch (approved):
LGTM :D

PR #580, opened 2019-07-18
export KeyHash constructor

avieth:
I have an application in which I need to construct a KeyHash from an
already hashed thing, namely a StakeholderId from cardano-sl.

This doesn't make it any less safe. `KeyHash` has `FromCBOR` so you
could always use that to construct one that's not actually a hash of a
`VerificationKey`.


ruhatch (approved):
LGTM!

bors r+

PR #586, opened 2019-07-25
Remove epochTransition from updateHeader

ruhatch:
This updates the epoch and header processing to match recent changes in the Byron chain specification.


dnadales (approved):
LGTM

PR #591, opened 2019-07-30
Export validateHeaderMatchesBody in Cardano.Chain.Block.Validation

intricate:
This function is required for https://github.com/input-output-hk/ouroboros-network/issues/595. It seems I forgot to export it in https://github.com/input-output-hk/cardano-ledger/pull/549.


mrBliss (approved):
LGTM

We could work around this by passing an `ABlock` record containing the header and body to the exported `validateBlockProofs` function, but I think that's less clean.

PR #596, opened 2019-07-31
Add update payload to valid chain conformance tests

### LINKED ISSUE #536: Add Update mechanism to conformance tests

dnadales:
- Bump to the latest version of `cardano-ledger-specs`.
- Elaborate the constants used to calculate the minimum fee.
- Fix delegation-certificate scheduling: we were adding certificates at the front of the list, instead of at the back.
- Update `updateDelegation` to conform to the `DELEG` rule.
- Preserve injectivity constraint when activating a delegation by ignoring the delegation.
- Change application versions map on empty votes as well.
- Change the constant needed to elaborate maximum blocks sizes. The constant needs to be greater or equal than the one used to elaborate maximum transaction sizes, otherwise the conformance tests won't pass, since valid models will be invalid due to proposals where the maximum (concrete) transaction size is larger than the maximum (concrete) block size.
- Add examples scaffolding to aid debugging.
- Adapt to vanishing of cfmThd in CLS
- Adopt the software version only considering stable and confirmed proposal. We adapted the spec, but didn't adapt the implementation.

Closes #536


cardano-ledger/cardano-ledger.cabal
```diff
@@ -280,7 +281,7 @@ test-suite cardano-ledger-test
                        -fno-warn-missing-import-lists
                        -fno-warn-safe
                        -fno-warn-unsafe
-                       "-with-rtsopts=-K450K -M500M"
+                       "-with-rtsopts=-K5M -M500M"
```
intricate, line 286: Should we be concerned that we now require a 5M stack to run the test-suite or is this okay?

cardano-ledger/src/Cardano/Chain/Delegation/Validation/Interface.hs
```diff
@@ -169,10 +166,6 @@ updateDelegation env is certificates = do
   Environment { protocolMagic, allowedDelegators, k, currentEpoch, currentSlot }
     = env
 
-  inWindow s = subSlotCount d currentSlot <= s && s <= addSlotCount d currentSlot
```
intricate, line 172: Just curious: was this check incorrect? I haven't referenced the rules or anything but just making sure that the `(currentSlot + 1 <=)` condition above is correct.

cardano-ledger/src/Cardano/Chain/Update/Validation/Endorsement.hs
```diff
@@ -148,7 +148,7 @@ register env st endorsement =
    where
     -- Stable and confirmed proposals.
     scps     = M.filter (<= stableAt) confirmedProposals
-    stableAt = currentSlot - twice k
+    stableAt = currentSlot `subSlotNumber` twice k
```
intricate: :+1:

cardano-ledger/src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -194,15 +194,15 @@ registerUpdate env st Signal { proposal, votes, endorsement } = do
   -- Register proposal if it exists
   st' <- case proposal of
     Nothing -> pure st
-    Just p  -> registerProposal env st p
+    Just p  -> do
```
intricate: We can remove this `do`, yeah?

cardano-ledger/src/Cardano/Chain/Delegation/Map.hs
```diff
@@ -47,6 +48,11 @@ instance ToCBOR Map where
 memberR :: KeyHash -> Map -> Bool
 memberR b = Bimap.memberR b . unMap
 
+-- TODO: maybe we should call these @delegate@ and @notADelegate@ (and add also a @delegator@) function.
```
ruhatch, line 51: Good point, we should definitely make these more specific :)

intricate (commented):
Not a complete review but a couple questions.

cardano-ledger/src/Cardano/Chain/Update/Validation/Endorsement.hs
ruhatch: Hmm, I think subtracting `SlotNumber`s is a bit weird. I think `subSlotCount` and `kSlotSecurityParam` would be more natural. So calculating `SlotCount` need to be substracted.

cardano-ledger/src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -252,46 +252,72 @@ registerProposal env st proposal = do
...
                     | (!pid, !sv) <- M.toList registeredSoftwareUpdateProposals
-                    , pid `elem` M.keys confirmedProposals'
+                    , pid `elem` M.keys stableProposals
                     ]
+  pure $!
+    st' { -- Note that it's important that the new application versions are passed
+          -- as the first argument of @M.union@, since the values in this first
+          -- argument overwrite the values in the second.
+          appVersions = M.union appVersions' appVersions
+          -- TODO: consider using the `Relation` instances from `cardano-ledger-specs` (see `Ledger.Core`)
```
ruhatch, line 299: Yeah, I was about to take this on before I moved over to Shelley work. I think it would be good and shouldn't impact performance too much.

cardano-ledger/src/Cardano/Chain/Update/Validation/Endorsement.hs
ruhatch: Although maybe your thing is clearer :man_shrugging:! Up to you :)

cardano-ledger/src/Cardano/Chain/Delegation/Validation/Interface.hs
dnadales: We don't keep a rolling window anymore, so we apply the scheduled certificates up to `currentSlot` and then keep the ones starting from `currentSlot + 1`.

cardano-ledger/src/Cardano/Chain/Update/Validation/Interface.hs
dnadales: Ups. It seems a vestige from the `Debug.trace` I inserted below this `do`. I'll delete it.

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Update.hs
```diff
@@ -35,32 +39,51 @@ elaboratePParams :: Abstract.PParams -> Concrete.ProtocolParameters
 elaboratePParams pps = Concrete.ProtocolParameters
   { Concrete.ppScriptVersion      = fromIntegral $ Abstract._scriptVersion pps
   , Concrete.ppSlotDuration       = Concrete.ppSlotDuration dummyProtocolParameters
-  , Concrete.ppMaxBlockSize       = 748 * Abstract._maxBkSz pps
+  , Concrete.ppMaxBlockSize       = 4096 * Abstract._maxBkSz pps --748 * Abstract._maxBkSz pps
   , Concrete.ppMaxHeaderSize      = 95 * Abstract._maxHdrSz pps
```
ruhatch, line 43: Did you skip this one on purpose?

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Update.hs
```diff
@@ -35,32 +39,51 @@ elaboratePParams :: Abstract.PParams -> Concrete.ProtocolParameters
...
-  intToLovelace x =
-    case Concrete.mkLovelace (fromIntegral x) of
-    Left err -> panic $ "intToLovelace: " <> show err
-    Right l -> l
+
+elaborateFeePolicy
+  :: Int
+  -> Int
+  -> Concrete.TxFeePolicy
+-- TODO: we should pass the factors wrapped in some type, to avoid errors
```
ruhatch, line 78: Good idea :)

ruhatch: `FeeConstant` and `FeeCoefficient`?

stack.yaml
```diff
@@ -54,6 +54,8 @@ extra-deps:
   - git: https://github.com/input-output-hk/cardano-sl-x509
     commit: ec96c64c665b741c17b4e38f611315bae9b0b054
 
+  - Unique-0.4.7.6@sha256:a1ff411f4d68c756e01e8d532fbe8e57f1ac77f2cc0ee8a999770be2bca185c5,2723
```
ruhatch: I don't usually include the `sha256` in here as I don't think we're really concerned and it pollutes the file

ruhatch (approved):
Yeah, looks really good :)

cardano-ledger/src/Cardano/Chain/Update/Validation/Endorsement.hs
dnadales: Yeah! Subtracting slot numbers doesn't sound right. And since we already have `kSlotSecurityParam` I'll change this like you suggested. I like it better :) Thanks!

dnadales: I'm deleting `twice` and `subSlotNumber` as well.

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Update.hs
dnadales: Ups, I should delete that comment. I had to bump that number since otherwise me maximum transaction size would be larger than the maximum block size.

cardano-ledger/test/Test/Cardano/Chain/Elaboration/Update.hs
dnadales: Done. Although I should probably add those wrappers in the specs as well, but I don't want to rabbit hole.

stack.yaml
dnadales: Oh, I thought the new version of stack required that. I'll remove it.

PR #608, opened 2019-08-08
Initial scaffolding for invalid chains conformance tests.

### LINKED ISSUE #711: Address TODOs in the new `.API` module
The `.API` module, copied from the old `.Aux` module in consensus, contains a bunch of TODOs, mostly related to questions about spec (related: #706) and responsibilities (related: #708). These should be addressed.

dnadales:
Draft PR for now to get early feedback.

Closes #711.

### DIFF
--- a/cardano-ledger/test/Test/Cardano/Chain/Block/Model.hs
+++ b/cardano-ledger/test/Test/Cardano/Chain/Block/Model.hs
@@ -163,14 +163,14 @@ ts_prop_invalidDelegationSignalsAreRejected =
     let traceLength = 100 :: Word64
     tr <- forAll $ invalidTrace @CHAIN traceLength failureProfile
     let ValidationOutput { elaboratedConfig, result }
-          = applyTrace (Invalid.Trace.prefix tr)
+          = applyTrace (Invalid.Trace.validPrefix tr)
     case result of
       Left error -> do
         footnote $ "Expecting a valid prefix but got: " ++ show error
         failure
       Right concreteState ->
-        let abstractState = lastState (Invalid.Trace.prefix tr)
-            block = Invalid.Trace.sig tr
+        let abstractState = lastState (Invalid.Trace.validPrefix tr)
+            block = Invalid.Trace.signal tr
             result' = elaborateAndUpdate
                     elaboratedConfig
                     concreteState
@@ -211,9 +211,10 @@ ts_prop_invalidDelegationSignalsAreRejected =
                         <*> invalidDelegationCerts
           where
             addDelegation block delegationCerts =
-              let newBody = (Abstract._bBody block)
-                              { Abstract._bDCerts = delegationCerts }
-              in block { Abstract._bBody = newBody }
+              Abstract.updateBody
+                block
+                (\body -> body { Abstract._bDCerts = delegationCerts })
+
             invalidDelegationCerts = Gen.list (Range.constant 0 10)
                                               (randomDCertGen delegationEnv)
               where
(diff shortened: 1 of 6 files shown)


cardano-ledger/test/Test/Cardano/Chain/Block/Model.hs
```diff
@@ -150,3 +139,88 @@ classifyTransactions =
...
+            pure ()
+          (abstractResult, concreteResult) -> do
+            footnote "Validation results mismatch."
+            footnote $ "Abstract result: " ++ show abstractResult
+            footnote $ "Concrete result: " ++ show concreteResult
+            failure
+  where
+    -- TODO: define this.
+    failureProfile :: [(Int, PredicateFailure CHAIN)]
+    failureProfile = [(1, panic mempty)] -- TODO: we need to modify the failure
```
dnadales: @nc6 @mhuesch I think we might not need three separate tests, but instead pass a failure profile here that looks like this:

```haskell
failureProfile :: [(Int, Constr)]
failureProfile =
  [ (1, toConstr DelegationFailure)
  , (1, toConstr BUPIFailure)
  , (1, toConstr UTXOWSFailure)
  ] 
```

But something tells me this might be overly simplistic and not match that @mhuesch has on the goblins side. So let's discuss whether this makes sense, and what do we need to implement this.

mhuesch: In order to match the `PredicateFailure CHAIN` datatypes, wouldn't the 3 constructors in `failureProfile` actually have to be wrapped with `BBodyFailure`, so that they were of type `PredicateFailure CHAIN`?

I have been using a more "specific" equality predicate than the `toConstr x == toConstr y` one.

mhuesch: @dnadales But I agree with what you're saying, and it's very close to what I have been doing with goblins, modulo the `PredicateFailure` equality predicate.

If we wrote out all the constructors of `PredicateFailure CHAIN` in `failureProfile`, then we should get good coverage.

mhuesch (commented):
I will do a more thorough review of this Friday.

cardano-ledger/test/Test/Cardano/Chain/Block/Model.hs
dnadales: :+1: It seems we need to agree on how we want to specify the predicate failures.

cardano-ledger/test/Test/Cardano/Chain/Block/Model.hs
```diff
@@ -150,3 +156,115 @@ classifyTransactions =
...
     . fmap (length . Abstract._bUtxo . Abstract._bBody)
     . traceSignals NewestFirst
+
+ts_prop_invalidDelegationSignalsAreRejected :: TSProperty
+ts_prop_invalidDelegationSignalsAreRejected =
+  withTestsTS 300  $ property $ do
+    let traceLength = 100 :: Word64
+    tr <- forAll $ invalidTrace @CHAIN traceLength failureProfile
+    let ValidationOutput { elaboratedConfig, result }
+          = applyTrace (Invalid.Trace.prefix tr)
```
ruhatch: Maybe `validPrefix` would be clearer than `prefix`?

ruhatch (approved):
Nice, this is really cool :)

ruhatch:
Feel free to ask for another review once you've got it ready to merge :)

cardano-ledger/test/Test/Cardano/Chain/Block/Model.hs
dnadales: Nice one! I'm changing this in `cardano-ledger-specs` (luckily I saw this before merging the other PR :D)

ruhatch (approved):
LGTM!

mhuesch (approved):
LGTM :shipit:

PR #614, opened 2019-08-12
[#543] Do not allow delegations whose epoch is past the next one.

### LINKED ISSUE #543: Do not allow delegations whose epoch is past the next one
## Background

The Byron spec does not allow delegations where the epoch in the certificate is past the next epoch from the epoch the certificate was issued.

image

## Proposed Solution

Adapt the `scheduleCertificate` function so that this check is enforced. I.e. change this condition accordingly:

```haskell
  -- Check that the delegation epoch is greater than or equal to the current one
  currentEpoch <= delegationEpoch
    `orThrowError` PastEpoch currentEpoch delegationEpoch
```

csoroz:
Checks the limits of the delegation epoch as defined in the Byron spec: `0 <= e_d - e <= 1`

Fixes [#543](https://github.com/input-output-hk/cardano-ledger/issues/543)

### DIFF
--- a/cardano-ledger/src/Cardano/Chain/Delegation/Validation/Scheduling.hs
+++ b/cardano-ledger/src/Cardano/Chain/Delegation/Validation/Scheduling.hs
@@ -105,8 +105,8 @@ data Error
   | NonGenesisDelegator KeyHash
   -- ^ This delegator is not one of the allowed genesis keys
 
-  | PastEpoch EpochNumber EpochNumber
-  -- ^ This delegation is for a past epoch
+  | WrongEpoch EpochNumber EpochNumber
+  -- ^ This delegation is for a past or for a too future epoch
 
   deriving (Eq, Show)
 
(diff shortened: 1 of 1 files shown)


ruhatch (approved):
Thanks for this!

Not sure about the phrase 'too future', haha, but I can't think of anything much better

csoroz:
> Not sure about the phrase 'too future', haha, but I can't think of anything much better

Please, correct it if you find a better alternative! ;-P

dnadales:
Now we will never know whether this discrepancy would have been caught by the invalid conformance testing :stuck_out_tongue:  Thanks!

dnadales:
> Now we will never know whether this discrepancy would have been caught by the invalid conformance testing  Thanks!

Well, that's actually not true :) I just hooked up the conformance tests for invalid-chains and this is what I got:

image

So I'm going to rebase to `master` and see which other discrepancies we find.

PR #623, opened 2019-08-22
Modify the MempoolPayload data type

intricate:
When originally implementing #514 (PR #554), I wasn't _exactly_ aware of how this `MempoolPayload` data type was going to be used and I defined it incorrectly. The mempool needs to be able to hold _individual_ ledger objects (transactions, delegation certs, update proposals, and update votes) while the `*Payload` data types I originally utilized hold collections of the aforementioned objects.


cardano-ledger/src/Cardano/Chain/MempoolPayload.hs
```diff
@@ -30,28 +32,35 @@ type MempoolPayload = AMempoolPayload ()
...
-  | MempoolUpdatePayload !(Update.APayload a)
-  -- ^ An update payload.
-  deriving (Eq, Show)
+  = MempoolTxPayload !(ATxAux a)
+  -- ^ A transaction payload (transaction and witness).
+  | MempoolDlgPayload !(Delegation.ACertificate a)
+  -- ^ A delegation certificate payload.
+  | MempoolUpdateProposalPayload !(Update.AProposal a)
+  -- ^ An update proposal payload.
+  | MempoolUpdateVotePayload !(Update.AVote a)
```
intricate: @ruhatch @dnadales: Wouldn't mind an opinion on the naming here. Do you think `MempoolUpdate*` is too verbose? Could use `MempoolUp*` or `MempoolUpd*`.

dnadales: I think verbosity is fine here. @ruhatch once referred me to [this article](https://chrisdone.com/posts/german-naming-convention/) about long names and why they're ok. There are exceptions, but I don't think this is one of them ;)

intricate:
It seems that this should be sufficient for https://github.com/input-output-hk/ouroboros-network/issues/592. Ready for review.

cardano-ledger/src/Cardano/Chain/MempoolPayload.hs
intricate: In that case, maybe I could expand `MempoolDlgPayload` to `MempoolDelegationPayload` (or perhaps that's the exception :smile:).

deepfire: I think that using _well_ established short names is useful -- i.e. `Tx` vs. `Transaction` and `Dlg` vs. `Delegation` bring value.

I think that is because the brain has some resource limitation for visual processing, so if you clog it too much with low-SNR, the perception throughput drops.

Now, this doesn't answer the question of how do you arrive at the situation where a certain name shortening is _well_ established, but that question is separate, I think.

`Dlg` and `Tx` seem to be well-established in the ledger code -- speaking as a relative outsider, so take my view with a grain of salt.

deepfire (approved):
LGTM!

cardano-ledger/src/Cardano/Chain/MempoolPayload.hs
intricate: @deepfire fair enough. I think, as you said, `Dlg` is established enough in this repo to continue using it.

ruhatch: @intricate you could just do:
```
data AMempoolPayload a
  = MempoolTx !(ATxAux a)
  -- ^ A transaction payload (transaction and witness).
  | MempoolDlg !(Delegation.ACertificate a)
  -- ^ A delegation certificate payload.
  | MempoolUpdateProposal !(Update.AProposal a)
  -- ^ An update proposal payload.
  | MempoolUpdateVote !(Update.AVote a)
```

ruhatch (approved):
LGTM :D

cardano-ledger/src/Cardano/Chain/MempoolPayload.hs
intricate: @ruhatch: Done. Thanks.

PR #628, opened 2019-09-03
Add CBOR serialization instances for UTxO, update, and delegation error types

intricate:
Required as part of https://github.com/input-output-hk/ouroboros-network/issues/592 since the `LocalTxSubmission` `Codec` requires the ability to CBOR encode and decode `ApplyTxErr`s (which can be either transaction, update, or delegation errors).

I'll be adding golden tests in follow-up PRs (I believe there are around 40 golden tests required as a result of this PR). Also, I still need to add generators or round-trip tests for `Update`-related error types.

### DIFF
--- a/cardano-ledger/src/Cardano/Chain/Delegation/Validation/Scheduling.hs
+++ b/cardano-ledger/src/Cardano/Chain/Delegation/Validation/Scheduling.hs
@@ -23,10 +23,15 @@ import qualified Data.Set as Set
 
 import Cardano.Binary
   ( Annotated(..)
+  , Decoder
+  , DecoderError(..)
   , FromCBOR(..)
   , ToCBOR(..)
+  , decodeListLen
+  , decodeWord8
   , encodeListLen
   , enforceSize
+  , matchSize
   )
 import Cardano.Chain.Common (BlockCount, KeyHash, hashKey)
 import Cardano.Chain.Delegation.Certificate (ACertificate)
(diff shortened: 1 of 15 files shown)


intricate:
@ruhatch: 

The modules containing the update-related `Error` types for which I have to write generators and tests are not exposed as part of `cardano-ledger` (and therefore cannot be imported in the test-suite). It seems that there is good reason for keeping these modules internal as the update functionality is supposed to be exposed from `Cardano.Chain.Update.Validation.Interface`. 

Would you mind if I re-export the required `Error` types through the `Interface` module? Also, I wonder if it'd make sense to rename them since they're all named `Error`.

Perhaps a rename to: `EndorsementError`, `RegistrationError`, and `VotingError`?

mhuesch:
@intricate I feel a little bad suggesting this _after_ you did all the work of manually implementing CBOR instances (wish I had asked you about the work earlier when you said you were starting it!), but could we use TH to derive these boilerplate-y instances?

I wrote https://github.com/input-output-hk/cardano-sl/blob/51ad7c0503b1c52a75a6eb36096c407934136468/binary/src/Pos/Binary/Class/TH.hs#L142-L143 in `cardano-sl` and that could probably be modified relatively easily to work with `ToCBOR`/`FromCBOR`. I had to exercise my TH-fu for goblins, so I am relatively fresh on this and could implement it if desired.

We probably want to check with someone with higher context, if there's a reason not to use TH to derive these.

mhuesch:
Although, I just read the "Avoid TemplateHaskell" section of http://www.stephendiehl.com/posts/production.html 😆

I still think that in this case we're only deriving instances which look fairly manual to hand-write, so I think it is a legitimate use of TH. Now that the work is done, though, perhaps it's too late to justify implementing a TH derivation function.

ruhatch (approved):
LGTM! Have fun with the golden tests ;)

intricate:
> @intricate I feel a little bad suggesting this _after_ you did all the work of manually implementing CBOR instances (wish I had asked you about the work earlier when you said you were starting it!), but could we use TH to derive these boilerplate-y instances?
> 
> I wrote https://github.com/input-output-hk/cardano-sl/blob/51ad7c0503b1c52a75a6eb36096c407934136468/binary/src/Pos/Binary/Class/TH.hs#L142-L143 in `cardano-sl` and that could probably be modified relatively easily to work with `ToCBOR`/`FromCBOR`. I had to exercise my TH-fu for goblins, so I am relatively fresh on this and could implement it if desired.
> 
> We probably want to check with someone with higher context, if there's a reason not to use TH to derive these.

@mhuesch: This is a good suggestion but, in the interest of getting this merged so I can utilize it for https://github.com/input-output-hk/ouroboros-network/issues/592, would you mind if I merge this and open an issue about potentially utilizing TH for some of the CBOR instances throughout `cardano-ledger` (I believe there are a bunch more just like these)?

mhuesch:
@intricate of course! I was just thinking aloud. If a derivation function is written that covers these instances, it’s easy to delete them.

PR #631, opened 2019-09-05
Redxaxder/new annotations

redxaxder:
(also: https://github.com/input-output-hk/cardano-base/pull/41)

Previously, we used a division of types between `Block` and `ABlock ByteString` to ensure that we don't accidentally re-serialize previously deserialized bytes.

We can accomplish the same goal without this division by putting a lazy reference to a `ByteString` in the constructor for `Block`. For a `Block` constructed locally, this will point to a thunk with serialization logic. For a deserialized `Block`, this will point to the original bytes.

For a type that has gone through this change, we later "serialize" it by extracting the above reference.

This fails the round trip encoding tests because nested annotations aren't being constructed properly during the decoding when interacting with the previous scheme. If this change was propagated to the rest of the types that would be resolved.

### DIFF
--- a/cardano-ledger/src/Cardano/Chain/UTxO/TxWitness.hs
+++ b/cardano-ledger/src/Cardano/Chain/UTxO/TxWitness.hs
@@ -4,11 +4,10 @@
 {-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TemplateHaskell     #-}
 {-# LANGUAGE TypeApplications    #-}
 
 module Cardano.Chain.UTxO.TxWitness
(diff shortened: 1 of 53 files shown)


cardano-ledger/src/Cardano/Chain/UTxO/TxWitness.hs
```diff
@@ -44,15 +46,26 @@ import Cardano.Crypto
...
   , shortHashF
   )
 
 
 -- | A witness is a proof that a transaction is allowed to spend the funds it
 --   spends (by providing signatures, redeeming scripts, etc). A separate proof
 --   is provided for each input.
-type TxWitness = Vector TxInWitness
+data TxWitness = TxWitness
+  { txInWitnesses :: Vector TxInWitness
```
ruhatch: This should probably be strict!

PR #635, opened 2019-09-16
Keep the software update metadata in the update state

### LINKED ISSUE #622: Keep the software update metadata in the update state
## Background
At the moment the implementation differs from the specification in that the former doesn't keep the metadata of the update proposal.

Specification (part of the update state):

image

Implementation (of the update state):

image

## Proposed Solution

Update the implementation to match the formal spec.

dnadales:
Closes #622


intricate (approved):
LGTM

PR #636, opened 2019-09-19
Integrate goblins into the conformance tests

mhuesch:
We add a `ts_prop_invalidTxWitsAreRejected` conformance property to `Test.Cardano.Chain.Block.Model`. Adding this property required reworking of a number of types in `cardano-ledger-specs` to capture bounds which were not previously "knowable" from the type of the value.

***

This relates to https://github.com/input-output-hk/cardano-ledger-specs/issues/788 & https://github.com/input-output-hk/cardano-ledger-specs/issues/817.

### DIFF
--- a/cardano-ledger/src/Cardano/Chain/UTxO/UTxO.hs
+++ b/cardano-ledger/src/Cardano/Chain/UTxO/UTxO.hs
@@ -102,6 +102,9 @@ fromBalances =
     . concat
     . fmap (fromTxOut . uncurry TxOut)
 
+-- | Construct a UTxO from a TxOut. This UTxO is a singleton with a TxIn that
+-- references an address constructed by hashing the TxOut address. This means
+-- it is not guaranteed (or likely) to be a real address.
 fromTxOut :: TxOut -> UTxO
 fromTxOut out = fromList [(TxInUtxo (coerce . hash $ txOutAddress out) 0, out)]
 
--- a/cardano-ledger/test/Test/Cardano/Chain/UTxO/Model.hs
+++ b/cardano-ledger/test/Test/Cardano/Chain/UTxO/Model.hs
@@ -17,6 +17,7 @@ import Cardano.Prelude hiding (trace)
 import Test.Cardano.Prelude
 
 import Control.Lens ((^.))
+import Data.Coerce (coerce)
 import qualified Data.Map.Strict as M
 
 import Hedgehog (MonadTest, evalEither, forAll, property)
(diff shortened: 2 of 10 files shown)


cardano-ledger/src/Cardano/Chain/UTxO/UTxO.hs
```diff
@@ -102,6 +102,9 @@ fromBalances =
     . concat
     . fmap (fromTxOut . uncurry TxOut)
 
+-- | Construct a UTxO from a TxOut. This UTxO is a singleton with a TxIn that
```
dnadales, line 105: :100: 

image

cardano-ledger/test/Test/Cardano/Chain/UTxO/Model.hs
```diff
@@ -131,12 +132,7 @@ elaborateTxWitsBSWithMap
 elaborateTxWitsBSWithMap txIdMap abstractTxWits = (concreteTxWitness, txIdMap')
  where
   concreteTxWitness = E.elaborateTxWitsBS
-    ( fromMaybe
```
dnadales, line 134: I'm not sure about this, before this PR we wanted the lookup to fail if we cannot find the corresponding abstract txid was not found, since that indicated a bug in the elaborator code. Now it seems that we might be covering these errors. On the other hand I get that you might need this for testing with non-existing transactions. But since I don't see a better alternative I'd just leave this complaint :)

dnadales (approved):
LGTM.

cardano-ledger/test/Test/Cardano/Chain/UTxO/Model.hs
mhuesch: Would this reflect a bug in the elaborator or in the generators? I suppose if the elaborator didn't use the right `txIdMap` and then that was empty...

I can't think of another way to do it either, since we can't distinguish between Goblin-tweaked `TxId`s which are not in the UTxO, versus ones which indicate bugs.

cardano-ledger/src/Cardano/Chain/UTxO/UTxO.hs
mhuesch: 😆🙏

cardano-ledger/test/Test/Cardano/Chain/UTxO/Model.hs
dnadales: > Would this reflect a bug in the elaborator or in the generators? I suppose if the elaborator didn't use the right txIdMap and then that was empty...

Well, you could imagine that if I forget to include the id of an elaborated transaction I might be getting this kind of error. After this PR is merged, we will detect this when the spec passes, and the concrete implementation fails to validate.

But you're right, at the moment, we can't distinguish between goblins and bugs.

mhuesch: Ah, I see - you mean including the `TxId` of an elaborated transaction into the initial UTxO? Thus all the lookups after that point will fail. I guess then the difference is that those transactions will not refer to actual `TxOut`s, but will just "invent" new `TxId`s to receive value from. Which, if I'm not mistaken, is much like how [`fromTxOut`](https://github.com/input-output-hk/cardano-ledger/blob/c9fcfa439796db0ae6b8f2316e6f2b01abb63a8b/cardano-ledger/src/Cardano/Chain/UTxO/UTxO.hs#L108-L109) generates new value to seed the UTxO.

mhuesch: > But you're right, at the moment, we can't distinguish between goblins and bugs.

👍

PR #638, opened 2019-09-20
Add CompactRedeemVerificationKey

intricate:
Related to #618.

This implementation of `RedeemVerificationKey` reduces the memory usage of loading the mainnet `Genesis.Config` from around ~45MB to ~15MB:

visualization_compactredeemverificationkey

### DIFF
--- a/crypto/src/Cardano/Crypto/Signing/Redeem/Compact.hs
+++ b/crypto/src/Cardano/Crypto/Signing/Redeem/Compact.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
@@ -45,6 +46,7 @@ data CompactRedeemVerificationKey =
                                {-# UNPACK #-} !Word64
                                {-# UNPACK #-} !Word64
   deriving (Eq, Generic, Show)
+  deriving NoUnexpectedThunks via UseIsNormalForm CompactRedeemVerificationKey
   deriving anyclass NFData
 
 getCompactRedeemVerificationKey :: Get CompactRedeemVerificationKey
--- a/crypto/src/Cardano/Crypto/Signing/Redeem/Compact.hs
+++ b/crypto/src/Cardano/Crypto/Signing/Redeem/Compact.hs
@@ -1,0 +1,106 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+module Cardano.Crypto.Signing.Redeem.Compact
+  ( CompactRedeemVerificationKey(..)
+  , fromCompactRedeemVerificationKey
+  , toCompactRedeemVerificationKey
+  )
+where
+
+import Cardano.Prelude
+
+import Data.Aeson
+  ( FromJSON(..)
+  , FromJSONKey(..)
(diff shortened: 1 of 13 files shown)


crypto/src/Cardano/Crypto/Signing/Redeem/Compact.hs
```diff
@@ -0,0 +1,107 @@
...
+    <$> getWord64le
+    <*> getWord64le
+    <*> getWord64le
+    <*> getWord64le
+
+putCompactRedeemVerificationKey :: CompactRedeemVerificationKey -> Put
+putCompactRedeemVerificationKey (CompactRedeemVerificationKey a b c d) =
+  putWord64le a >> putWord64le b
+                >> putWord64le c
+                >> putWord64le d
```
mrBliss: I would align these (add some newlines) such that it is clear that there are 4 of them. While glancing over the code, this caught my eye because it looked like the put for `a` was missing.

crypto/src/Cardano/Crypto/Signing/Redeem/Compact.hs
```diff
@@ -0,0 +1,107 @@
...
+  redeemVKBuild bs
+ where
+  bs :: ByteString
+  bs = BSL.toStrict $ runPut $
+    putCompactRedeemVerificationKey compactRvk
+
+instance Ord CompactRedeemVerificationKey where
+  a `compare` b =
+    fromCompactRedeemVerificationKey a
+      `compare` fromCompactRedeemVerificationKey b
```
mrBliss: Alternatively:
```haskell
  compare = compare `on` fromCompactRedeemVerificationKey
```

crypto/src/Cardano/Crypto/Signing/Redeem/Compact.hs
```diff
@@ -0,0 +1,107 @@
...
+
+instance FromJSON CompactRedeemVerificationKey where
+    parseJSON = fmap toCompactRedeemVerificationKey . parseJSON
+
+instance Monad m => ToObjectKey m CompactRedeemVerificationKey where
+  toObjectKey = pure . toJSString . formatToString redeemVKB64UrlF . fromCompactRedeemVerificationKey
+
+instance MonadError SchemaError m => FromObjectKey m CompactRedeemVerificationKey where
+  fromObjectKey =
+    fmap (Just .toCompactRedeemVerificationKey)
```
mrBliss: Missing space after `.`

mrBliss: Are we worried that we're just building a thunk in the `Just` that holds on to the `RedeemVerificationKey`? Based on the convincing heap profile, I suppose this is forced later on.

mrBliss:
From 45 MB to 15 MB, nice result! :tada:

crypto/src/Cardano/Crypto/Signing/Redeem/Compact.hs
intricate: The keys in a `Map` (both lazy and strict `Map`s) should be forced so I don't _think_ this should be a problem. i.e. as you said, this is forced later on. But perhaps I could add a `$!` or `seq` to ensure that the value within the `Just` is forced so that we're not needlessly building thunks.

crypto/src/Cardano/Crypto/Signing/Redeem/Compact.hs
intricate: Didn't know this was a thing. Thanks :smile:

mrBliss (approved):
LGTM

crypto/test/Test/Cardano/Crypto/Signing/Redeem/Compact.hs
```diff
@@ -0,0 +1,44 @@
...
+import Test.Cardano.Crypto.Gen (genRedeemVerificationKey)
+
+--------------------------------------------------------------------------------
+-- Compact RedeemVerificationKey
+--------------------------------------------------------------------------------
+
+roundTripCompactRedeemVerificationKey :: Property
+roundTripCompactRedeemVerificationKey =
+  eachOf
+    1000
```
dnadales, line 25: Do we need 1000 tests here? (assuming that's what 1000 means here...)

dnadales (approved):
Nice work!

crypto/test/Test/Cardano/Crypto/Signing/Redeem/Compact.hs
intricate: Not specifically. Did you have another number in mind?

PR #645, opened 2019-09-25
Dnadales/787 conformance tests with top level chain predicate failures

dnadales:


cardano-ledger/test/Test/Cardano/Chain/Block/Model.hs
```diff
@@ -157,6 +186,35 @@ elaborateAndUpdate config (cvs, abstractToConcreteIdMaps) (ast, ab) =
...
+                           config
+                           dCert
+                           chainValidationState
+                           abstractBlock
+      where
+        dCert = rcDCert
+                  (abstractBlock ^. Abstract.bHeader . Abstract.bhIssuer)
+                  stableAfter
+                  abstractState
+          where
```
nc6, line 214: Bikeshedding: not bother with the nested `where` clauses?

cardano-ledger/test/Test/Cardano/Chain/Block/Model.hs
```diff
@@ -412,3 +471,119 @@ applyTrace tr =
...
+  setMaxHeaderSize <$> Gen.integral (Range.constant 0 maxSize)
+  where
+    setMaxHeaderSize newSize = state { cvsUpdateState = newUpdateState }
+      where
+        newUpdateState =
+          (cvsUpdateState state) { adoptedProtocolParameters = newParameters }
+          where
+            newParameters =
+              (adoptedProtocolParameters (cvsUpdateState state)) { ppMaxHeaderSize = newSize }
+              -- And this is why lenses where invented...
```
nc6: :-D

nc6:
```suggestion
              -- And this is why lenses were invented...
```

nc6 (approved):
LGTM

cardano-ledger/test/Test/Cardano/Chain/Block/Model.hs
dnadales: I guess I wanted to add an extra `where` :rofl: 

Thanks!

cardano-ledger/test/Test/Cardano/Chain/Block/Model.hs
dnadales: I kind of like them, but I feel like a Taliban (hello NSA) when using them :)

PR #648, opened 2019-09-29
787 test with invalid block sizes

dnadales:
Part of https://github.com/input-output-hk/cardano-ledger-specs/issues/787


mhuesch:
@dnadales should we have been requested as reviewers instead of added as assignees? 🙂

dnadales:
> @dnadales should we have been requested as reviewers instead of added as assignees? slightly_smiling_face

Ups. Thanks for the heads up!

intricate (approved):
This LGTM :100:

PR #653, opened 2019-10-04
Add CertificateId and VoteId data types

intricate:
For https://github.com/input-output-hk/ouroboros-network/issues/592 (PR https://github.com/input-output-hk/ouroboros-network/pull/999), each kind of generalized transaction has an identifier (for which we are using a hash of the generalized tx). I think it makes sense that these data types will be maintained by and exposed as part of `cardano-ledger`. Thoughts?

Also, I'm not really sure if I like the suffix, `*Id`, in this case. I was thinking that `CertificateHash` and `VoteHash` were nicer but I don't want to make things inconsistent as we already have `TxId` and `UpId`.

### DIFF
--- a/cardano-ledger/src/Cardano/Chain/Delegation/Certificate.hs
+++ b/cardano-ledger/src/Cardano/Chain/Delegation/Certificate.hs
@@ -8,6 +8,7 @@
 {-# LANGUAGE NamedFieldPuns        #-}
 {-# LANGUAGE OverloadedStrings     #-}
 {-# LANGUAGE TypeApplications      #-}
+{-# LANGUAGE TypeFamilies          #-}
 {-# LANGUAGE TypeSynonymInstances  #-}
 
 {-# OPTIONS_GHC -fno-warn-orphans #-}
@@ -26,6 +27,7 @@ module Cardano.Chain.Delegation.Certificate
   -- * Certificate Accessor
   , epoch
   , certificateId
+  , recoverCertificateId
 
   -- * Certificate Predicate
   , isValid
@@ -42,10 +44,12 @@ import Text.JSON.Canonical
   (FromJSON(..), Int54, JSValue(..), ToJSON(..), fromJSField, mkObject)
 
 import Cardano.Binary
-  ( Annotated(..)
+  ( Annotated(Annotated, unAnnotated)
   , ByteSpan
+  , Decoded(..)
   , FromCBOR(..)
   , ToCBOR(..)
+  , annotatedDecoder
   , fromCBORAnnotated
   , encodeListLen
   , enforceSize
@@ -60,6 +64,7 @@ import Cardano.Crypto
   , Signature
   , VerificationKey(unVerificationKey)
   , hash
+  , hashDecoded
   , safeSign
   , safeToVerification
   , verifySignatureDecoded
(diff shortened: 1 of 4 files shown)


cardano-ledger/src/Cardano/Chain/Delegation/Certificate.hs
```diff
@@ -130,6 +137,9 @@ unsafeCertificate e = UnsafeACertificate (Annotated e ())
 epoch :: ACertificate a -> EpochNumber
 epoch = unAnnotated . aEpoch
 
+certificateId :: ACertificate a -> CertificateId
+certificateId = hash . void
```
mrBliss, line 148: If we have a `ACertificate ByteString`, it would be more efficient to use `hashDecoded` to avoid reserialising, right? I'm basing this on `recoverUpId`. I don't know what the convention is, but maybe we could have this `certificateId` *and* `recoverCertificateId` (which uses `hashDecoded`)?

The same applies to `VoteId`.

intricate: To utilize `hashDecoded`, I need to implement a `Decoded` instance of `ACertificate ByteString` and, for its definition of `recoverBytes`, I probably need to add a field to store the `annotation` in `ACertificate a`. I was purposely avoiding this for now as Ru told me that annotations were being reworked in this codebase.

mrBliss: Ah, right. Best to make a ticket reminding us to take advantage of the reworked annotations when finished. Ideally, the ticket refers to the relevant issue/PR in `cardano-ledger` and to the locations where we should take advantage of it (looking for `void`/`hash`/`serialise` should cover most cases).

intricate: Okay so I just added two new functions: `recoverCertificateId` and `recoverVoteId`. I added an `annotation` field and also implemented a `Decoded` instance for both `ACertificate` and `AVote`.  I figure it made sense to go ahead and just do this now and deal with the fallout as part of the new annotations implementation.

cardano-ledger/src/Cardano/Chain/Delegation/Certificate.hs
```diff
@@ -8,6 +8,7 @@
 {-# LANGUAGE NamedFieldPuns        #-}
 {-# LANGUAGE OverloadedStrings     #-}
 {-# LANGUAGE TypeApplications      #-}
+{-# LANGUAGE TypeFamilies          #-}
```
mrBliss, line 11: I'm curious (and too lazy to try it out locally :slightly_smiling_face:) why is this needed?

mrBliss (approved):
LGTM

cardano-ledger/src/Cardano/Chain/Delegation/Certificate.hs
intricate: It's needed for this line in particular: 

https://github.com/input-output-hk/cardano-ledger/pull/653/files#diff-7d7801119305711a2a81d77bb358e2b9R202

mrBliss: Ah, of course

PR #637, opened 2019-09-20
Ensure that GenesisAvvmBalances Map returned by fromJSON is strict

intricate:
_Somewhat related to #618_ 

Unfortunately, because `canonical-json` doesn't utilize operations from `Data.Map.Strict` but only those from `Data.Map.Lazy` (i.e. `fromJSON` will return a `Map` that is not necessarily strict in its values), we need to be careful in order to ensure that we're still dealing with a `Map` that's strict in both its keys and values.

To remedy this, I've used `Map.map (\a -> seq a a)` to convert the `Map` to one that is now guaranteed to be strict in both its keys and values.

### Before the Fix

- Notice that the `gdAvvmDistr` field, the last heap object at level 1 of the tree, is a `ConstrClosure` which represents the `Bin` constructor from `Data.Map`. This would make sense since `GenesisAvvmBalances` is a simple newtype wrapper around a `Map RedeemVerificationKey Lovelace`. Notice how the `Bin` `ConstrClosure`'s second pointer (an actual value of the `Data.Map`) references a `ThunkClosure`.

_n.b. The second pointer of a `Bin` is actually the value field of an entry in the `Map`. So, in this case, it should be a `Lovelace` which, in WHNF, would be represented by a `W64#` `ConstrClosure` since `Lovelace` is a newtype wrapper around a `Word64`._

```
ConstrClosure {info = StgInfoTable {entry = Nothing, ptrs = 6, nptrs = 2, tipe = CONSTR, srtlen = 0, code = Nothing}, ptrArgs = [0x0000004205caa788/1,0x0000004205caa7c0/1,0x0000004205caa800/1,0x0000000004c07e90/2,0x0000004205caa828/1,0x0000004205caa8b0/1], dataArgs = [2160,764824073], pkg = "cardano-ledger-0.1.0.0-BZ3sG7KtG0uBft8EdU7TrW", modl = "Cardano.Chain.Genesis.Data", name = "GenesisData"}
|
+- ConstrClosure {info = StgInfoTable {entry = Nothing, ptrs = 3, nptrs = 1, tipe = CONSTR, srtlen = 0, code = Nothing}, ptrArgs = [0x000000420591be40/1,0x000000420591be60/1,0x000000420591be98/1], dataArgs = [7], pkg = "containers-0.6.0.1", modl = "Data.Set.Internal", name = "Bin"}
|  |
[...]

PR #654, opened 2019-10-04
Use forceElemsToWHNF in GenesisAvvmBalances FromJSON instance

intricate:
We recently added `forceElemsToWHNF` to `cardano-prelude`, so probably makes sense that we utilize it here.

Related: #637

### DIFF
--- a/cardano-ledger/src/Cardano/Chain/Genesis/AvvmBalances.hs
+++ b/cardano-ledger/src/Cardano/Chain/Genesis/AvvmBalances.hs
@@ -10,10 +10,8 @@ module Cardano.Chain.Genesis.AvvmBalances
   )
 where
 
-import Cardano.Prelude hiding (toStrict)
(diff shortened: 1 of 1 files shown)


mrBliss (approved):
LGTM

Just a thought: what if we didn't expose the `GenesisAvvmBalances` constructor and just exposed a smart constructor that calls `forceElemsToWHNF`?

intricate:
> LGTM
> 
> Just a thought: what if we didn't expose the `GenesisAvvmBalances` constructor and just exposed a smart constructor that calls `forceElemsToWHNF`?

I wonder if, in that case, we'd want to do the same for every data type that utilizes a strict `Map`. But then maybe we'd actually just want to introduce a `StrictMap` data type since the real problem here is that `Data.Map.Strict` and `Data.Map.Lazy` utilize the same data type, `Map`.

PR #658, opened 2019-10-09
Expose module Cardano.Chain.Update.Vote

intricate:
If one imports `Cardano.Chain.Update (annotation)`, the field `annotation` of `AVote` conflicts with that of `AProposal`.

Needed for https://github.com/input-output-hk/ouroboros-network/pull/999

### DIFF
--- a/cardano-ledger/cardano-ledger.cabal
+++ b/cardano-ledger/cardano-ledger.cabal
@@ -45,6 +45,7 @@ library
                        Cardano.Chain.Update
                        Cardano.Chain.Update.Proposal
                        Cardano.Chain.Update.Validation.Interface
+                       Cardano.Chain.Update.Vote
                        Cardano.Chain.ValidationMode
 
   other-modules:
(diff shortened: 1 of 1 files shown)


mrBliss (approved):

intricate:
I could also add `Cardano.Chain.Update.Vote` as an exposed module to get around this :man_shrugging:.

dnadales:
> I could also add Cardano.Chain.Update.Vote as an exposed module to get around this :man_shrugging:.

Ohhh, now I see why you can't import `Vote` qualified. 

Yeah, I'd prefer to expose `Vote` and use it qualified. But I'm fine either way.

intricate:
@dnadales:
> Yeah, I'd prefer to expose `Vote` and use it qualified. But I'm fine either way.

Decided to go this route.

dnadales (approved):

intricate:
I've got no idea what the problem is here.

Issue #576, opened 2019-07-16
Size of TxAux


mrBliss:
For the Mempool in the consensus layer, we need to know the size of a transaction (including its auxiliary data), i.e. the size of `Cardano.Chain.UTxO.ATxAux ByteString` in bytes.

Currently we have:
```haskell
import qualified Data.ByteString as Strict

txAuxSize :: ATxAux ByteString -> Word32
txAuxSize atxaux = fromIntegral txByteSize
  where
    txByteSize = 1
               + (Strict.length . annotation . aTaTx      $ atxaux)
               + (Strict.length . annotation . aTaWitness $ atxaux)
```
However, this feels a bit brittle: what if the format of `ATxAux` changes?

There is also the `encodedSizeExpr` method of `ToCBOR`, could this somehow be used here?

We would like a function in `cardano-ledger` that gives us the size of an `ATxAux ByteString` in bytes.

ruhatch:
I imagine what we should really do here is add an annotation to the `ATxAux` and the take the size of that

ruhatch:
Although equally the size of `ATxAux` isn't going to change, so the brittle one is also probably fine

intricate:
@mrBliss: As a result of #660, I guess we can close this since we can now just do `length . recoverBytes`.

PR #660, opened 2019-10-10
TxAux and MempoolPayload improvements

mrBliss:


cardano-ledger/src/Cardano/Chain/MempoolPayload.hs
```diff
@@ -52,6 +54,16 @@ instance ToCBOR MempoolPayload where
...
 
+instance ToCBOR (AMempoolPayload ByteString) where
+  toCBOR (MempoolTx tp) =
+    encodeListLen 2 <> toCBOR (0 :: Word8) <> encodePreEncoded (recoverBytes tp)
+  toCBOR (MempoolDlg dp) =
+    encodeListLen 2 <> toCBOR (1 :: Word8) <> encodePreEncoded (recoverBytes dp)
+  toCBOR (MempoolUpdateProposal upp) =
+    encodeListLen 2 <> toCBOR (2 :: Word8) <> encodePreEncoded (recoverBytes upp)
+  toCBOR (MempoolUpdateVote upv) =
+    encodeListLen 2 <> toCBOR (3 :: Word8) <> encodePreEncoded (recoverBytes upv)
```
intricate, line 65: :100:

intricate (approved):

cardano-ledger/src/Cardano/Chain/UTxO/TxAux.hs
```diff
@@ -40,14 +48,25 @@ import Cardano.Chain.UTxO.TxWitness (TxWitness)
 type TxAux = ATxAux ()
 
 mkTxAux :: Tx -> TxWitness -> TxAux
-mkTxAux tx tw = ATxAux (Annotated tx ()) (Annotated tw ())
+mkTxAux tx tw = ATxAux (Annotated tx ()) (Annotated tw ()) ()
+
+annotateTxAux :: TxAux -> ATxAux ByteString
+annotateTxAux ta = Lazy.toStrict . slice bs <$> ta'
```
mrBliss, line 54: This function should eventually be generalised to any `f ()`.

cardano-ledger/test/Test/Cardano/Chain/MempoolPayload/CBOR.hs
```diff
@@ -27,29 +43,78 @@ import Test.Options (TSGroup, TSProperty, concatTSGroups, eachOfTS)
 -- MempoolPayload
 --------------------------------------------------------------------------------
 
+fillInByteString
```
mrBliss, line 49: Is there a better place for this?

cardano-ledger/test/Test/Cardano/Chain/MempoolPayload/CBOR.hs
```diff
@@ -27,29 +43,78 @@ import Test.Options (TSGroup, TSProperty, concatTSGroups, eachOfTS)
...
+
+    dec :: Decoder s (f ByteString)
+    dec = fmap (LBS.toStrict . slice bytes) <$> fromCBOR
+
 goldenMempoolPayload :: Property
 goldenMempoolPayload = goldenTestCBOR
   exampleMempoolPayload
   "test/golden/cbor/mempoolpayload/MempoolPayload"
 
+goldenMempoolPayload' :: Property
```
mrBliss: What do you think about the duplication I have here?

cardano-ledger/test/Test/Cardano/Chain/MempoolPayload/CBOR.hs
intricate: It's not being used anywhere else, so I don't mind it being here.

intricate: Could probably just move it to another "section" in this src file to indicate that it's a sort of helper/utility function.

```
-- Helpers
```

or something.

cardano-ledger/test/Test/Cardano/Chain/MempoolPayload/CBOR.hs
```diff
@@ -23,6 +40,47 @@ import Test.Cardano.Crypto.Gen (feedPM)
...
+    bytes :: LByteString
+    bytes = serialize a
+
+    dec :: Decoder s (f ByteString)
+    dec = fmap (LBS.toStrict . slice bytes) <$> fromCBOR
+
+-- | Variant of 'goldenTestCBOR' that does not use the @'ToCBOR' (f ())@
+-- instance, but the @'ToCBOR' (f ByteString)@ instance. The latter instance
+-- allows reusing the annotation when serialising instead of reserialising
+-- from scratch.
```
intricate, line 64: :+1:

cardano-ledger/src/Cardano/Chain/MempoolPayload.hs
mrBliss: It would be even nicer if `ToCBOR` instances could be parameterised over `ByteString` and `()` so that they only have to be written once. Maybe the resulting code would be "too clever".

I guess only sum types suffer from this duplication.

cardano-ledger/src/Cardano/Chain/UTxO/TxAux.hs
```diff
@@ -40,14 +48,25 @@ import Cardano.Chain.UTxO.TxWitness (TxWitness)
...
+  where
+    bs  = serialize ta
+    ta' = unsafeDeserialize bs
 
 data ATxAux a = ATxAux
-  { aTaTx      :: !(Annotated Tx a)
-  , aTaWitness :: !(Annotated TxWitness a)
+  { aTaTx         :: !(Annotated Tx a)
+  , aTaWitness    :: !(Annotated TxWitness a)
+  , aTaAnnotation :: !a
```
dnadales, line 62: We have two styles of writing fields at the moment with data name prefix and without. I know the `cardano-ledger` is inconsistent in this regard, but we might want to settle on one style. As far as I remember @intricate introduced fields without the prefix.

dnadales (approved):
LGTM

cardano-ledger/src/Cardano/Chain/UTxO/TxAux.hs
intricate: Yeah, it is quite inconsistent. I think I'm okay with this one as the other fields are already prefixed and we also have an example of the same thing being done here (not that referencing a single example "proves" that this is okay):

https://github.com/input-output-hk/cardano-ledger/blob/efd59dbd1ceafb6ed56b0d8ab655a8e9de886202/cardano-ledger/src/Cardano/Chain/Block/Block.hs#L155

intricate:
Closes #576

PR #669, opened 2019-10-21
Expose missing modules in cardano-ledger-test

mrBliss:
Some of these generators are useful to other libraries.

The module list was copied from the `other-modules` field out of the
`cardano-ledger-test` test suite in `cardano-ledger.cabal`.


cardano-ledger/test/cardano-ledger-test.cabal
```diff
@@ -20,29 +20,63 @@ library
...
+                       Test.Cardano.Chain.Epoch.File
+
                        Test.Cardano.Chain.Genesis.Dummy
                        Test.Cardano.Chain.Genesis.Example
                        Test.Cardano.Chain.Genesis.Gen
                        Test.Cardano.Chain.Genesis.Json
+
+                       Test.Cardano.Chain.MempoolPayload.CBOR
+                       Test.Cardano.Chain.MempoolPayload.Example
+                       Test.Cardano.Chain.MempoolPayload.Gen
```
mrBliss, line 59: This is the module that prompted this PR. I plan to use it in https://github.com/input-output-hk/ouroboros-network/pull/1129.

mrBliss:
If you don't want all of them exposed, let me know which to remove from the list. I'm only interested in the generators.

cardano-ledger/test/cardano-ledger-test.cabal
dnadales: It seems all modules should be exposed unless there's a good reason not to do so. But I can only comment for the Byron specs. We might want to start using the `Internal` convention if we want to signal that certain modules aren't intended to be used for the clients of the library, but could be used if needed.

I like Edward Kmett's approach to this https://www.reddit.com/r/haskell/comments/2nkiiq/testing_internals_without_exposing_them/

dnadales (approved):
LGTM (Byron side only :) ). Let's wait on @JaredCorduan or @mgudemann

intricate (approved):

mgudemann:
@intricate @dnadales I don't see any problem

mgudemann (approved):

PR #677, opened 2019-11-07
Use common functionality from `iohk-nix` library for CI setup

dnadales:


nc6:
There's since been a big change to haskell.nix, so I think it makes sense to close this PR

PR #682, opened 2019-11-20
Bump dependencies

edsko:


erikd (approved):
Will need Nix voodoo but approved.

PR #687, opened 2019-11-27
Change AProtocolMagic and ProtocolMagicId

redxaxder:


crypto/src/Cardano/Crypto/ProtocolMagic.hs
```diff
@@ -35,40 +39,48 @@ import Cardano.Binary (Annotated(..), FromCBOR, ToCBOR)
 -- mhueschen: As part of CO-353 I am adding `getRequiresNetworkMagic` in
 -- order to pipe configuration to functions which must generate & verify
 -- Addresses (which now must be aware of `NetworkMagic`).
-data AProtocolMagic a = AProtocolMagic
-  { getAProtocolMagicId      :: !(Annotated ProtocolMagicId a)
-  , getRequiresNetworkMagic :: !RequiresNetworkMagic
+data ProtocolMagic = ProtocolMagic
+  { getProtocolMagicId       :: !ProtocolMagicId
+  , getRequiresNetworkMagic  :: !RequiresNetworkMagic
```
intricate: Not a big deal, but the `::` might be over-indented here (there's an extra space).

crypto/src/Cardano/Crypto/ProtocolMagic.hs
```diff
@@ -25,7 +29,7 @@ import qualified Data.Aeson as A
 import Data.Aeson ((.:), (.=))
 import Text.JSON.Canonical (FromJSON(..), JSValue(..), ToJSON(..), expected)
 
-import Cardano.Binary (Annotated(..), FromCBOR, ToCBOR)
+import Cardano.Binary (FromCBOR(..), ToCBOR(..), serialize', FromCBORAnnotated(..), withSlice', encodePreEncoded)
```
intricate: This line's too long and unsorted:

```suggestion
import Cardano.Binary 
  ( FromCBOR(..)
  , FromCBORAnnotated(..)
  , ToCBOR(..)
  , encodePreEncoded
  , serialize'
  , withSlice'
  )
```

crypto/src/Cardano/Crypto/ProtocolMagic.hs
```diff
@@ -35,40 +39,48 @@ import Cardano.Binary (Annotated(..), FromCBOR, ToCBOR)
...
   } deriving (Show, Eq, Generic)
-    deriving newtype (FromCBOR, ToCBOR)
-    deriving anyclass (NFData, NoUnexpectedThunks)
+    deriving anyclass (NFData)
+    deriving NoUnexpectedThunks via AllowThunksIn '["serializeProtocolMagicId"] ProtocolMagicId
+
+{-# COMPLETE ProtocolMagicId #-}
+pattern ProtocolMagicId :: Word32 -> ProtocolMagicId
+pattern ProtocolMagicId { unProtocolMagicId } <- ProtocolMagicId' unProtocolMagicId _
+  where
```
intricate:
```suggestion
 where
```

In `cardano-ledger`, `where`s have typically been indented like this:

https://github.com/input-output-hk/cardano-ledger/blob/48a69d7597d3eb26b47ba2fe65d3db381b27b5b0/cardano-ledger/src/Cardano/Chain/Block/Validation.hs#L388-L389

crypto/src/Cardano/Crypto/ProtocolMagic.hs
```diff
@@ -1,21 +1,25 @@
...
 {-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE DerivingVia                #-}
+{-# LANGUAGE DataKinds                  #-}
 {-# LANGUAGE FlexibleContexts           #-}
 {-# LANGUAGE FlexibleInstances          #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase                 #-}
+{-# LANGUAGE PatternSynonyms            #-}
+{-# LANGUAGE NamedFieldPuns             #-}
+{-# LANGUAGE ViewPatterns               #-}
```
intricate: I don't think you're using any `ViewPatterns`, so this can be removed.

crypto/src/Cardano/Crypto/ProtocolMagic.hs
intricate: Not sure if this is worth changing for this particular definition since a bunch of other `pattern` definitions in the codebase are now following the above convention.

crypto/src/Cardano/Crypto/ProtocolMagic.hs
```diff
@@ -1,21 +1,25 @@
...
 module Cardano.Crypto.ProtocolMagic
-  ( ProtocolMagicId(..)
-  , ProtocolMagic
-  , AProtocolMagic(..)
+  ( ProtocolMagicId(unProtocolMagicId)
+  , ProtocolMagic(..)
   , RequiresNetworkMagic(..)
   , getProtocolMagic
-  , getProtocolMagicId
+  , pattern ProtocolMagicId
```
intricate: Wouldn't it be cleaner to export this as part of the `ProtocolMagicId` data type export?

```
module Cardano.Crypto.ProtocolMagic
  ( ProtocolMagicId (ProtocolMagicId, unProtocolMagicId)
  , ...
```
That way, you wouldn't need to use `PatternSynonyms` along with the `pattern` keyword wherever importing it.

@redxaxder @dnadales @mhuesch: Thoughts?

crypto/src/Cardano/Crypto/ProtocolMagic.hs
intricate: Also, these `LANGUAGE` pragmas are now unsorted.

crypto/src/Cardano/Crypto/ProtocolMagic.hs
```diff
@@ -1,21 +1,25 @@
...
 {-# LANGUAGE DeriveGeneric              #-}
 {-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE DerivingVia                #-}
+{-# LANGUAGE DataKinds                  #-}
 {-# LANGUAGE FlexibleContexts           #-}
 {-# LANGUAGE FlexibleInstances          #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase                 #-}
+{-# LANGUAGE PatternSynonyms            #-}
+{-# LANGUAGE NamedFieldPuns             #-}
```
intricate: I don't think you're using any `NamedFieldPuns`, so this can be removed.

cardano-ledger/test/Test/Cardano/Chain/Update/Example.hs
```diff
@@ -51,7 +51,7 @@ import Cardano.Chain.Update
   , signVote
   , signProposal
   )
-import Cardano.Crypto (ProtocolMagicId(..), hash)
+import Cardano.Crypto (pattern ProtocolMagicId, hash)
```
intricate: If the `ProtocolMagicId` pattern were exported as part of the `ProtocolMagicId` data type, you would be able to write:
```
import Cardano.Crypto (ProtocolMagicId(..), hash)
```

intricate (changes requested):

cardano-ledger/test/Test/Cardano/Chain/Update/Example.hs
redxaxder: Oh really? I didn't know you could do that!

intricate: I only learned that a few days ago tbh

crypto/src/Cardano/Crypto/ProtocolMagic.hs
intricate: Looks like @redxaxder addressed this.

mhuesch (commented):
LGTM, I see that Luke has requested some changes so I'll just comment rather than approve.

cardano-ledger/src/Cardano/Chain/Block/Block.hs
```diff
@@ -13,16 +13,14 @@
 module Cardano.Chain.Block.Block
   (
   -- * Block
-    Block (blockHeader, blockBody, blockSerialized)
+    Block (Block, blockHeader, blockBody, blockSerialized)
 
   -- * Block Constructors
-  , pattern Block
```
intricate, line 19: :+1: Thanks for going through the other smart constructors in the project and adjusting this.

cardano-ledger/src/Cardano/Chain/Delegation/Payload.hs
```diff
@@ -10,8 +10,7 @@
 {-# LANGUAGE PatternSynonyms            #-}
```
intricate, line 10: We can remove this `PatternSynonyms`

cardano-ledger/src/Cardano/Chain/UTxO/TxWitness.hs
```diff
@@ -8,8 +8,7 @@
 {-# LANGUAGE PatternSynonyms     #-}
```
intricate, line 8: We can remove this `PatternSynonyms`

cardano-ledger/test/Test/Cardano/Chain/Delegation/Example.hs
```diff
@@ -1,3 +1,5 @@
+{-# LANGUAGE PatternSynonyms #-}
+
```
intricate: We can remove this `PatternSynonyms`

crypto/test/Test/Cardano/Crypto/Dummy.hs
```diff
@@ -1,32 +1,20 @@
+{-# LANGUAGE PatternSynonyms #-}
```
intricate: We can remove this `PatternSynonyms`

crypto/test/Test/Cardano/Crypto/Example.hs
```diff
@@ -1,4 +1,5 @@
 {-# OPTIONS_GHC -fno-warn-incomplete-uni-patterns #-}
+{-# LANGUAGE PatternSynonyms #-}
```
intricate: We can remove this `PatternSynonyms`

crypto/test/Test/Cardano/Crypto/Gen.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE PatternSynonyms #-}
```
intricate: We can remove this `PatternSynonyms`

crypto/test/Test/Cardano/Crypto/CBOR.hs
```diff
@@ -1,5 +1,6 @@
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE PatternSynonyms     #-}
```
intricate: We can remove this `PatternSynonyms`

intricate (changes requested):
Good work. Just looks like some leftover `PatternSynonyms` `LANGUAGE` pragmas to remove.

cardano-ledger/src/Cardano/Chain/Delegation/Payload.hs
redxaxder: I got rid of a bunch of them, but this one in particular has to stay because a pattern is being defined in this file. Same for the one below.

intricate (approved):
LGTM. Great work :100:

PR #689, opened 2019-12-01
Add tests for isRedeemAddress

erikd:
Add symmetrical Base58 encoder/decoder and deprecate fromCBORTextAddress which had no direct encoder partner.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -4,8 +4,8 @@
 # many reviewers which often leads to nobody feeling responsible to actually
 # perform the review.
 
-crypto/  @ruhatch @dnadales
-nix/     @ruhatch @nc6
-scripts/ @ruhatch @nc6
-src/     @ruhatch @dnadales
-test/    @ruhatch @dnadales
+crypto/  @dnadales
+nix/     @nc6
+scripts/ @nc6
+src/     @dnadales
+test/    @dnadales
--- a/cardano-ledger/src/Cardano/Chain/Common/Address.hs
+++ b/cardano-ledger/src/Cardano/Chain/Common/Address.hs
@@ -25,10 +25,12 @@ module Cardano.Chain.Common.Address
   , checkVerKeyAddress
   , checkRedeemAddress
 
-  -- * Encoding
+  -- * Encoding/Decoding
   , addrToBase58
   , toCBORAddr
   , toCBORAddrCRC32
+  , decodeAddressBase58
+  , encodeAddressBase58
 
   -- * Utilities
   , addrAttributesUnwrapped
@@ -90,7 +92,6 @@ import Cardano.Crypto.Signing
   ( VerificationKey
   , RedeemVerificationKey
   )
-
 
 -- | Hash of this data is stored in 'Address'. This type exists mostly
 --   for internal usage.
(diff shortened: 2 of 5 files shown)


CODEOWNERS
```diff
@@ -4,8 +4,8 @@
 # many reviewers which often leads to nobody feeling responsible to actually
 # perform the review.
 
-crypto/  @ruhatch @dnadales
-nix/     @ruhatch @nc6
-scripts/ @ruhatch @nc6
-src/     @ruhatch @dnadales
-test/    @ruhatch @dnadales
```
intricate, line 11: :disappointed: :+1:

cardano-ledger/src/Cardano/Chain/Common/Address.hs
```diff
@@ -196,6 +197,7 @@ addressF :: Format r (Address -> r)
 addressF = build
 
 -- | A function which decodes base58-encoded 'Address'
+{-# DEPRECATED fromCBORTextAddress "Use decodeAddressBase58 instead" #-}
```
intricate, line 200: @erikd: is this preferred over renaming this function to `decodeAddressBase58`?

cardano-ledger/test/Test/Cardano/Chain/Common/Address.hs
```diff
@@ -12,19 +12,27 @@ where
...
+    nm <- forAll genNetworkMagic
+    addr <- forAll (genAddressWithNM nm)
+    nm === addrNetworkMagic addr
+
+ts_prop_isRedeemAddress :: TSProperty
+ts_prop_isRedeemAddress =
+  withTestsTS 1000 . property $ do
+    addr <- forAll genAddress
+    cover 30 "Redeem Address" $ isRedeemAddress addr
+    cover 30 "Pubkey Address" $ isRedeemAddress addr
```
intricate: I'm a bit confused by this as these are the exact same but with different labels. 

Is one of these intended to be `not (isRedeemAddress addr)`?

cardano-ledger/test/Test/Cardano/Chain/Common/CBOR.hs
```diff
@@ -116,9 +119,22 @@ golden_Address4 :: Property
...
+    H.assert $ isRedeemAddress exampleAddress2
+
 ts_roundTripAddressCBOR :: TSProperty
 ts_roundTripAddressCBOR = eachOfTS 1000 genAddress roundTripsCBORBuildable
 
+ts_roundTripAddress :: TSProperty
+ts_roundTripAddress =
+  eachOfTS 1000 genAddress $ \ addr -> do
+    cover 30 "Redeem Address" $ isRedeemAddress addr
+    cover 30 "Pubkey Address" $ isRedeemAddress addr
```
intricate: Same here. Is one of these intended to be `not (isRedeemAddress addr)`?

intricate (changes requested):

cardano-ledger/src/Cardano/Chain/Common/Address.hs
erikd: Yes, this leaves the name and produces a warning wherever it is used. The old name can be removed later.

cardano-ledger/test/Test/Cardano/Chain/Common/Address.hs
erikd: Yeah, I screwed up. Will fix it.

intricate (approved):

PR #700, opened 2020-01-29
Move auxiliary definitions from ouroboros-consensus to cardano-ledger

### LINKED ISSUE #676: Move auxiliary definitions from ouroboros-consensus to cardano-ledger
Start with reviewing [https://github.com/input-output-hk/ouroboros-network/pull/1196/files](https://github.com/input-output-hk/ouroboros-network/pull/1196/files) and the Ouroboros.Consensus.Ledger.Byron.Aux module.

Firstly, the stuff in the Aux module can be grouped into three basic sections:

- The block validation interface
    - `validateBody`
    - `validateBlock`
    - `validateHeader`
    - `applyBoundary` - probably nothing here specifically worth testing
- The anachronistic protocol view interface
    - `applyScheduledDelegations`
- The mempool interface
    - `applyTxAux`
    - `applyCertificate`
    - `applyUpdateProposal`
    - `applyUpdateVote`

These sections represent the interface surface between the ledger and the consensus layer. Currently only the block validation interface is particularly specified. So we should also consider what properties hold of the others. We should also consider whether there are any other points of interface - for example, the consensus layer may wish to get certain "global constants" from the ledger - an obvious one is the map of epoch to epoch lengths.

A first task would be to move this module to `cardano-ledger`, make clear the distinct interface sections, and seek to include any further interface needed. We might want to split the interface into distinct modules to better represent their use - mempool, anachronism, block validation etc.

Where possible, the interface should define minimal logic, but there may be some sensible unit tests which could be written. In some cases, the existing conformance testing could be expanded to cover parts of the interface:

- `validateBlock`
- Parts of the mempool interface (e.g. we can check that transactions which get validated correctly as part of block `n` can be added to the mempool atop a chain stopping at block `n-1`.

We may wish to consider moving more of the anachronistic view to the ledger (and potentially specifying it in the formal spec). This would then allow more testing to cover that (which would be great, since it encodes some potentially subtle logic). The testing could be done along the lines of the time travel property as already encoded in the chain level spec.

redxaxder:
This closes #676

### DIFF
--- a/cardano-ledger/src/Cardano/Chain/Byron/API.hs
+++ b/cardano-ledger/src/Cardano/Chain/Byron/API.hs
@@ -1,19 +1,19 @@
-{-# LANGUAGE DeriveAnyClass      #-}
-{-# LANGUAGE DeriveFunctor       #-}
-{-# LANGUAGE DeriveGeneric       #-}
-{-# LANGUAGE FlexibleContexts    #-}
-{-# LANGUAGE FlexibleInstances   #-}
-{-# LANGUAGE LambdaCase          #-}
-{-# LANGUAGE NamedFieldPuns      #-}
-{-# LANGUAGE OverloadedStrings   #-}
-{-# LANGUAGE RankNTypes          #-}
-{-# LANGUAGE RecordWildCards     #-}
-{-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TypeApplications    #-}
-{-# LANGUAGE TypeFamilies        #-}
+{-# LANGUAGE DeriveAnyClass           #-}
+{-# LANGUAGE DeriveFunctor            #-}
+{-# LANGUAGE DeriveGeneric            #-}
+{-# LANGUAGE DisambiguateRecordFields #-}
+{-# LANGUAGE FlexibleContexts         #-}
+{-# LANGUAGE FlexibleInstances        #-}
+{-# LANGUAGE LambdaCase               #-}
+{-# LANGUAGE NamedFieldPuns           #-}
+{-# LANGUAGE OverloadedStrings        #-}
+{-# LANGUAGE RankNTypes               #-}
+{-# LANGUAGE ScopedTypeVariables      #-}
+{-# LANGUAGE TypeApplications         #-}
+{-# LANGUAGE TypeFamilies             #-}
 
 -- | Auxiliary definitions to make working with the Byron ledger easier
-module Cardano.Chain.Byron.Auxiliary (
+module Cardano.Chain.Byron.API (
     -- * Extract info from genesis config
     allowedDelegators
   , boundaryBlockSlot
@@ -213,8 +213,7 @@ applyChainTick :: Gen.Config
                -> CC.ChainValidationState
                -> CC.ChainValidationState
 applyChainTick cfg slotNo cvs = cvs {
-      CC.cvsLastSlot        = slotNo
-    , CC.cvsUpdateState     = CC.epochTransition
+      CC.cvsUpdateState     = CC.epochTransition
                                 (mkEpochEnvironment cfg cvs)
                                 (CC.cvsUpdateState cvs)
                                 slotNo
@@ -580,7 +579,7 @@ fromCBORABlockOrBoundaryHdr epochSlots = do
     fromCBOR @Word >>= \case
       0 -> ABOBBoundaryHdr <$> CC.fromCBORABoundaryHeader
       1 -> ABOBBlockHdr    <$> CC.fromCBORAHeader epochSlots
-      t -> panic $ "Unknown tag in encoded HeaderOrBoundary" <> show t
+      t -> fail $ "Unknown tag in encoded HeaderOrBoundary" <> show t
 
 -- | The analogue of 'Data.Either.either'
 aBlockOrBoundaryHdr :: (CC.AHeader         a -> b)
--- a/cardano-ledger/src/Cardano/Chain/Byron/Auxiliary.hs
+++ b/cardano-ledger/src/Cardano/Chain/Byron/Auxiliary.hs
@@ -26,7 +26,6 @@ module Cardano.Chain.Byron.Auxiliary (
   , applyChainTick
   , validateBlock
   , validateBoundary
-  , applyScheduledDelegations
   , previewDelegationMap
     -- * Applying transactions
   , ApplyMempoolPayloadErr(..)
@@ -56,14 +55,10 @@ import qualified Codec.CBOR.Read as CBOR
 import qualified Codec.CBOR.Write as CBOR
 import           Control.Monad.Except
 import           Control.Monad.Reader
-import           Data.Bimap (Bimap)
-import qualified Data.Bimap as Bimap
 import           Data.ByteString (ByteString)
 import qualified Data.ByteString.Lazy as Lazy
 import           Data.Either (isRight)
-import qualified Data.Foldable as Foldable
 import           Data.Sequence (Seq)
-import qualified Data.Sequence as Seq
 import           Data.Set (Set)
 import qualified Data.Set as Set
 import qualified Data.Text as T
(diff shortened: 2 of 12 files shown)


cardano-ledger/src/Cardano/Chain/Delegation/Validation/Interface.hs
```diff
@@ -139,30 +141,13 @@ updateDelegation
...
-      { Scheduling.scheduledDelegations = Seq.filter
-        ((currentSlot + 1 <=) . Scheduling.sdSlot)
-        delegations
-      , Scheduling.keyEpochDelegations = Set.filter
-        ((>= currentEpoch) . fst)
-        keyEpochs
-      }
-
-  pure $ State {schedulingState = ss', activationState = as}
+  pure $ tickDelegation currentEpoch currentSlot
```
dnadales, line 149: The comment in the function header "This corresponds to the `DELEG` rule from the Byron ledger specification" We should document this, and probably make a ticket on `cardano-ledger-specs` to adapt the Byron spec.

dnadales (approved):

cardano-ledger/src/Cardano/Chain/Byron/Auxiliary.hs
```diff
@@ -0,0 +1,639 @@
...
+{-# LANGUAGE NamedFieldPuns      #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE RankNTypes          #-}
+{-# LANGUAGE RecordWildCards     #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications    #-}
+{-# LANGUAGE TypeFamilies        #-}
+
+-- | Auxiliary definitions to make working with the Byron ledger easier
+module Cardano.Chain.Byron.Auxiliary (
```
nc6: The naming of `Auxiliary` makes sense when this is inside ouroboros-consensus, but I think less so when moved to `cardano-ledger` - these things aren't auxiliary now, they're the principal interface. So I would go with `Cardano.Chain.Byron.API`

cardano-ledger/src/Cardano/Chain/Delegation/Validation/Interface.hs
dnadales: Oh, now I see that the activation state is still updated here. Sorry.

nc6 (approved):
This looks good, and I see no reason not to merge it, but I also wonder if we can go further here. What surface are we now exposing to `ouroboros-consensus`? Things like `tickDelegation` suggests to me that this surface is still too big - I'd still aim for something like the Shelley API which abstracts away such concerns.

cardano-ledger/src/Cardano/Chain/Byron/API.hs
```diff
@@ -0,0 +1,639 @@
...
+  deriving (Eq, Show, Functor, Generic, NoUnexpectedThunks)
+
+fromCBORABlockOrBoundaryHdr :: CC.EpochSlots
+                            -> Decoder s (ABlockOrBoundaryHdr ByteSpan)
+fromCBORABlockOrBoundaryHdr epochSlots = do
+    enforceSize "ABlockOrBoundaryHdr" 2
+    fromCBOR @Word >>= \case
+      0 -> ABOBBoundaryHdr <$> CC.fromCBORABoundaryHeader
+      1 -> ABOBBlockHdr    <$> CC.fromCBORAHeader epochSlots
+      t -> panic $ "Unknown tag in encoded HeaderOrBoundary" <> show t
```
edsko: This should not be `panic`, but `fail` (from the `MonadFail` instance of `Decoder`).

redxaxder:
bors retry

cardano-ledger/src/Cardano/Chain/Byron/API.hs
```diff
@@ -0,0 +1,639 @@
+{-# LANGUAGE DeriveAnyClass      #-}
+{-# LANGUAGE DeriveFunctor       #-}
+{-# LANGUAGE DeriveGeneric       #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE FlexibleInstances   #-}
+{-# LANGUAGE LambdaCase          #-}
+{-# LANGUAGE NamedFieldPuns      #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE RankNTypes          #-}
+{-# LANGUAGE RecordWildCards     #-}
```
intricate:
```suggestion
```

I don't think we actually use any `RecordWildCards` in here.

cardano-ledger/test/Test/Cardano/Chain/Byron/API.hs
```diff
@@ -0,0 +1,114 @@
...
+module Test.Cardano.Chain.Byron.API
+  ( genApplyMempoolPayloadErr
+  , ts_roundTripApplyMempoolPayloadErrCompat
+  , ts_scheduledDelegations
+  , tests
+  )
+  where
+
+import Cardano.Prelude
+  -- import Test.Cardano.Prelude
```
intricate:
```suggestion
```

cardano-ledger/test/Test/Cardano/Chain/Byron/API.hs
```diff
@@ -0,0 +1,114 @@
...
+      (cvs, _) = either (panic . show) identity $
+        foldM
+          (elaborateAndUpdate config)
+          (initialState, initialAbstractToConcreteIdMaps)
+          (STS.preStatesAndSignals STS.OldestFirst sampleTrace)
+   in (cvs, config)
+
+-- | getDelegationMap . applyChainTick slot == previewDelegationMap slot
+ts_scheduledDelegations :: TSProperty
+ --FIXME: This is a failing test.
```
intricate: Does this comment still apply? If not:

```suggestion
```

redxaxder:
bors retry

cardano-ledger/src/Cardano/Chain/Byron/API.hs
```diff
@@ -0,0 +1,639 @@
...
+-- updates are conform spec. See
+--
+-- <https://github.com/input-output-hk/cardano-ledger-specs/issues/1046>
+-- <https://github.com/input-output-hk/ouroboros-network/issues/1291>
+applyChainTick :: Gen.Config
+               -> CC.SlotNumber
+               -> CC.ChainValidationState
+               -> CC.ChainValidationState
+applyChainTick cfg slotNo cvs = cvs {
+      CC.cvsLastSlot        = slotNo
```
edsko: @redxaxder , since Bors is timed out anyway, could I ask you to make a change here and remove the assignment to `cvsLastSlot` before merging? I will make the same change to `ouroboros-network`.

cardano-ledger/src/Cardano/Chain/Byron/API.hs
```diff
@@ -0,0 +1,639 @@
+{-# LANGUAGE DeriveAnyClass      #-}
+{-# LANGUAGE DeriveFunctor       #-}
+{-# LANGUAGE DeriveGeneric       #-}
+{-# LANGUAGE DisambiguateRecordFields #-}
```
mrBliss, line 4: Run `stylish-haskell`

cardano-ledger/src/Cardano/Chain/Delegation/Validation/Interface.hs
```diff
@@ -174,3 +159,37 @@ updateDelegation env is certificates = do
...
+-- | Perform delegation update without adding certificates
+tickDelegation :: EpochNumber -> SlotNumber -> State -> State
+tickDelegation currentEpoch currentSlot =
+  prune . activateDelegations currentSlot
+  where
+   prune s =
+     let ss' = pruneScheduledDelegations currentEpoch currentSlot (schedulingState s)
+      in s{ schedulingState = ss'}
+
+-- Activate certificates up to this slot
```
mrBliss: Make this a haddock comment.

cardano-ledger/src/Cardano/Chain/Delegation/Validation/Interface.hs
```diff
@@ -174,3 +159,37 @@ updateDelegation env is certificates = do
...
+activateDelegations :: SlotNumber -> State -> State
+activateDelegations currentSlot s@(State ss as) =
+  let Scheduling.State delegations _keyEpochs = ss
+      as' = foldl Activation.activateDelegation as
+        (Seq.filter ((<= currentSlot) . Scheduling.sdSlot) delegations)
+   in s { activationState = as' }
+
+-- Remove stale values from 'Scheduling.State'
```
mrBliss: Make this a haddock comment.

cardano-ledger/src/Cardano/Chain/Byron/API.hs
mrBliss: @edsko This will have to go.

PR #713, opened 2020-02-05
Add generic `ToJSON` instances 

Jimbo4350:
Generic `ToJSON` instances have been added to debug blocks, txs, delegation certificates and update proposals. These instances are necessary when decoding `CBOR` files and re-encoding the resulting haskell types as `JSON`.

### DIFF
--- a/cardano-ledger/src/Cardano/Chain/Common/Attributes.hs
+++ b/cardano-ledger/src/Cardano/Chain/Common/Attributes.hs
@@ -30,7 +30,9 @@ where
 import Cardano.Prelude
 import qualified Prelude
 
+import Data.Aeson (ToJSON(..))
 import qualified Data.ByteString.Lazy as LBS
+import qualified Data.ByteString.Lazy.Char8 as LChar8
 import qualified Data.Map.Strict as M
 import Formatting (bprint, build, int)
 import Formatting.Buildable (Buildable)
@@ -49,7 +51,6 @@ import Cardano.Binary
   , dropWord8
   )
 
-
 -- | Representation of unparsed fields in Attributes. Newtype wrapper is used
 --   for clear backward compatibility between previous representation (which was
 --   just a single ByteString) during transition from Store to CBOR.
@@ -58,6 +59,11 @@ newtype UnparsedFields =
   deriving (Eq, Ord, Show, Generic)
   deriving newtype HeapWords
   deriving anyclass (NFData, NoUnexpectedThunks)
+
+-- Used for debugging purposes only
+instance ToJSON UnparsedFields where
+    toJSON (UnparsedFields map') = toJSON $ M.map LChar8.unpack map'
+
 
 fromUnparsedFields :: UnparsedFields -> Map Word8 LBS.ByteString
 fromUnparsedFields (UnparsedFields m) = m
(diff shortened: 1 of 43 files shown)


cardano-ledger/src/Cardano/Chain/Common/Attributes.hs
```diff
@@ -59,6 +64,17 @@ newtype UnparsedFields =
   deriving newtype HeapWords
   deriving anyclass (NFData, NoUnexpectedThunks)
 
+-- Used for debugging purposes only
+instance ToJSON UnparsedFields where
+
+-- Used for debugging purposes only
+instance ToJSON LBS.ByteString where
```
erikd: I presume this is the reason for the `-fno-warn-orphans`.

Probably best to define all these orphan instances in their own module so that we do not have orphans and non-orphans in the same file and *only* have `-fno-warn-orphans` in modules that only contain orphans.

erikd (changes requested):

cardano-ledger/src/Cardano/Chain/Common/Attributes.hs
Jimbo4350: @erikd changes made

cardano-ledger/src/Cardano/Chain/Orphans.hs
```diff
@@ -0,0 +1,20 @@
...
+
+import Data.Aeson (ToJSON(..), Value(String))
+import qualified Data.ByteString.Char8 as Char8
+import qualified Data.ByteString.Lazy as LBS
+import qualified Data.ByteString.Lazy.Char8 as LChar8
+
+-- Used for debugging purposes only
+-- Required for `UnparsedFields` `ToJSON` instance.
+instance ToJSON LBS.ByteString where
+  toJSON bs = String . toS $ LChar8.unpack bs
```
nc6: Creating a ToJSON orphan instance for something as generic as bytesting makes me a little worried that this will come around to bite us later. If it's needed just for one type, can we not just fold the bytestring encoder into that type?

cardano-ledger/src/Cardano/Chain/Orphans.hs
```diff
@@ -0,0 +1,20 @@
...
+
+-- Used for debugging purposes only
+-- Required for `HDAddressPayload` `ToJSON` instance.
+instance ToJSON Char8.ByteString where
+  toJSON bs = String . toS $ Char8.unpack bs
```
nc6: Ditto

nc6 (commented):
Comment only: I usually don't care about orphans, but I'm a little worried here where the orphan is for something as generic as `ToJSON ByteString`. I've suggested another approach; let me know if it's plausible. If not, I'll approve the PR

cardano-ledger/src/Cardano/Chain/Orphans.hs
Jimbo4350: @nc6 fair point, I'll make the changes.

Jimbo4350:
> Comment only: I usually don't care about orphans, but I'm a little worried here where the orphan is for something as generic as `ToJSON ByteString`. I've suggested another approach; let me know if it's plausible. If not, I'll approve the PR

@nc6 Done

erikd (approved):

nc6 (approved):

PR #716, opened 2020-02-06
add currentSlot argument to mkDelegationEnvironment et al

### LINKED ISSUE #715: Use current slot instead of latest block's slot when validating delegation certificates
## Background

This line is incorrect. The semantics of `cvsLastSlot` were recently refined such that it is now _always_ the slot of the latest block on the chain, which need not be the _current_ slot.

https://github.com/input-output-hk/cardano-ledger/blob/98048d95823b72e409b07481761b8a59c892c56c/cardano-ledger/src/Cardano/Chain/Byron/API.hs#L358

If I understand correctly, that code is only reached from a `applyMempoolPayload` call, and since that function is used within `ouroboros-consensus` to flush invalid transactions (including _expired_ delegation certificate transactions) from the mempool, it must actually use the _current_ slot.

## Proposed Solution

It suffices to simply add a `currentSlot` argument to `applyMempoolPayload`, which will be provided by the recently-added `tickedSlotNo` in the call in the `ouroboros-consensus` repo/library.

## Next Action
- [ ] Establish a local repro? (My current repro is a `test-consensus` failure; I'm not sure how to distill it down for inclusion into this repo).
- [x] PR here to fix: just add that `currentSlot` argument. (Edit: I've opened #716 for this.)
- [x] Update `ouroboros-consensus` (Should this step be a separate `ouroboros-consensus` ticket?)

nfrisby:
This fixes the implementation for #715. But it doesn't add a test/repro.

Please review the PR labels too.

### DIFF
--- a/cardano-ledger/src/Cardano/Chain/Byron/API.hs
+++ b/cardano-ledger/src/Cardano/Chain/Byron/API.hs
@@ -337,25 +337,21 @@ mkUtxoEnvironment cfg cvs = Utxo.Environment {
     updateState   = CC.cvsUpdateState cvs
 
 mkDelegationEnvironment :: Gen.Config
-                        -> CC.ChainValidationState
+                        -> CC.SlotNumber
                         -> D.Iface.Environment
-mkDelegationEnvironment cfg cvs = D.Iface.Environment {
+mkDelegationEnvironment cfg currentSlot = D.Iface.Environment {
       D.Iface.protocolMagic     = getAProtocolMagicId protocolMagic
     , D.Iface.allowedDelegators = allowedDelegators cfg
     , D.Iface.k                 = k
-      -- By rights the 'currentEpoch' for checking a delegation certificate
-      -- should be the epoch of the block in which the delegation certificate
-      -- is included. However, we don't have such a block yet, and so we can
-      -- only use the epoch from the ledger state. This does mean that we might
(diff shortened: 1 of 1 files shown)


cardano-ledger/src/Cardano/Chain/Byron/API.hs
```diff
@@ -337,25 +337,21 @@ mkUtxoEnvironment cfg cvs = Utxo.Environment {
...
     , D.Iface.k                 = k
-      -- By rights the 'currentEpoch' for checking a delegation certificate
-      -- should be the epoch of the block in which the delegation certificate
-      -- is included. However, we don't have such a block yet, and so we can
-      -- only use the epoch from the ledger state. This does mean that we might
-      -- say a transaction is valid now, but will become invalid by the time we
-      -- actually include it in a block.
+      -- The @currentSlot@/@currentEpoch@ for checking a delegation certificate
+      -- must be that of the block in which the delegation certificate is/will
+      -- be included.
```
edsko, line 348: :+1:

edsko (approved):
Great, let's get this merged asap.

nfrisby:
The current phase of CI has been building for ~6h~ 25h (!). Is this normal? This is my first time opening PR on this repo.

nc6 (approved):

nfrisby:
"Passed in 2d 21h" 🤯

PR #733, opened 2020-03-03
Update with changes to cardano-ledger-specs

nc6:
This merely updates to the latest version of cardano-ledger-specs

### DIFF
--- a/cardano-ledger/test/Test/Cardano/Chain/Block/Model.hs
+++ b/cardano-ledger/test/Test/Cardano/Chain/Block/Model.hs
@@ -77,7 +77,7 @@ import Cardano.Chain.Update
 import Cardano.Chain.Update.Validation.Interface (adoptedProtocolParameters)
 
 import Cardano.Ledger.Spec.STS.UTXO (UTxOEnv(..))
-import Cardano.Ledger.Spec.STS.UTXOW (coverUtxoFailure, tamperedTxWitsList)
+import Cardano.Ledger.Spec.STS.UTXOW (coverUtxoFailure, tamperedTxList)
 import Cardano.Spec.Chain.STS.Rule.BBody
   ( PredicateFailure(InvalidBlockSize, InvalidDelegationHash,
                  InvalidUpdateProposalHash, InvalidUtxoHash)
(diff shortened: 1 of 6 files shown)


cardano-ledger/test/Test/Cardano/Chain/UTxO/Model.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
```
intricate, line 1:
```suggestion
{-# LANGUAGE BangPatterns      #-}
```

cardano-ledger/test/Test/Cardano/Chain/Block/Model.hs
```diff
@@ -410,7 +410,7 @@ ts_prop_invalidTxWitsAreRejected =
     block <- sigGenChain NoGenDelegation NoGenUTxO NoGenUpdate env st
     let (_slot, _sgs, _h, utxoSt, _delegSt, _upiSt) = st
         utxoEnv = UTxOEnv { utxo0 = utxo, pps = pparams }
-    txWitsList <- tamperedTxWitsList utxoEnv utxoSt
+    txWitsList <- tamperedTxList utxoEnv utxoSt
```
intricate, line 413: Should this binding perhaps be renamed or is it still accurate?

```suggestion
    txList <- tamperedTxList utxoEnv utxoSt
```

nc6: There are still witnesses there

cardano-ledger/test/Test/Cardano/Chain/UTxO/Model.hs
```diff
@@ -13,7 +14,7 @@ module Test.Cardano.Chain.UTxO.Model
   )
 where
 
-import Cardano.Prelude hiding (trace)
+import Cardano.Prelude hiding (trace, traceM, traceShow)
```
intricate: It seems that these additional imports are redundant:

```suggestion
import Cardano.Prelude hiding (trace)
```

intricate, line 17: It seems that these additional hidden symbols are redundant:

```suggestion
import Cardano.Prelude hiding (trace)
```

erikd (approved):

Issue #725, opened 2020-02-17
registeredSoftwareUpdatesProposals is supposed to be preserved


ishaq:
## Background
During audit, we noticed that https://github.com/input-output-hk/cardano-ledger/blob/bccf097bc589d082600132b63e9e787071568e12/cardano-ledger/src/Cardano/Chain/Update/Validation/Interface.hs#L497 sets `registeredSoftwareUpdateProposals` to `M.empty`.  According to the spec (Ledger rule 49, Fig. 45), it should be preserved.

nc6:
Giving this high priority; discrepancy needs to be at least investigated.

nc6:
Addressed through https://github.com/input-output-hk/cardano-ledger-specs/pull/1279 - we bring the formal spec in line with the exec spec and the implementation.

PR #738, opened 2020-03-09
Preserve software updates through epoch boundaries.

nc6:
Software update proposals do not trigger on the epoch boundary, they get confirmed as soon as we have a minimum number of endorsements based on block issuance. Zeroing out the software updates on an epoch change will inhibit this.

This is also in line with the current spec

Addresses #725


dnadales (approved):
:+1:

PR #744, opened 2020-03-11
Disallow duplicate votes

nc6:
Addresses #739 

It is now invalid to cast a duplicate vote. We bump the dependency on `cardano-ledger-specs` to include https://github.com/input-output-hk/cardano-ledger-specs/pull/1287, which makes the corresponding change there.

PR will probably need to be updated when that PR is merged.

### DIFF
--- a/cardano-ledger/src/Cardano/Chain/Update/Validation/Voting.hs
+++ b/cardano-ledger/src/Cardano/Chain/Update/Validation/Voting.hs
@@ -78,6 +78,7 @@ data Error
   = VotingInvalidSignature
   | VotingProposalNotRegistered UpId
   | VotingVoterNotDelegate KeyHash
+  | VotingVoteAlreadyCast KeyHash
   deriving (Eq, Show)
 
 instance ToCBOR Error where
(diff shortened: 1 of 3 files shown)


nc6:
Confirm that mainnet mirror validates against this rule change.

cardano-ledger/src/Cardano/Chain/Update/Validation/Voting.hs
```diff
@@ -171,6 +178,12 @@ registerVote pm vre votes vote = do
     Nothing -> throwError (VotingVoterNotDelegate voter)
     Just d  -> pure d
 
+  -- Check that the vote has not already been cast
```
dnadales, line 181: Isn't "already" redundant here? I'd say something like "check that there is no vote for the same proposal id with the same key".

dnadales (approved):

dnadales:
Damn. If the tests pass this means that we didn't generate test cases in which the same key votes twice for the same proposal :(

nc6:
@dnadales Not necessarily, since I updated (and bumped!) the specs first!

dnadales:
> @dnadales Not necessarily, since I updated (and bumped!) the specs first!

Ohhhh I wanted to see a failure :D But that's also great news :)

PR #746, opened 2020-03-12
Avoid thunks in candidateProtocolUpdates

mrBliss:
This was popping up as an unexpected thunk in the consensus tests.

This is field is a list, so the bang doesn't have much effect, so we must
force its elements.


mrBliss:
We first have to wait for https://github.com/input-output-hk/ouroboros-network/pull/1793 to be merged.

nc6 (approved):

Issue #707, opened 2020-02-05
Test the new .API module


edsko:
Although @redxaxder moved the consensus auxiliary glue layer around the Byron implementation to the new `.API` module in `cardano-ledger`, these new definitions are not yet incorporated in the _testing_ of the Byron ledger, which was the primary motivation for moving them in the first place: until these are tested, we cannot be sure that the Byron integration is correct.

nfrisby:
Issue #715 is the kind of bug we hope this Issue will flush out.

nc6:
We could test this with the following property:

A transaction tx should validate in the mempool in slot n iff a header-valid block containing only that transaction validates when issues in slot n

This should hopefully catch issues such as #715 (since the slot would have been incorrect and thus in some cases would not have validated in one place), whilst also enforcing a general relation between mempool operations and block validation.

PR #748, opened 2020-03-16
[#707] Test validation in applyMempoolPayload

redxaxder:


cardano-ledger/test/Test/Cardano/Chain/Byron/API.hs
```diff
@@ -103,10 +134,92 @@ ts_chainTick :: TSProperty
...
+  let (cvs, config, _) = setupChainValidationState sampleTrace
       n0 = unSlotNumber . cvsLastSlot $ cvs
       k = unSlotCount . configSlotSecurityParam $ config
   n2 <- forAll $ Gen.word64 (Range.linear n0 (n0 + 2*k))
   n1 <- forAll $ Gen.word64 (Range.linear n0 n2)
   let tick n = applyChainTick config (SlotNumber n)
   (tick n2 . tick n1) cvs === tick n2 cvs
+
+-- | A transaction should validate in the mempool when
+-- | it validates in a block.
```
nc6:
```suggestion
-- | A transaction should validate in the mempool at a given slot when
--   it validates in a block issued for that same slot.
```

nc6: Also maybe should say something about "given a shared chain state", because obviously this isn't true in general (e.g. if a block has come in the meantime, it could have spent an output)

cardano-ledger/test/Test/Cardano/Chain/Byron/API.hs
```diff
@@ -103,10 +134,92 @@ ts_chainTick :: TSProperty
...
+        (_stateSlot, allowedDelegators, h, utxoState, diState, upiState)
+          = STS.lastState sampleTrace
+      (cvs, config, abstractToConcreteIdMaps) = setupChainValidationState sampleTrace
+      pm = Genesis.configProtocolMagicId config
+      txIdMap = transactionIds abstractToConcreteIdMaps
+      upIdMap = proposalIds abstractToConcreteIdMaps
+  genSlot <- forAll $ (slot +) <$> Gen.integral (Range.linear 1 20)
+  let nextSlot = Spec.Slot genSlot
+      addAnnotation :: MempoolPayload -> AMempoolPayload ByteString
+      addAnnotation = reAnnotateUsing toCBOR fromCBOR
```
nc6: Given this is pure utility, I would prefer it defined in a separate `where` block rather than inline in monadic code.

cardano-ledger/test/Test/Cardano/Chain/Byron/API.hs
```diff
@@ -103,10 +134,92 @@ ts_chainTick :: TSProperty
...
+
+  let pparams = Spec.protocolParameters upiState
```
nc6: What's the rationale behind the different `let` groupings?

cardano-ledger/test/Test/Cardano/Chain/Byron/API.hs
```diff
@@ -103,10 +134,92 @@ ts_chainTick :: TSProperty
...
+  sampleTrace <- forAll $ STS.trace @CHAIN () traceLength
+  let (Spec.Slot slot,utxo0,_,_,blockCount) = STS._traceEnv sampleTrace
+      abstractChainState@
+        (_stateSlot, allowedDelegators, h, utxoState, diState, upiState)
+          = STS.lastState sampleTrace
+      (cvs, config, abstractToConcreteIdMaps) = setupChainValidationState sampleTrace
+      pm = Genesis.configProtocolMagicId config
+      txIdMap = transactionIds abstractToConcreteIdMaps
+      upIdMap = proposalIds abstractToConcreteIdMaps
+  genSlot <- forAll $ (slot +) <$> Gen.integral (Range.linear 1 20)
```
nc6, line 159: I find the `forAll`s nested in here a little worrying. Feel free to tell me I shouldn't be worried, but if they're just there to print things then it would be better to remove them.

nc6 (changes requested):
General looks good, but I think it could do with a bit of cleaning.

cardano-ledger/test/Test/Cardano/Chain/Byron/API.hs
redxaxder: Most of them are there for random value selection.

We have `forAll :: Gen a -> PropertyT m a`.

I think a couple of the ones I used for debugging are still in there. I'll replace them.

I took a closer look at Hedgehog, and it seems like the ones used for selection could be scrubbed as well if you really wanted. I suspect this isn't related to your worry -- I mention this because I'm not sure I understand it.

The implementation is basically `PropertyT . lift . generalize :: Gen a -> PropertyT m a` which we could use with less indirection.

cardano-ledger/test/Test/Cardano/Chain/Byron/API.hs
redxaxder: There isn't one.

nc6 (approved):

PR #750, opened 2020-03-16
Add mainnetProtocolMagicId constant

erikd:


nc6:
@erikd What's the reason for having this here? Is it not in the genesis JSON?

dnadales (approved):

erikd:
It is in the Genesis JSON but for the crdano-api it would be nice to have to avoid reading it.

Issue #745, opened 2020-03-11
Reduce the window in which endorsements are counted


edsko:
This is the implementation side of the change in the specification described in https://github.com/input-output-hk/cardano-ledger-specs/issues/1288.

nc6:
Will probably fix this and #743 together, since they're quite related.

PR #753, opened 2020-03-18
Update endorsement count rules.

nc6:
This addresses #743 and #745. Endorsements must now be processed 4k
slots before the start of an epoch, rather than 2k slots before the
first block in the next epoch.

I've also tried to tidy the PVBUMP implementation to bring it closer to
the specs.

### DIFF
--- a/cardano-ledger/src/Cardano/Chain/ProtocolConstants.hs
+++ b/cardano-ledger/src/Cardano/Chain/ProtocolConstants.hs
@@ -4,6 +4,7 @@
 
 module Cardano.Chain.ProtocolConstants
   ( kSlotSecurityParam
+  , kUpdateStabilityParam
   , kChainQualityThreshold
   , kEpochSlots
   )
@@ -20,6 +21,12 @@ import Cardano.Chain.Slotting.SlotCount (SlotCount(..))
 --   property. It's basically @blkSecurityParam / chainQualityThreshold@.
 kSlotSecurityParam :: BlockCount -> SlotCount
 kSlotSecurityParam = SlotCount . (*) 2 . unBlockCount
+
+-- | Update stability parameter expressed in number of slots. This is the time
+--   between an protocol version update receiving its final endorsement and
+--   being accepted, and is set to double the security param.
+kUpdateStabilityParam :: BlockCount -> SlotCount
+kUpdateStabilityParam = SlotCount . (*) 4 . unBlockCount
 
 -- | Minimal chain quality (number of blocks divided by number of
 --   slots) necessary for security of the system.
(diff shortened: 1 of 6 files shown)


cardano-ledger/src/Cardano/Chain/Update/Validation/Interface.hs
```diff
@@ -501,23 +507,21 @@ registerEpoch env st lastSeenEpoch = do
...
-    subEnv = PVBump.Environment k currentSlot candidateProtocolUpdates
+    subEnv = PVBump.Environment k firstSlot candidateProtocolUpdates
 
     subSt =
       PVBump.State
-        currentEpoch
         adoptedProtocolVersion
         adoptedProtocolParameters
 
+    firstSlot = epochFirstSlot (kEpochSlots k) lastSeenEpoch
```
dnadales: Maybe it's too long, but I'd call this variable something like `firstSlotOfLastSeenEpoch`, or `epochFirstSlot'`.

cardano-ledger/src/Cardano/Chain/Update/Validation/Interface/ProtocolVersionBump.hs
```diff
@@ -23,44 +23,39 @@ import Cardano.Chain.Update.Validation.Endorsement
...
-  , nextProtocolVersion       :: !ProtocolVersion
+  { nextProtocolVersion       :: !ProtocolVersion
   , nextProtocolParameters    :: !ProtocolParameters
   }
 
 -- | Change the protocol version when an epoch change is detected, and there is
--- a candidate protocol update that was confirmed at least @2 * k@ slots ago,
--- where @k@ is the chain security parameter.
+-- a candidate protocol update that was confirmed at least @4 * k@ slots before
+-- the start of the epoch, where @k@ is the chain security parameter.
```
dnadales: Maybe we should add a small comment or refer to the issue that required the stability of the confirmation to have occurred `4k` slots before the start of the next epoch.

dnadales:
```suggestion
-- the start of the new epoch, where @k@ is the chain security parameter.
```

cardano-ledger/src/Cardano/Chain/Slotting/EpochSlots.hs
```diff
@@ -33,3 +36,9 @@ data WithEpochSlots a = WithEpochSlots
   , unWithEpochSlots :: a
   }
   deriving (Show, Eq)
+
+-- | Calculate the first slot in an epoch. Note that this function will fail if
```
dnadales: Actually the function won't fail, but will give an incorrect result :)

cardano-ledger/src/Cardano/Chain/ProtocolConstants.hs
```diff
@@ -21,6 +22,12 @@ import Cardano.Chain.Slotting.SlotCount (SlotCount(..))
 kSlotSecurityParam :: BlockCount -> SlotCount
 kSlotSecurityParam = SlotCount . (*) 2 . unBlockCount
 
+-- | Update stability parameter expressed in number of slots. This is the time
+--   between an protocol version update receiving its final endorsement and
+--   being accepted, and is set to double the security param.
```
dnadales, line 27: Maybe add a short explanation on why?

dnadales (approved):

PR #755, opened 2020-03-24
Document the address discrimination problem from cardano-sl (NetworkMagic)

### LINKED ISSUE #727: Add documentation for RequiresNetworkMagic and NetworkMagic (address discrimination)
There isn't much information around regarding the origin and purpose of the `RequiresNetworkMagic` and `NetworkMagic` data types or how/why we utilized them to solve the address discrimination problem in `cardano-sl`. 

Most of that information is in our old issue tracker (YouTrack) and it should definitely be publicly documented somewhere. So I'll add a markdown document somewhere in the repo that explains this.

intricate:
Closes #727.

### DIFF
--- a/doc/network-magic.md
+++ b/doc/network-magic.md
@@ -1,0 +1,123 @@
+# Address Discrimination (`NetworkMagic`)
+
+Address discrimination was a problem in the original
+[`cardano-sl`](https://github.com/input-output-hk/cardano-sl) implementation
+that came to our attention in mid-2018. This document briefly describes the
+problem as well as the solution implemented.
+
+For those with access to IOHK's old internal issue tracker, YouTrack, all of
+the nitty-gritty details can still be found under issues `CO-353` and
+`CO-354`.
+
+## The Problem
+
+In the original `cardano-sl` design, there was no way of distinguishing
+between addresses of different networks. This was undesirable because, for
+example, it was totally valid for a user to attempt a transfer of funds from a
+mainnet/staging address to a testnet address and vice versa which could result
+in a loss of those funds.
+
+## The Solution
+
+Our solution was to introduce a new field to addresses which would contain a
+specified protocol magic value (a magic value used to distinguish between
+networks). That way, we could check the protocol magic values contained within
+each address when validating transactions.
+
+However, because pre-existing mainnet and staging addresses would not contain
+this new field, we had to ensure that its inclusion would be optional in order
+to maintain backward compatibility.
+
+### `NetworkMagic`
+
+Given these requirements, we introduced a new data type to be used in
+representing this new field in each address:
+
+```
+data NetworkMagic
+  = NetworkMainOrStage
+  | NetworkTestnet !Word32
+```
+
+`NetworkMagic` effectively represents an _optional_ protocol magic ID value.
+Depending on the type of network, `NetworkMagic` is utilized in different ways
+with regard to its inclusion in addresses:
+
+  - For maintaining backward compatibility, all mainnet and staging addresses
+  *must not* contain a network magic, i.e. they would contain a value of
+  `NetworkMainOrStage`.
+
+  - All testnet addresses *must* contain a network magic, i.e. they would
+  contain a value of `NetworkTestnet`. At the time, we decided that it wasn't
+  necessary to maintain backward compatibility with pre-existing testnet
+  addresses.
+
+### `RequiresNetworkMagic`
+
+We also required a way to specify, via the node's configuration, whether a
+network requires the inclusion of network magic in its addresses. Thus, we
+introduced a new data type, `RequiresNetworkMagic`:
+
+```
+data RequiresNetworkMagic
+  = RequiresNoMagic
+  | RequiresMagic
+```
+
+We also added a `RequiresNetworkMagic` field to the `ProtocolMagic` data type.
+The `ProtocolMagic` data type would now contain two fields for which values
+would be specified in the configuration:
+
+  - One field for `RequiresNetworkMagic` which is specified in the node's
+  configuration.
+
+  - One field for the actual protocol magic ID (essentially a `Word32`) which
+  is specified in the genesis configuration.
+
+Here's a simplified version of the `ProtocolMagic` data type to display the
+result of adding this new field:
+
+```
+data ProtocolMagic = ProtocolMagic !Word32 !RequiresNetworkMagic
+```
+
+One might question why we didn't just allow the `NetworkMagic` value to be
+specified in the node's configuration as opposed to adding this new
+`RequiresNetworkMagic` type. After all, this would clearly indicate whether
+the network requires a `NetworkMagic` value to be serialized in its addresses
+since there are only two clear cases:
+
+  - `NetworkMainOrStage` - No network magic value serialized in addresses.
+
+  - `NetworkTestnet !Word32` - Network magic value serialized in addresses.
+
+However, the problem with this is that, in the `NetworkTestnet` case, the
+configuration could potentially specify a `Word32` value that differs from
+the protocol magic ID value already specified in the immutable genesis
+configuration. Because these values _must_ be the same, we figured it'd make
+sense to limit the potential for human error and, instead, derive the
+`NetworkMagic` value from the `ProtocolMagic` whenever necessary.
+
+Here's a simplified version of how this derivation might look:
+
+```
+makeNetworkMagic :: ProtocolMagic -> NetworkMagic
+makeNetworkMagic (ProtocolMagic pmId reqNetMagic) = case reqNetMagic of
+  RequiresNoMagic -> NetworkMainOrStage
+  RequiresMagic   -> NetworkTestnet pmId
+```
+
+### Transaction Validation With Address Discrimination
+
+With the "address discrimination" functionality:
+
+  - You cannot transfer funds from a mainnet/staging address to a testnet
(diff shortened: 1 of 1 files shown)


doc/network-magic.md
```diff
@@ -0,0 +1,53 @@
...
+  - All testnet addresses *must* contain a network magic, i.e. they would contain a value of `NetworkTestnet`. At the time, we decided that it wasn't necessary to maintain backward compatibility with pre-existing testnet addresses.
+
+### `RequiresNetworkMagic`
+
+We also required a way to specify, via the node's configuration, whether a network requires the inclusion of network magic in its addresses. Thus, we introduced a new data type, `RequiresNetworkMagic`:
+
+```
+data RequiresNetworkMagic
+  = RequiresNoMagic
+  | RequiresMagic
```
erikd, line 64: I have always wondered why we need this level of indirection. Would it not be possible to use a `NetworkMagic` value in the node configuration?

doc/network-magic.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+We also added a `RequiresNetworkMagic` field to the `ProtocolMagic` data type. The `ProtocolMagic` data type would now contain two fields for which values would be specified in the node's configuration: one for `RequiresNetworkMagic` and another for the `ProtocolMagicId` (a `newtype` wrapper around a `Word32`).
```
erikd: Probably useful to show that `ProtocolMagic` type here.

erikd (changes requested):
Would like to see `ProtocolMagic` show here. The other comment is mostly a question.

doc/network-magic.md
intricate: I was considering that, but it's a bit more complicated in `cardano-ledger` since it's actually an "annotated" `ProtocolMagic` (`AProtocolMagic`) :sweat_smile::

```
data AProtocolMagic a = AProtocolMagic
  { getAProtocolMagicId      :: !(Annotated ProtocolMagicId a)
  , getRequiresNetworkMagic :: !RequiresNetworkMagic
  } deriving (Eq, Show, Generic, NFData, NoUnexpectedThunks)
```

Perhaps I'll just show a simplified version of it that gets the point across:

```
data ProtocolMagic = ProtocolMagic !ProtocolMagicId !RequiresNetworkMagic
```

doc/network-magic.md
intricate: So we could have gone that route and instead allow the `NetworkMagic` value to be specified in the configuration, e.g.:
```
networkMagic: NetworkTestnet 1337
```
or
```
networkMagic: NetworkMainOrStage
```

However, we would still need to specify the protocol magic value in the configuration since it wouldn't be known for the mainnet or staging case (`NetworkMainOrStage`). Then that opens up the possibility that the `NetworkMagic` value could be configured to something different than the protocol magic (which would be invalid).

Because of this, we figured it'd be easier to derive the `NetworkMagic` value from the configured `ProtocolMagic`.

However, an alternative would be to define `NetworkMagic` like this:

```
data NetworkMagic
  = NetworkMainnet
  | NetworkStaging
  | NetworkTestnet !Word32
```

... and hardcode the protocol magic values for mainnet and staging. That way, you could just use `NetworkMagic` in the configuration.

intricate: So we could have gone that route and instead allow the `NetworkMagic` value to be specified in the configuration, e.g.:
```
networkMagic: NetworkTestnet 1337
```
or
```
networkMagic: NetworkMainOrStage
```

However, we would still need to specify the protocol magic value in the configuration since it wouldn't be known for the mainnet or staging case (`NetworkMainOrStage`). That would then open up the possibility for the `NetworkMagic` value to be configured to something different than the protocol magic. E.g., someone could set:

```
networkMagic: NetworkTestnet 1337
protocolMagic: 12345
```

... which would be invalid.

Because of this, we figured it'd be easier to derive the `NetworkMagic` value from the configured `ProtocolMagic`.

However, an alternative would be to define `NetworkMagic` like this:

```
data NetworkMagic
  = NetworkMainnet
  | NetworkStaging
  | NetworkTestnet !Word32
```

... and hardcode the protocol magic values for mainnet and staging. That way, you could just use `NetworkMagic` in the configuration.

doc/network-magic.md
erikd: Yeah, that makes sense, as long as you do note that its a simplified version.

doc/network-magic.md
erikd: I actually like that 3 constructor version.

intricate: Deleted my previous reply which was totally wrong :smile:.

So we could have gone that route and instead allowed the `NetworkMagic` value to be specified in the configuration, e.g.:
```
networkMagic: NetworkTestnet 1337
```
or
```
networkMagic: NetworkMainOrStage
```

However, because the protocol magic is already specified in the genesis config, this would then open up the possibility for the `NetworkMagic` value (in the `NetworkTestnet` case) to be configured to a value different than that of the protocol magic (which would be invalid). 

Overall, it seemed cleaner to introduce `RequiresNetworkMagic`, which would be specified in the node's configuration, and derive the `NetworkMagic` value from the protocol magic value specified in the genesis config, when needed.

doc/network-magic.md
```diff
@@ -0,0 +1,53 @@
+# Address Discrimination (`NetworkMagic`)
```
dnadales, line 1: Maybe break lines at 80 characters, to make it easy to review this and future PR's.

dnadales: I guess most editors would come with a command to automatically do this.

doc/network-magic.md
```diff
@@ -0,0 +1,53 @@
...
+
+Address discrimination was a problem in the original [`cardano-sl`](https://github.com/input-output-hk/cardano-sl) implementation that originally came to our attention in mid-2018. This document briefly describes the original problem as well as the solution implemented.
```
dnadales:
```suggestion
Address discrimination was a problem in the original [`cardano-sl`](https://github.com/input-output-hk/cardano-sl) implementation that came to our attention in mid-2018. This document briefly describes the problem as well as the solution implemented.
```

doc/network-magic.md
```diff
@@ -0,0 +1,53 @@
...
+
+## The Problem
+
+In the original `cardano-sl` design, there was no way of distinguishing between addresses of different networks. This was undesirable because, for example, it was totally valid for a user to attempt a transfer of funds from a mainnet/staging address to a testnet address and vice versa which would result in a loss of those funds.
+
+## The Solution
+
+Our solution was to introduce a new field to addresses which would contain a specified protocol magic value (a magic value used to distinguish between networks). That way, we could check the protocol magic values contained within each address when validating transactions.
+
+However, because pre-existing mainnet and staging addresses would not contain this new field, we had to ensure that its inclusion would be optional in order to maintain backward compatibility.
```
dnadales: backwards compatibility?

doc/network-magic.md
dnadales: Maybe we should explain this as well. Although I'm still not sure I get why one couldn't simply use `NetworkMagic` everywhere dispensing from the need of using `RequiresNetworkMagic`.

dnadales (approved):
Thanks Luke!

doc/network-magic.md
dnadales: Oh, maybe it'd be worthwhile explaining briefly what the `ProtocolMagic` is, and what would have happened if `RequiresNetworkMagic` was not included here. (Another way of saying why its inclusion in this data type is important).

doc/network-magic.md
intricate: Oops, my bad.

doc/network-magic.md
intricate: Did I misuse the term here or are you suggesting a spelling error, i.e. "backward" -> "backwards"? If the latter, I think it can be referred to as either.

dnadales: No you're right: "backward compatibility" is fine!

doc/network-magic.md
intricate: Will do.

Jimbo4350 (approved):
LGTM

doc/network-magic.md
```diff
@@ -0,0 +1,122 @@
+# Address Discrimination (`NetworkMagic`)
+
+Address discrimination was a problem in the original
+[`cardano-sl`](https://github.com/input-output-hk/cardano-sl) implementation
+that came to our attention in mid-2018. This document briefly describes the
+problem as well as the solution implemented.
+
+For those with access to our old issue tracker, YouTrack, all of the
+nitty-gritty details can still be found under issues `CO-353` and `CO-354`.
```
erikd: This is a public document, so its probably worth mentioning that those issues are in the "internal IOHK Youtrack bugtracker".

doc/network-magic.md
```diff
@@ -0,0 +1,122 @@
...
+
+## The Problem
+
+In the original `cardano-sl` design, there was no way of distinguishing
+between addresses of different networks. This was undesirable because, for
+example, it was totally valid for a user to attempt a transfer of funds from a
+mainnet/staging address to a testnet address and vice versa which would result
+in a loss of those funds.
```
erikd, line 18: I would replace "would" with "could" (If the person doing that transfer had access to the private keys of the destiination address they are still recoverable.)

erikd (dismissed):
Two really minor changes required, and once those are sorted, please merge away.

PR #766, opened 2020-04-08
Reject null update proposals

### LINKED ISSUE #759: Null update proposals are ignored, not rejected
As per https://github.com/input-output-hk/cardano-ledger/blob/512c26a66a6a63278846646b81bf8eadcd4ae99c/cardano-ledger/src/Cardano/Chain/Update/Validation/Registration.hs#L266-L286, the current ledger code allows proposals that make no change to either the application or protocol versions. This is not in line with the spec, which requires that _at least one_ of these components must be updated.

nc6:
This addresses #759


intricate (approved):
LGTM.

